PR #165, opened 2014-11-08
Add byte[] bypass methods to JsonGenerator

sethp-jive:
Allows users with some known combination of byte[] and Charset to avoid unnecessary copies if the output generator supports doing so. Provides an easy API in other cases.

These methods take an explicit java.nio.Charset instead of a JsonEncoding or attempting autodetection to allow users to pass in fragments or encodings that compose/transcode to valid json but are not json on their own.


cowtowncoder:
Closing this as I am not sure how to go about it; may be re-opened or new issue created if further improvements are desired.

PR #181, opened 2015-02-14
fix failure parsing -Infinity on buffer boundary

brharrington:
In cases where a -Infinity value gets split due to a buffer
boundary an exception was thrown indicating an invalid numeric
value. With the test case added to TestNumericValues I get an
exception:

```
testParsingOfLongerSequencesWithNonNumeric(com.fasterxml.jackson.core.json.TestNumericValues)  Time elapsed: 5.306 sec  <<< ERROR!
com.fasterxml.jackson.core.JsonParseException: Invalid numeric value: Missing integer part (next char 'I' (code 73))
 at [Source: java.io.StringReader@164b3e4f; line: 1, column: 50690]
	at com.fasterxml.jackson.core.JsonParser._constructError(JsonParser.java:1501)
	at com.fasterxml.jackson.core.base.ParserMinimalBase._reportError(ParserMinimalBase.java:522)
	at com.fasterxml.jackson.core.base.ParserBase.reportInvalidNumber(ParserBase.java:1025)
	at com.fasterxml.jackson.core.json.ReaderBasedJsonParser._parseNumber2(ReaderBasedJsonParser.java:1050)
	at com.fasterxml.jackson.core.json.ReaderBasedJsonParser._parseNegNumber(ReaderBasedJsonParser.java:964)
	at com.fasterxml.jackson.core.json.ReaderBasedJsonParser.nextToken(ReaderBasedJsonParser.java:668)
	at com.fasterxml.jackson.core.json.TestNumericValues.testParsingOfLongerSequencesWithNonNumeric(TestNumericValues.java:470)
```

I don't know if it is the best fix or not, but it was fairly
easy to fix the problem I was seeing by changing ReaderBasedJsonParser
to call `_handleInvalidNumberStart` instead of `reportInvalidNumber`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -66,6 +66,7 @@ com.fasterxml.jackson.core.*;version=${project.version}
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
         <configuration>
+          <argLine>-Xmx128m</argLine>
           <redirectTestOutputToFile>${surefire.redirectTestOutputToFile}</redirectTestOutputToFile>
           <excludes>
             <exclude>**/failing/*.java</exclude>
--- a/src/main/java/com/fasterxml/jackson/core/json/ReaderBasedJsonParser.java
+++ b/src/main/java/com/fasterxml/jackson/core/json/ReaderBasedJsonParser.java
@@ -1047,7 +1047,7 @@ public class ReaderBasedJsonParser // final in 2.3, earlier
         }
         // Also, integer part is not optional
         if (intLen == 0) {
-            reportInvalidNumber("Missing integer part (next char "+_getCharDesc(c)+")");
+            return _handleInvalidNumberStart(c, neg);
         }
 
         int fractLen = 0;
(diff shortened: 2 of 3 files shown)


cowtowncoder:
Thanks! I'll use test as modified, but will need to probably apply changes manually here.
FWIW, looks like 2.3 and 2.4 are not affected, so it is regression in 2.5 (or, possibly, depends on different buffer sizes to trigger the problem?).

cowtowncoder:
Weird. Can not reproduce from IDE (Eclipse), but can from command-line. Also sensitive to size of input, but in unpredictables ways. I guess that is due to buffer recycling preventing problems in certain cases.

cowtowncoder:
I went ahead and added test as suggested, as well as fix: I think it makes sense. Control flow is bit convoluted due to earlier optimizations, and I wonder if exception is optimal now. But more important is that no exception is thrown on buffer boundary; exception messages can be improved later on.

Thank you for reporting & investigating this; fix will be in 2.5.2.

PR #205, opened 2015-07-19
Force Unix LF for indent test

christophercurrie:
The indent test fails to pass on windows, because the DefaultIndenter uses the system linefeed, but the test is written using unix line feeds.


cowtowncoder:
Thanks!

PR #206, opened 2015-07-30
Replace the Travis png badge with svg

0xced:


cowtowncoder:
No objections per se, but curious: what is the benefit? Allows scaling?

0xced:
It looks much better on high definition screens.

cowtowncoder:
Ok, that makes sense.

PR #212, opened 2015-08-20
Set Accept header on ObjectMapper.readTree for HTTP URLs

mattnelson:
When using ObjectMapper.readTree(URL) for a HTTP URL the `Accept: application/json` header is not supplied. If a service produces multiple outputs for the same URL and the accept header is not supplied it appears to default to `Accept: text/html, image/gif, image/jpeg, *; q=.2, */*; q=.2`

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/JsonFactory.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonFactory.java
@@ -7,6 +7,7 @@ package com.fasterxml.jackson.core;
 import java.io.*;
 import java.lang.ref.SoftReference;
 import java.net.URL;
+import java.net.URLConnection;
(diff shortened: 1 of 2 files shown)


cowtowncoder:
First of all, thank you for your contribution. I understand why this would be useful.

But I don't think it is a good idea to add specific handling to URL this way; users might (for example) want different header(s) being added. I would be open to having an extension for handlers to do that (so caller would provide something to set properties of connection), but at that point it would be as easy to just construct `InputStream` manually from said URL. So it might not simplify handling a lot.

Another way to tackle this would be to add an overloaded variant that would either take a `String` to pass as `Accept` header; or a `Map<String,String>` of properties to add.

PR #214, opened 2015-08-22
copyCurrentStructure(JsonParser) is made non-recursive

santhosh-tekuri:


cowtowncoder:
Would be nice to know benefit of such change. In theory could be more efficient that way, but if that's the goal, a micro-benchmark that shows benefits would be necessary to verify there is actual difference.
For now will close; may be reopened with more supporting evidence for change.

PR #221, opened 2015-10-06
Fixed ArrayIndexOutOfBounds exception

a-lerion:
Got an exception here today and decided to make a fix. Is this a right branch for that?


cowtowncoder:
I think this would work, yes. Although if it exists in earlier branches I might go for even earlier version.

One thing that would be great would be a simple unit tests, to watch for regression.

cowtowncoder:
This did affect earlier versions, so I ended up making suggested change in 2.5 branch (for potential 2.5.5), merged upwards to be in 2.6.3.

Thank you very much for the fix -- even without test case it's an obvious bug.

Issue #37, opened 2012-11-05
JsonParser.getTokenLocation() doesn't update after field names.


mikelehen:
There's a unit test to repro the issue below.  Basically, when you're on a FIELD_NAME token, if you call getTokenLocation() and then nextToken() and then getTokenLocation() again, you'll get the same location for both calls to getTokenLocation(), even though you've advanced to a new token.

The issue seems to be the _nextToken logic in ReaderBasedJsonParser and UTF8StreamJsonParser.  When calling nextToken() on a FIELD_NAME, it calls _nextAfterName(), which updates _currToken but doesn't update _tokenInputRow and _tokenInputCol for the new token's location.

I started to try to fix it, but the _nextToken logic is spread across so much code that it looked like it'd be a pretty major surgery.  Not something I'm willing to do at this point. :-)

    public void testTokenLocationAfterFieldName() throws Exception
    {
        _testTokenLocationAfterFieldName(false);
        _testTokenLocationAfterFieldName(true);
    }

    private void _testTokenLocationAfterFieldName(Boolean useStream) throws Exception
    {
        final String DOC = "{\"name\":123}";
        JsonFactory jf = new JsonFactory();
        JsonParser jp = useStream ?
                jf.createJsonParser(new ByteArrayInputStream(DOC.getBytes("UTF-8")))
                : jf.createJsonParser(new StringReader(DOC));

        assertEquals(JsonToken.START_OBJECT, jp.nextToken());
        assertEquals(JsonToken.FIELD_NAME, jp.nextToken());
        assertEquals(JsonToken.VALUE_NUMBER_INT, jp.nextToken());
        assertEquals(1, jp.getTokenLocation().getLineNr());
        assertEquals(9, jp.getTokenLocation().getColumnNr());
        jp.close();
    }

aborg0:
It gets worse when the values are not the first in an object. For example:

    {
      "book" : [ {
        "year" : 1999,
        "title" : "Title",
        "author" : "Author"
      }]
    }

In this case for `title`, the `getTokenLocation()` and the `getCurrentLocation()` select both for the `FIELD_NAME` and the string value the following text:

                            ,
        "author"

Please recognize no value is selected this time, and the previous `,` is also selected. Similar behaviour for the `author`.
(This is with version: 2.4.3.)

cowtowncoder:
Is this with byte- or char-backed source? Or both.

aborg0:
[...]

PR #227, opened 2015-11-10
Fixed issue #37

complexmath:
This is a fix for Issue #37 (location not updating properly when parsing objects). The basic issue was that the UTF8StreamParser was aggressively pre-parsing the value portion of object fields and caching that token, but not caching the corresponding parse location as well. Also, the parse position was not being properly updated in all locations.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/json/ReaderBasedJsonParser.java
+++ b/src/main/java/com/fasterxml/jackson/core/json/ReaderBasedJsonParser.java
@@ -25,6 +25,33 @@ public class ReaderBasedJsonParser // final in 2.3, earlier
 
     /*
     /**********************************************************
+    /* Information about starting location of event
+    /* Reader is pointing to; updated on-demand
+    /**********************************************************
+     */
+
+    // // // Location info at point when current token was started
+
+    /**
+     * Total number of bytes/characters read before start of current token.
+     * For big (gigabyte-sized) sizes are possible, needs to be long,
+     * unlike pointers and sizes related to in-memory buffers.
+     */
+    protected long _nextTokenInputTotal = 0;
+
+    /**
+     * Input row on which current token starts, 1-based
+     */
+    protected int _nextTokenInputRow = 1;
+
+    /**
+     * Column on input row that current token starts; 0-based (although
+     * in the end it'll be converted to 1-based)
+     */
+    protected int _nextTokenInputCol = 0;
+
+    /*
+    /**********************************************************
     /* Input configuration
     /**********************************************************
      */
@@ -50,7 +77,7 @@ public class ReaderBasedJsonParser // final in 2.3, earlier
      * buffer.
      */
     protected boolean _bufferRecyclable;
-    
+
     /*
     /**********************************************************
     /* Configuration
@@ -60,7 +87,7 @@ public class ReaderBasedJsonParser // final in 2.3, earlier
     protected ObjectCodec _objectCodec;
 
     final protected CharsToNameCanonicalizer _symbols;
-    
+
     final protected int _hashSeed;
 
     /*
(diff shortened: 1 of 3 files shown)


cowtowncoder:
Thanks! Would be great thing to fix. My only concern here is performance; I'd have to verify with something like `jackson-benchmarks` that this has negligible impact on parsing speed (and/or how to mitigate if there is bigger change). Normally wouldn't be a big deal, but this is directly on critical path for decoding, and changes sometimes have outsized effects.

One other thing is that there are 2 parser implementations, so `ReaderBasedJsonParser` would also need to be fixed similarly, and tested.

complexmath:
I'll look at the ReaderBasedJsonParser. With luck, it should be easy to apply the same fixes there.

cowtowncoder:
Yes, it should follow same patterns and fix should be repeatable.

complexmath:
That should do it. Let me know if I can help with the benchmark.

cowtowncoder:
@complexmath You should be able to get `jackson-benchmarks`, and run with 2.7.0-SNAPSHOT, after local build. No changes to benchmark should be needed, just recompile and build with different local version ("mvn clean package" should be enough, after locally doing "mvn clean install" for 'jackson-core' -- you also need to locally build `jackson-databind` snapshot, maybe other modules too, but except for core, no need to re-install between tests, just once).

Script like `run-mediaitem.sh` (or similar) should show how to do it; here we'd care about "JsonStdReadVanilla" (that is, for json format, no afterburner ... although "JsonStdReadAfterburner" would also be of interest). When run over long enough time (like 5 sets of 9 iterations or whatever) it should give some indication if there is any difference.

cowtowncoder:
It looks like performance impact is at most 1-2% on my local build, using default `jackson-benchmarks` case (which is modification of `jvm-serializers` type). That's small enough to fall within error of margin, so it may be less, or slightly more, but that seems acceptable. I'll want to see if Afterburner case is affect significantly, but I suspect it should not show much more effect.

About the only possible idea for change might be to see if instead of copying values, state was checked in `getLocation()`, to choose which source to use (name vs any other token). This could avoid one set of copy. It's not a big deal and might not have much or any effect, but thought I'll mention it for fun.

But aside from that, one process thing: unless I already asked for and received Contributor License Agreement (CLA, from https://github.com/FasterXML/jackson/blob/master/contributor-agreement.pdf), we would need to get filled one before merging.
It is a one-time thing and good for all contributions; we need one just because corporate users are very cautious about OSS projects having provenance for code. But it's the only piece of process we have and I'm happy to keep that to minimum. :)
The usual way is for developer to print the doc, fill, sign, scan and email to `info` at fasterxml dot com. And then I can merge the contribution.

Looking forward to merging this patch -- thank you once again, as you know this is a long-standing bug and many users are going to be happy to see it fixed!

complexmath:
Yeah, I was going for the simplest approach possible. I imagine things could be tuned a tad, though I suspect the performance difference wouldn't be measurable. I'll see what I can do about signing the document.

cowtowncoder:
@complexmath thanks!

cowtowncoder:
@complexmath Looks like there is also issue #229, which seems easy enough to fix. I think my best option may be to do bit of manual merging here, as I have some ideas of tiny simplifications to update, mostly eliminate (or, rather, move) those +1 and -1 offset modifications. But I can start fix the simple fix, tests, and then see if I can do some tuning.

cowtowncoder:
Starting to merge parts manually, first offlining simple update case, calling it from appropriate places. This is enough to fix #229. Next step will be to add temporary offsets for value (during name parsing), which is needed for this problem.

complexmath:
Awesome. The contract thing has been held up because my manager has been out of the office for a family emergency. Not sure when that will happen, but hopefully soon.

cowtowncoder:
Fixed piece meal; thank you once again!

complexmath:
Thank you!

Issue #222, opened 2015-10-08
Charset autodetection fail


Spikhalskiy:
Hello,

JsonFactory#createParser(InputStream in) has the [comment](https://github.com/FasterXML/jackson-core/blob/master/src/main/java/com/fasterxml/jackson/core/JsonFactory.java#L783):
```
Note: no encoding argument is taken since it can always be auto-detected as suggested by JSON RFC.
```
Looks like it doesn't work and now we have no ability to explicitly setup encoding.

Below I explain [demo](https://github.com/Spikhalskiy/jackson-utf-8-ISO-8859-1-detect-error/blob/master/src/test/java/EncodingTest.java) for stream with ISO-8859-1 content when our application encoding is UTF-8 (common for java).

Description:
[Hex dump](http://bin-hex-converter.online-domain-tools.com/link/3ca460gp0D236o5Yy/) - hex dump of valid json string payload encoded in ISO-8859-1, contains ã symbol, which has different encoding in UTF-8 and ISO-8859-1
```
String hexDump = <hex dump >;
byte[] bytes = BaseEncoding.base16().decode(hexDump.toUpperCase()); //load hex dump to byte array
System.out.println("Try to read print it with UTF-8" + new String(bytes)); //we see ? - something can't be decoded
System.out.println("Better (ISO-8859-1): " + new String(bytes, Charset.forName("ISO-8859-1"))); //do it with right encoding and will get valid json dump in console

//trying to create json parser
JsonFactory jf = new JsonFactory();
JsonParser parser = jf.createParser(new ByteArrayInputStream(bytes));
System.out.println(parser.getClass().getSimpleName());  //it will be UTF8StreamJsonParser, which already seems to be incorrect
 while (parser.nextToken() != null) {}
```

As a result we will get
```
com.fasterxml.jackson.core.JsonParseException: Invalid UTF-8 middle byte 0x5f
 at [Source: java.io.ByteArrayInputStream@373bc99b; line: 1, column: 764]
	at com.fasterxml.jackson.core.JsonParser._constructError(JsonParser.java:1581)
	at com.fasterxml.jackson.core.base.ParserMinimalBase._reportError(ParserMinimalBase.java:533)
	at com.fasterxml.jackson.core.json.UTF8StreamJsonParser._reportInvalidOther(UTF8StreamJsonParser.java:3470)
[...]

PR #232, opened 2015-11-29
Clarified javadocs about charset auto-detection for issue #222

Spikhalskiy:
Added clarification about supported charsets in Jackson auto-detection according to discussion in issue #222


cowtowncoder:
Much appreciated, thank you!

Issue #118, opened 2013-11-14
feature to allow trailing commas (non-compliant JSON)


cburroughs:
Splitting off from #52 and originally at http://jira.codehaus.org/browse/JACKSON-618

Presumably a feature named: ALLOW_TRAILING_COMMA

If enabled the following would be valid and yield the same result:
- `[1,2,3,]    [1,2,3]`
- `{1:1,}       {1:1}`

albertocavalcante:
I'm in a situation where a third party is sending a leading comma:

- `{,1:1} {1:1}`

Any chance already having a feature flag similar to the ALLOW_TRAILING COMMA? Or any tips on implementing this custom deserialization?

Thanks

cowtowncoder:
@albertocavalcante There is `JsonReadFeature.ALLOW_MISSING_VALUES` (and deprecated `JsonParser.Feature.ALLOW_MISSING_VALUES`) which might work.

Issue #117, opened 2013-11-13
Support for missing values (non-compliant JSON)


cemo:
This is not my top priority but might be useful for someone else too. As we discussed at https://github.com/FasterXML/jackson-databind/issues/345,  I am opening this issue for further references.

```
      ObjectMapper objectMapper = new ObjectMapper();
      objectMapper.configure(JsonParser.Feature.ALLOW_UNQUOTED_FIELD_NAMES, true);
      objectMapper.configure(JsonParser.Feature.ALLOW_SINGLE_QUOTES, true);
      objectMapper.enable(DeserializationFeature.ACCEPT_EMPTY_STRING_AS_NULL_OBJECT);
      String str = "{a:[1,2,3]}";
      Map x = objectMapper.reader(Map.class).readValue(str);
      System.out.println(((ArrayList)x.get("a")).get(0));
      str = "{a:[1,null,3]}";
      x = objectMapper.reader(Map.class).readValue(str);
      System.out.println(((ArrayList)x.get("a")).get(0));
      str = "{a:[1,,3]}";
      x = objectMapper.reader(Map.class).readValue(str);
      System.out.println(((ArrayList)x.get("a")).get(0));
```

cowtowncoder:
Thanks to @LokeshN this is now implemented, will be in 2.8.0.

cowtowncoder:
Looks like implementation lead to #616 (non-closing root level stream, serving `VALUE_NULL`s.).

PR #234, opened 2015-12-08
Add JsonParser feature to ignore trailing commas (#118)

bdhess:


cowtowncoder:
Thank you for the contribution! I will have to take a look here, partly since changes occur within critical-path from performance perspective. This will probably also need to go after 2.7.0, but that will give bit more time to ensure everything works properly.

bdhess:
I did clean up the language in the Javadoc and force-pushed over it.  Hope that didn't disturb you!  Looking forward to following up.

bdhess:
@cowtowncoder - can I trouble you to bump this up for review?  Just rebased against the latest in master.  I've run the perf tests and the difference is negligible.

cowtowncoder:
Forgot to add a note, but I think #117 covers same issue to some degree. It will induce `null`s, which may be different, but at least allows usage.

PR #235, opened 2015-12-11
Add `JsonGenerator.Feature` to use single quotation marks instead of double quotes

dgolub:


cowtowncoder:
Thank you for the contribution. This could be considered for 2.8. I think I'd change configuration slightly to have local `char`/`byte` to contain quote character to use, otherwise probably could be merged as-is.
I hope to look into merging this in near/medium future; right now there's big backlog of 2.7 fixes.

cowtowncoder:
Made a small refactoring so that both generators now use a field `_quoteChar` to contain character/byte used as quote character. This should simplify implementation a bit.

fak98:
I need this feature. Is there an ETA for it yet?

cowtowncoder:
No. Not sure it'll be worked anytime soon.

grexe:
well 1 year later and I stumble over the same need, can this PR please be integrated for 2.9 at least? Escaped XML in JSON responses looks a bit ugly...

cowtowncoder:
Created/transferred matching issue, #617

cowtowncoder:
If anyone wants to pick this up, PR is pretty good -- the one suggestion I would make is that instead of creating a new method for writing out quote (`_writeQuote()`), I would actually add a field to contain quote character and use that in places.
I will add into CREDITS both original submitter here and whoever works on updated patch if need be.

cowtowncoder:
As pointed out by @angelyan, this would implemented as #549 for more general applicability.

PR #238, opened 2015-12-14
Code quality fix - final classes should not have "protected" members.

faisal-hameed:
This pull request is focused on resolving occurrences of rule: 
squid:S2156 - final classes should not have "protected" members.

You can find more information about the issue here:
http://dev.eclipse.org/sonar/rules/show/squid:S2156

Please let me know if you have any questions.

Faisal

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/Base64Variant.java
+++ b/src/main/java/com/fasterxml/jackson/core/Base64Variant.java
@@ -78,17 +78,17 @@ public final class Base64Variant
      * Note that this is the only non-transient field; used when reading
      * back from serialized state
      */
-    protected final String _name;
+    final String _name;
 
     /**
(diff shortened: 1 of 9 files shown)


cowtowncoder:
I think I'll pass on this one actually; apologies for extra work. The reason here is that I don't really think this is superbly confusing to developers (I'd call it more a "pet peeve"); but also since I have in the past opened up final classes (allowed sub-classing), there's some chance I'd have to go back and re-do changes. Further, it is quite possible some/many of these really should be made `private`; so in the end mass-conversion may not be all that useful.

Thank you again for doing PRs; I will still be considering all as input and suggestions, so I may still selectively apply changes from all remaining PRs.

faisal-hameed:
Thanks for the explanation. This helps me to understand your concern.

PR #246, opened 2016-01-17
Fix UTF8JsonGenerator to allow QUOTE_FIELD_NAMES to be toggled.

philipa:


cowtowncoder:
Thank you for contributing this! I ended up merging it in slightly modified form, partly since I wanted it backported in 2.7, and master is now for 2.8.0-SNAPSHOT, and partly to refactor to make char-backed output also use same mechanism. Fix will be included in 2.7.2 to be released soon.

Issue #248, opened 2016-01-26
VersionUtil.versionFor() unexpectedly return null instead of Version.unknownVersion() stated in JavaDoc


sammyhk:
As titled, after commit of 88c296c67391f6dea4c1581caa82d801c5d94535 , it unexpectedly return `null` instead of `Version.unknownVersion()` which cause our application broken after upgrade from 2.4.x to 2.6.x.

cowtowncoder:
Thank you for reporting this. Looks like unintended consequence of refactoring; will fix.

PR #249, opened 2016-01-26
#248 - fix VerisonUti.versionFor() unintended to return null instead …

sammyhk:
…of Version.unknownVersion() after commit 88c296c67391f6dea4c1581caa82d801c5d94535


cowtowncoder:
Thanks!

Issue #86, opened 2013-07-05
Allow inclusion of request body for JsonParseException


cowwoc:
As a follow-up to issue #77, please include the entire request body in the exception's message in case of a parsing error. Users will need this extra context to figure out why the parsing error occurred.

cowtowncoder:
So, as can be seen from above, #258 makes it possible to attach content into exception.
Could perhaps consider automation for simple cases where input comes as `String` or `byte[]`...

PR #258, opened 2016-03-15
Issue #86 - Adding request body to the JsonParseException on demand

LokeshN:
This commit adds the request body(issue #86 ) to the JsonParseException on demand.
By setting the property addRequestBodyOnError in JsonParser, the request body is added to JsonParseException. 
The request body is added as a separate property to the JsonParseException which can be retrieved using JsonParseException.getRequestBody()

By default the request body will not be added to the JsonParseException as the flag is turned off, it can be turned on by using JsonParser.addRequestBodyOnError(true);

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/JsonParseException.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonParseException.java
@@ -5,6 +5,13 @@
 
 package com.fasterxml.jackson.core;
 
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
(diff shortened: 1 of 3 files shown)


cowtowncoder:
While I appreciate your suggestion for solution, I don't think I will accept it. Most `InputStream`s and `Reader`s are not rewindable and so most of the contents would not be available to be included in exception. There are other potential issues (max length of contents to read should probably be bounded), but that is the main one.

LokeshN:
I have created another pull request with a different solution.
Closing this for now.

PR #259, opened 2016-03-16
Issue86 - Adding request body to the JsonParseException

LokeshN:
This commit fixes the issue #86, where request body as byte[] is taken as input (taking byte[] as input to avoid any rewind errors) to the JsonParser along with the charset and will be displayed in the JsonParseException.
 Along with that there is an additional method - getRequestBody() to get only the request body from the JsonParseException.
By default, the request body will not be shown in the exception as it will be null. Once the request body is set it will be shown as part of the exception.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/JsonParseException.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonParseException.java
@@ -13,6 +13,8 @@ import java.io.StringWriter;
 import java.nio.CharBuffer;
 import java.nio.charset.Charset;
 
+import com.fasterxml.jackson.core.util.RequestPayloadWrapper;
+
 /**
  * Exception type for parsing problems, used when non-well-formed content
  * (content that does not conform to JSON syntax as per specification)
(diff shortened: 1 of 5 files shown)


cowtowncoder:
Ok, I think I can see how something this could work. Since inclusion would be controlled externally, concerns about maximum contents could be alleviated (and possible other security concerns about showing sources).

One aspect that could perhaps be further improved would be to encapsulate details of input source (byte[] vs String, mostly) in a wrapper object, on which `toString()` could be called to produce input?
That would also slightly simplify passing of source, and allow handling of common case of non-aligned/partially full buffer (that is, one with non-zero start offset and/or length that does extend until end of buffer). I would be fine with reuse of a wrapper for this from within JDK, or simple new type if need be.

LokeshN:
Thanks for the suggestion. I have created a simple wrapper to hold the data(bytes).
Let me know if this is fine.

cowtowncoder:
I think that makes sense. One suggestion on payload container is that it probably should also accept a `String` as alternate source? Also: exception should probably have alternate getter to access wrapper as-is, without constructing `String`, for sake of completeness.

LokeshN:
Updated the PR according to the review comments.

Issue #209, opened 2015-08-04
Make use of `_allowMultipleMatches` in `FilteringParserDelegate`


zazi:
Currently, it looks like that the _allowMultipleMatches attribute in FilteringGeneratorDelegate is not utilised (i.e. no value is assigned to this variable). Re. the documentation this attribute offers some useful functionality. So it would be nice, if it could be implemented properly. See https://groups.google.com/d/msg/jackson-user/VzZ94G9hvrs/JGFozl6lCQAJ

PR #263, opened 2016-03-20
Fix for issue 209 - make use of _allowMultipleMatches FilterParsingDelegate

LokeshN:
The commit fixes the issue #209 , when _allowMultipleMatches is set to false, then there should be only one match and if it is set to true then it continues to match the token until the end of the document.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/filter/FilteringParserDelegate.java
+++ b/src/main/java/com/fasterxml/jackson/core/filter/FilteringParserDelegate.java
@@ -221,6 +221,21 @@ public class FilteringParserDelegate extends JsonParserDelegate
     @Override
     public JsonToken nextToken() throws IOException
     {
+    	//Check for _allowMultipleMatches - false and atleast there is one token - which is _currToken
+    	// check for no buffered context _exposedContext - null
+    	//If all the conditions matches then check for scalar / non-scalar property
+    	if(!_allowMultipleMatches && _currToken != null && _exposedContext == null){
+    		//if not scalar and ended successfully, then return null
+    		if((_currToken.isStructEnd()  && _headContext.isStartHandled()) ){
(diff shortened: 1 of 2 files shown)


cowtowncoder:
I assume this works as per unit test. Will be happy to merge this.

One practical thing we need, before accepting your first contribution, is to get a filled Contributor License Agreement, from:

(there is also an alternative Corporate CLA available, but most developers choose individual one; both are fine and choice does not matter to us)

The usual way is to print it, fill, sign, scan and email to `info` at fasterxml dot com.
If you can do that, I can merge the patch once I receive the filled in CLA.

LokeshN:
Hi Tatu,
I have mailed the CLA to you.

Thanks,
Lokesh

cowtowncoder:
Thank you again! I merge this into 2.7 (took a bit longer, need to learn how to use rebase...) so it should be part of 2.7.4, as well as `master` (2.8.0-SNAPSHOT).

cowtowncoder:
Hmmh. Looks like this may have broken a unit test in `jackson-databind`, related to filtering of sub-trees.
Need to investigate, but seems the only change recently. Noticed breakage for master, need to verify with 2.7.

cowtowncoder:
Yes, fails with 2.7 too.

cowtowncoder:
Multiple possible reasons, too. One would be test failure, but tests are simple enough. Another could be misconfiguration of filter (not setting multiple matches); that seems most likely.

cowtowncoder:
Ah! Yes, with "readValues()", multiple matches MUST be allowed. With "readValue()" single match probably fine. Will fix in `ObjectReader`. Actually it is kind of cool that this feature is thus verified...

LokeshN:
Thanks Tatu for fixing it

Issue #265, opened 2016-03-22
JsonStringEncoder should take CharSequence parameters


mikaelstaldal:
It would be useful if the public methods in `com.fasterxml.jackson.core.io.JsonStringEncoder` declare its input parameter as `CharSequence` rather than `String`.

To me, it seems to be trivial to fix since the code only uses methods on the input parameter which are in `CharSequence` (`length` and  `charAt`).

cowtowncoder:
I can have a look at this. The main concern would be backwards compatibility; change of type here would be gnarly binary-incompatibility (esp. since it's source compatible), but addition of an alternate method could work.

But beyond this, performance via interface is probably inferior, so the first thing encoder would most likely do would be to use `toString()`. At which point caller could already do that...

If I find time I can do a micro-benchmark to see if my guess is correct wrt speed of `charAt()`.

mikaelstaldal:
This should only be done if it can be done without using `toString()` on the `CharSequence`, otherwise there would be no point. This point of this is performance, to avoid superfluous copying of the string.

I am quite sure it can be done without using `toString()` on the `CharSequence`.

cowtowncoder:
@mikaelstaldal Functionally speaking it can certainly be done, but the point is to keep performance high. Changing access from `String.charAt()` to go via `CharSequence` interface may impose significant overhead of its own unless JIT compiler can speculative inline accessors for call-points.

Jackson core itself has no use for sources other than `char[]` or `String`, so I assume this functionality would be for external usage.

Beyond performance test showing there is no measurable difference, what would be needed would be new method(s) to avoid backwards-compatibility breakage, and leaving existing method in place.

mikaelstaldal:
Yes, I request this for external use. Specifically to be able to pass in a `StringBuilder` without having to do `toString()` on it (which would copy its content). The goal is both to avoid superfluous copying of the string, and to avoid allocating a new object (and thus put pressure on the GC).

cowtowncoder:
@mikaelstaldal Makes sense. Actually I think that despite duplication of code that is unfortunate in general, perhaps just cut'n pasting method with new name, more generic type, would be the way to go here? I'd merge such a PR.

mikaelstaldal:
Does the method need to have a new name? Can't we rely on overloading?

cowtowncoder:
[...]

PR #266, opened 2016-03-25
CharSequence variant of JsonStringEncoder.quoteAsString

mikaelstaldal:
Fix for https://github.com/FasterXML/jackson-core/issues/265

### DIFF
--- a/src/test/java/com/fasterxml/jackson/core/io/TestJsonStringEncoder.java
+++ b/src/test/java/com/fasterxml/jackson/core/io/TestJsonStringEncoder.java
@@ -20,6 +20,19 @@ public class TestJsonStringEncoder
         assertArrayEquals("\\\"x\\\"".toCharArray(), result);
     }
 
+    public void testQuoteAsString2() throws Exception
+    {
+        JsonStringEncoder encoder = new JsonStringEncoder();
+        StringBuilder builder = new StringBuilder();
+        builder.append("foobar");
+        char[] result = encoder.quoteAsString(builder);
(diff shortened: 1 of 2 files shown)


src/test/java/com/fasterxml/jackson/core/io/TestJsonStringEncoder.java
```diff
@@ -37,7 +50,24 @@ public void testQuoteLongAsString() throws Exception
         assertEquals(exp, new String(result));
         
     }
-    
+
+    // For [JACKSON-853]
```
cowtowncoder: Could you remove this, or change to point to [core#266] (or whatever was id of the issue)?

cowtowncoder:
Looks good, and I'd be happy to merge it in. The only thing before that is to make sure I have an CLA for you -- I don't think I've asked for one yet?

You can find the doc at:

and the usual thing to do is to print, fill & sign, scan and email it to `info` at fasterxml dot com.
If you could do that, I can merge the PR as soon as I have it. Also, this is one time thing: further merges are fine for any and all Jackson projects. Apologies for having to do this, but it's the minimum we need from process perspective (some companies are very strict on ensuring OSS libs they use make sure to get confirmation that all contributions are valid etc).

src/test/java/com/fasterxml/jackson/core/io/TestJsonStringEncoder.java
mikaelstaldal: OK.

cowtowncoder:
Looks good, got CLA, merging.

PR #267, opened 2016-03-27
Fix for issue #117 - Supports the missing values in an array

LokeshN:
The fix (#117) supports the missing values in an array by replacing the missing value with null
This feature can be enabled by setting Feature.ALLOW_MISSING_VALUES in JsonParser.


cowtowncoder:
@LokeshN Thank you -- turns out to be much simpler than what I was expecting. I went ahead and changed placement slightly to cover more cases.

PR #268, opened 2016-04-02
Append result to supplied StringBuilder instead of returning it.

mikaelstaldal:


cowtowncoder:
Looks good to me, but just one question: would it make sense to take `java.lang.Appendable` instead of `StringBuilder`? That would make this even more widely usable.

mikaelstaldal:
I considered `java.lang.Appendable`, but it lacks the `append(char[] str, int offset, int len)` method. And it forces you to handle `IOException` which is annoying.

mikaelstaldal:
It occurred to me that now when the new method have two parameters, it can have the same name as the old method `quoteAsString`.

cowtowncoder:
@mikaelstaldal Ok.

Issue #201, opened 2015-06-26
Add ability to navigate with a JsonPointer in a JsonParser


marshallpierce:
I'm mixing streaming mode and databinding so that I can deserialize only the parts of a document that I care about. It would be nice to be able to give a JsonPointer to a JsonParser rather than manually walking through the tokens to get to where I need to be.

cowtowncoder:
Good idea.

Note that #137 is sort of related, although not quite the same.

cowtowncoder:
I think that combination of `FilteringParserDelegate`, and databind level `mapper.at(...)`/`objectReader.at(...)` should suffice for now.
If additional/alternative mechanisms desired, may be re-filed with more concrete ideas of how to go about it.

PR #271, opened 2016-04-09
Issue #201 - JsonPointerBasedParser implementation

LokeshN:
JsonPointerBasedParser ( #201 ) : Wrapper class for json parsing, which takes care of moving the parser to the given pointer location.
The current implementation does not support the pointer to the particular index in an array

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/JsonPointer.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonPointer.java
@@ -263,6 +263,18 @@ public class JsonPointer
         }
         return h;
     }
+    
+    /**
+     * Check to determine whether the JsonPointer is an EMPTY JsonPointer
+     * 
+     * @return true if the JsonPointer is EMPTY, false otherwise
+     */
+    public boolean isEmpty() {
+        if (this == EMPTY) {
+            return true;
+        }
+        return false;
+    }
 
     /*
     /**********************************************************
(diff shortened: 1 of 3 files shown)


cowtowncoder:
How does this differ from `FilteringParserDelegate` paired with `JsonPointerBasedFilter`?

LokeshN:
FilteringParserDelegate with JsonPointerBasedFilter, only give the json object filtered by the json pointer.
For eg, if the json is :{'a':1,'b':[1,2,3],'c':{'d':{'a':true}},'d':null} and the json pointer is /c, the result will be
{'c':{'d':{'a':true}}}.
But with this JsonPointerBasedParser, it will start from /c and continues till the end of the document, so it will return {'c':{'d':{'a':true}},'d':null}, so the user can say parse the json document starting from this json pointer.
Let me know if this does not make sense.
Thanks.

cowtowncoder:
Ok. Yes, I can see the difference.

My main concern here is more of whether this should be added in `jackson-core` itself; partly since there is no direct integration (unless I missed something, there is no way to associate it via API), but also since I try to keep `jackson-core` small by limiting additions.
So while I can see how this could be useful for specific use case, it may be better to keep this as a separate thing. I don't know how that should be done, as there is no "jackson-core-utilities" project, or sandbox, but something like that would seem like a good route.

I would have some other questions too (does this work with arrays? It seems like only field names are checked), but the big question is whether this should be directly included in core package.

LokeshN:
I feel this is a small utility addition to the main core project. We can keep this is in core, once we have more utilities we can move those to a new core-utilities project.
Also, the implementation does not support arrays as of now, will add that feature in the coming week.

cowtowncoder:
@LokeshN I could keep it in core repo, but my concern is more with actual jar size, number of classes: and once it is in, it is unlikely it may be removed because there are users. If `jackson-core` was multi-maven-project repo, it'd be easy enough add contrib, but currently there is just one jar (and test/javadoc/source, but from same source files).

LokeshN:
@cowtowncoder OK. I have anyways updated the PR with the changes for array parsing.

brennabaker73:
Y

cowtowncoder:
I think I will not integrate this into core and it is best for now to keep at separate util.

PR #280, opened 2016-05-06
FilteringGeneratorDelegate.writeUTF8String() should delegate to writeUTF8String()

tlrx:
In its `writeUTF8String()` method, a `FilteringGeneratorDelegate` uses the delegating `writeRawUTF8String()` method that do not manage additional escaping:

```java
@Override
public void writeUTF8String(byte[] text, int offset, int length) throws IOException
{
    // not exact match, but best we can do
    if (_checkRawValueWrite()) {
        delegate.writeRawUTF8String(text, offset, length);
    }
}
```

I think it should use the `delegate.writeUTF8String()` instead so that it correctly escapes required chars:

```java
    // not exact match, but best we can do
    if (_checkRawValueWrite()) {
        delegate.writeUTF8String(text, offset, length);
    }

```

This pull request changes that and adds a small unit test.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/filter/FilteringGeneratorDelegate.java
+++ b/src/main/java/com/fasterxml/jackson/core/filter/FilteringGeneratorDelegate.java
@@ -375,7 +375,7 @@ public class FilteringGeneratorDelegate extends JsonGeneratorDelegate
     {
         // not exact match, but best we can do
         if (_checkRawValueWrite()) {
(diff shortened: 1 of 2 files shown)


cowtowncoder:
Ah yes, good catch. Cut'n paste error. Thank you for reporting this!
I'll manually backport this in 2.7 as well.

tlrx:
@cowtowncoder awesome, thanks!

Issue #15, opened 2012-05-27
Add 'readText()' method in JsonParser


cowtowncoder:
Current JsonParser.getText() requires reading of the whole JSON String value as String.
While convenient, this may not be optimal when processing large payloads.

As an alternative method, there should be something like:

  boolean readText(Writer w);

which would read JSON String value, and pass it using given Writer; but possibly in separate chunks, without aggregating it. This allows caller to do incremental processing and avoid potentially big temporary memory usage.

In addition, for non-blocking parser implementations, this method could do partial decoding, meaning that it would only parse part of textual value; return value indicating whether full contents (true) or partial content (false) was processed.

MikePieperSer:
I need to parse JSON with huge text fields (up to 500MB). Using the readText(Writer) methods still needs a lot of memory, because it reads the whole text field into memory.

Is there any plan to make this more efficient?

From code reading I would assume that giving the writer down to _finishString() could help here. Then the string finisher could use only one (some?) segment by writing it to the writer if it's full and reusing it.

cowtowncoder:
No one is working on this currently as far as I know; I do not have time to work on this now and probably not for a while (unless I'd need it myself for some reason). But anyone who wants to work on it would be more than welcome to do so!

And yes, lazy initial handling (only decoding opening quote) is intended to allow more efficient read+write operation like you suggest. There are multiple backends (byte-based UTF8, character/Reader-based, async) to consider, but implementation could be relatively simple if it just addresses 2 common ones (Reader/byte-based; maybe DataInput one -- async could not be supported anyway I suspect.

cowtowncoder:
Looks like I re-filed this as #1288; could close that but in this case I'll do the opposite, close this, older issue.

PR #284, opened 2016-05-15
issue #15 - readtext in jsonparser

LokeshN:
I have implemented the readText(writer) which reads the character segments and writes to the given writer object. Let me know if this implementation is fine or we need to further divide the character segments into smaller chunks and pass it to the writer

Note: In the testcase for the longer texts, in datainput parser somehow the writer object length does not match (I have commented that testcase for now). Let me know if some additional implementation is required for datainput.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/JsonParser.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonParser.java
@@ -1018,10 +1018,11 @@ public abstract class JsonParser
      * Method to read the textual representation of the current token in chunks and 
      * pass it to the given Writer
      * 
-     *  @return true, if the text chunks has been processed completely and passed to the given writer
-     *          false, if the text chunks are still being processed
+     *  @return The number of characters written to the Writer 
+     *  
+     *  @since 2.5
      */
-    public abstract boolean readText(Writer writer) throws IOException;
+    public abstract int readText(Writer writer) throws IOException, UnsupportedOperationException;
 
     /**
      * Method similar to {@link #getText}, but that will return
--- a/src/main/java/com/fasterxml/jackson/core/JsonParser.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonParser.java
@@ -1013,6 +1013,15 @@ public abstract class JsonParser
      * Method can be called for any token type.
      */
     public abstract String getText() throws IOException;
+    
+    /**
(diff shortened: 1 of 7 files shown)


src/main/java/com/fasterxml/jackson/core/JsonParser.java
```diff
@@ -1013,6 +1013,15 @@ public void finishToken() throws IOException {
...
     public abstract String getText() throws IOException;
+    
+    /**
+     * Method to read the textual representation of the current token in chunks and 
+     * pass it to the given Writer
+     * 
+     *  @return true, if the text chunks has been processed completely and passed to the given writer
+     *          false, if the text chunks are still being processed
+     */
+    public abstract boolean readText(Writer writer) throws IOException;
```
cowtowncoder: Please add `@since 2.8` here to indicate when method was introduced.

Also, for backwards compatibility, please make this throw `UnsupportedOperationException`, so that existing non-json implementations will not fail to load with 2.8.

cowtowncoder:
I can see the usefulness in an accessor that can access content without necessarily creating a full intermediate copy. But I was wondering whether there is benefit to adding a stateful accessor -- that is, one that needs to be called multiple times -- versus just writing it all (possibly with multiple write calls internally). Woodstox does this (via Stax2 API), as per:

http://fasterxml.github.io/stax2-api/javadoc/4.0.0/org/codehaus/stax2/XMLStreamReader2.html

and this makes internal handling simpler. It also allows for skipping internal buffering at some additional expense (if `preserveContents` is false).

The problem with stateful access is that it will now be necessary to also consider case where caller decides to move on without accessing all text; if so, offset will be left as non-zero. It also seems like more work for caller, without necessarily much benefit. Although there may be cases where caller might want more control; it's hard to know without knowing intended use case.

Anyway: if call was to get and write all contents for the current event, that'd make more sense to me.
Whether to allow for not-preserving content (which would allow true streaming handling, if some work was spent to support it internally) is an open question; at first it would be possible to simply assume that yes content is preserved, which is what your implemented does.

LokeshN:
@cowtowncoder I have made the changes accordingly.
I have changed the api now for readText(Writer) to return the number of bytes read instead of true and false as we are reading all the contents at once.
Let me know if it is fine.

LokeshN:
@cowtowncoder Closing this PR as my branch got some merge issue, will open a new one.

PR #285, opened 2016-05-18
issue #15 - readtext in jsonparser

LokeshN:
I have implemented the readText(writer) which reads the character segments and writes to the given writer object. 
Fixed the review comments from previous PR.
I have changed the api now for readText(Writer) to return the number of bytes read instead of true and false as we are reading all the contents at once.
Let me know if this is fine.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/json/ReaderBasedJsonParser.java
+++ b/src/main/java/com/fasterxml/jackson/core/json/ReaderBasedJsonParser.java
@@ -1,6 +1,7 @@
 package com.fasterxml.jackson.core.json;
 
 import java.io.*;
+import java.util.List;
 
 import com.fasterxml.jackson.core.*;
 import com.fasterxml.jackson.core.base.ParserBase;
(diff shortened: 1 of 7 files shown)


src/main/java/com/fasterxml/jackson/core/json/ReaderBasedJsonParser.java
```diff
@@ -273,6 +274,51 @@ public final String getText() throws IOException
...
+            //if there are no character segments left, then read the string from the current segment, and 
+            //write them directly to the buffer
+            writer.write(_textBuffer.getCurrentSegment(), 0, _textBuffer.getCurrentSegmentSize());
+            len += _textBuffer.getCurrentSegmentSize();
+           
+        }
+        else if(t != null) {
+            switch (t.id()) {
+            case ID_FIELD_NAME:
+                writer.write(_parsingContext.getCurrentName());
```
cowtowncoder: These cases also need to indicate amount of content written (`len`)

cowtowncoder:
Looks good; I can add `len` handling as per comments.

cowtowncoder:
Actually I'll go ahead and rename this method as `getText()` to be bit more consistent with existing API: it's more of an overload than related to other `readXxx()` method; latter are related to callbacks to object binder (`readValueAs()`).

Also I need to add default impl as well as efficient definitions for all other data formats: this is why default implementation is important (instead of new abstract method), so as not to break many existing dataformat modules.

PR #287, opened 2016-05-23
Added support for parsing 'INF' as positive infinity

viirya:
Currently jackson supoorts 'Infinity' and '+Infinity' for positive infinity. However, it only supports '+INF'. If there is no special reason to skip this support, it should be better to be consistent.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/json/ReaderBasedJsonParser.java
+++ b/src/main/java/com/fasterxml/jackson/core/json/ReaderBasedJsonParser.java
@@ -1870,11 +1870,24 @@ public class ReaderBasedJsonParser // final in 2.3, earlier
             _reportError("Non-standard token 'NaN': enable JsonParser.Feature.ALLOW_NON_NUMERIC_NUMBERS to allow");
             break;
         case 'I':
-            _matchToken("Infinity", 1);
-            if (isEnabled(Feature.ALLOW_NON_NUMERIC_NUMBERS)) {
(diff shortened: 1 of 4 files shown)


cowtowncoder:
@viirya I would like to know where such value is used first, if you happen to have a link? I try to avoid adding speculative features, since there are potentially unlimited number of things that might be used; and specifically so in case of deviations from JSON specification.
I don't have objection to addition itself, if such usage exists in the wild.

viirya:
I am working on [this](https://github.com/apache/spark/pull/9759) to make allowNonNumericNumbers option work on Spark JSON datasource. I found that for positive infinity, 'Infinity', '+Infinity' and '+INF' works but 'INF' doesn't. This is an inconsistent behavior as I think. Although I don't think this update can be included soon, to have consistent behavior might be better.

cowtowncoder:
@viirya just because set of value is not symmetric does not have much relevance. Rather I'd want to keep set minimal with respect to values that are actually used. JDK, for example, allows smaller set of not-a-number values; XML Schema specifies an overlap. These are sources from which currently supported values are added.
Since these values are NOT valid JSON to begin with, I want to keep the set of values as small as possible.

So: if there is actual -- not potential or possible -- usage, new values may be added.

viirya:
Ok, got it. Thanks @cowtowncoder. I would close this since you have the concern.

cowtowncoder:
@viirya Thanks!

PR #288, opened 2016-05-24
JsonWriteContext#writeFieldName not updating _currentName correctly

mohitchugh:
Currently, for a JSON like this -

`{
"a": {
},
"b": {
"c": "cval"
}
}`

When writeFieldName is called for node 'b', the _currentName is not updated
as the _gotName value is already set to true (because of the earlier processing
done for 'a'). This means that later down the line, when I try to do a
context.getParent().getCurrentName(), I get 'a' back, instead of 'b'.

This PR tentatively fixes that by pulling the setting of the _currentName earlier
before the function exits.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/json/JsonWriteContext.java
+++ b/src/main/java/com/fasterxml/jackson/core/json/JsonWriteContext.java
@@ -162,11 +162,13 @@ public class JsonWriteContext extends JsonStreamContext
      * @return Index of the field entry (0-based)
      */
     public int writeFieldName(String name) throws JsonProcessingException {
+        if (_type == TYPE_OBJECT) {
+            _currentName = name;
+        }
         if ((_type != TYPE_OBJECT) || _gotName) {
             return STATUS_EXPECT_VALUE;
         }
         _gotName = true;
-        _currentName = name;
         if (_dups != null) { _checkDup(_dups, name); }
         return (_index < 0) ? STATUS_OK_AS_IS : STATUS_OK_AFTER_COMMA;
     }


cowtowncoder:
What would be most useful as the first step would actually be a unit test?

cowtowncoder:
Hmmh. `_gotName` should be cleared when `writeValue()` is called, so name should be properly set. And in fact if name is not set, an exception should be indicated by caller.

Code paths vary a bit between `Writer` and `OutputStream` backed generators so there may be some difference there.

mohitchugh:
Thanks for your prompt response!

Yes. I can see that writeValue() should clear the _gotName flag and in a test I wrote using UTF8JsonGenerator, I cannot replicate the problem.

E.g. 
com.fasterxml.jackson.core.json.JsonWriteContext.writeValue(JsonWriteContext.java:191)
com.fasterxml.jackson.core.json.UTF8JsonGenerator._verifyValueWrite(UTF8JsonGenerator.java:942)
com.fasterxml.jackson.core.json.UTF8JsonGenerator.writeStartObject(UTF8JsonGenerator.java:300)
com.fasterxml.jackson.core.json.GeneratorBasicTest.testParentContext(GeneratorBasicTest.java:168)

However, my use-case is using an ObjectMapper.valueToTree to convert an object to JSON (where I have a filter in place to evaluate the full-path of a node and filter it out accordingly).

When I use ObjectMapper (with or without the filter) it seems to be using the databind.util.TokenBuffer class which is making use of JsonWriteContext but doesn't seem to call writeValue in its equivalent implementation of writeStartObject and thus the _gotName doesn't get cleared. 

E.g. 

com.fasterxml.jackson.databind.util.TokenBuffer.writeStartObject(TokenBuffer.java:651)
com.fasterxml.jackson.databind.ser.std.MapSerializer.serialize(MapSerializer.java:515)
com.fasterxml.jackson.databind.ser.std.MapSerializer.serialize(MapSerializer.java:30)
com.fasterxml.jackson.databind.ser.std.MapSerializer.serializeFields(MapSerializer.java:633)
com.fasterxml.jackson.databind.ser.std.MapSerializer.serialize(MapSerializer.java:536)
com.fasterxml.jackson.databind.ser.std.MapSerializer.serialize(MapSerializer.java:30)
com.fasterxml.jackson.databind.ser.DefaultSerializerProvider.serializeValue(DefaultSerializerProvider.java:292)
com.fasterxml.jackson.databind.ObjectMapper.writeValue(ObjectMapper.java:2484)
com.fasterxml.jackson.databind.ObjectMapper.valueToTree(ObjectMapper.java:2626)

It is quite possible that the correct fix might be to modify TokenBuffer instead of JsonWriteContext, but I don't have enough context about the former class to propose a change there hence a more limited change in the PR.

I agree that submitting a unit-test along with change would be more useful, however considering that my use-case is going via jackson-databind, it is making it awkward for me to put both in the same PR.

If it helps, the following code is enough to surface the problem - 

`
Map<String, Object> inner = new TreeMap<>();
inner.put("c", "cval");
Map<String, Object> ret = new TreeMap<>();
ret.put("a", new TreeMap<String, Object>());
ret.put("b", inner);
ObjectMapper objectMapper = new ObjectMapper();
objectMapper.valueToTree(ret);
`

You'll see that when writeFieldValue is called for 'b', _gotName is already true.

cowtowncoder:
I'd be perfectly fine with test case against `jackson-databind`; I can figure out call sequence fix(es) for `TokenBuffer`?

mohitchugh:
Thanks. I just opened PR - https://github.com/FasterXML/jackson-databind/pull/1252 which illustrates the issue.

Unfortunately, I just saw the earlier part of your comment in email (which you probably elaborated on later and github didn't send out an email for) and ended up adding a more complex repro than just using TokenBuffer. It does illustrate the problem quite clearly though so hopefully it'll suffice.

mohitchugh:
Also added the TokenBuffer based test case as you suggested.

cowtowncoder:
Created new bug for databind, checked in fix (to be in 2.7.5). I feel bit torn about forcible updating field name; sometimes that would make sense, but in other cases it would remove information that could be useful for error message. In case of `TokenBuffer` we do not (I think) want to necessarily enforce constraints because fragment need not be valid (for example, could construct first half of JSON Object or Array separate from rest), so it is not necessarily possible to know cases where things are fine, from within context object.

So... for now I fixed `TokenBuffer`; if this issue resurfaces I am willing to consider changes to context update. Could, for example, have additional method for simple forcible setting or something.

PR #309, opened 2016-08-08
Added writeString(Reader, int) method to JsonGenerator as requested i…

### LINKED ISSUE #17: Add 'JsonGenerator.writeString(Reader r, int charLength)'
Current 'writeString()' method requires text value to be passed in its entirety. This may not work well with larger payloads; so we should an alternative method where one can define source for text, used for incremental/streaming read+write.
In addition to Reader, it should be possible to indicate exact number of bytes to include; this because some formats (not JSON however!) require such information to be written before data, or used for initializing write. It should be possible to pass marker value -1 to indicate "not known": not all generators are required to support it, however.

If length is indicated, it is an error for Reader to provide less data (more is ok; generator just will not read anything beyond indicated amount).

terziele:
Resolve #17 issue.
Well, at least I think so. :)

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
@@ -32,7 +32,7 @@ public abstract class JsonGenerator
      */
     public enum Feature {
         // // Low-level I/O / content features
-        
+
         /**
          * Feature that determines whether generator will automatically
          * close underlying output target that is NOT owned by the
@@ -74,7 +74,7 @@ public abstract class JsonGenerator
         FLUSH_PASSED_TO_STREAM(true),
 
         // // Quoting-related features
-        
+
         /**
          * Feature that determines whether JSON Object field names are
          * quoted using double-quotes, as specified by JSON specification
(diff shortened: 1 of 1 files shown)


cowtowncoder:
I don't think it works... which is also why unit test(s) would absolutely be required here. :)

src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
```diff
@@ -933,6 +933,34 @@ public void writeArray(double[] array, int offset, int length) throws IOExceptio
...
+     * @param reader
+     * @param charLength if this parameter is -1 that means we have a unknown
+     *                   size of char array. But all generators are required to support it.
+     *                   If length is indicated, it is an error for Reader to provide less data
+     *                   (more is ok; generator just will not read anything beyond indicated amount).
+     * @throws IOException
+     */
+    public void writeString(Reader reader, int charLength) throws IOException {
+        if (reader == null) {
+            return;
```
cowtowncoder: This should be an error probably ,not just bailing out.

src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
```diff
@@ -933,6 +933,34 @@ public void writeArray(double[] array, int offset, int length) throws IOExceptio
...
+        }
+
+        if (charLength == -1) {
+            while (reader.read() != -1) ;
```
cowtowncoder: Uh, what? If not known, quietly swallow content?!?!? And do that as inefficiently as possible, char by char, instead of skipping?

src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
```diff
@@ -933,6 +933,34 @@ public void writeArray(double[] array, int offset, int length) throws IOExceptio
...
+        } else {
+            int readChars = reader.read(new char[charLength]);
+
+            if (readChars < charLength) {
```
cowtowncoder: ,... and readers are entirely within their rights to return less.

cowtowncoder:
Uh no. This makes no sense as implementation, no unit test. What the hell.

PR #312, opened 2016-08-19
Enable Jackson users to strip security-sensitive data from parser errors.

AlexYursha:
By default Jackson parser exception messages will contain excerpts from original data to provide some context about the location of the error. That is understandable but there are use cases when such excerpts contain security-sensitive data. Still the rest of Jackson exception information is very useful if only we could disable/remove the data contained in the `_location` field.


cowtowncoder:
This can definitely be added for 2.9 (since it's API addition), hoping to switch `master` soon.

Another possibility (possibly complementary, not either-or) would be to add a feature to prevent assignment of location in the first place (just assigning `JsonLocation.NA`), something `JsonParser.Feature.INCLUDE_LOCATION_IN_EXCEPTIONS`, enabled by default.

PR #320, opened 2016-09-22
Migrate away from a magic slash to a constant.

DaveJarvis:
My code currently resembles:

    return '/' + reference.replace( getConfigSeparator(), '/' );

I'd rather that it resemble:

    import static JsonPointer.SEPARATOR;
    return SEPARATOR + reference.replace( getConfigSeparator(), SEPARATOR );

That way if JsonPointer ever changed its segment separator character (or added a configuration option for changing it), then my code won't have to change. Also, making a single location for the slash character follows the DRY principle while avoiding a magic value littered throughout the code.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/JsonPointer.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonPointer.java
@@ -20,6 +20,11 @@ import com.fasterxml.jackson.core.io.NumberInput;
 public class JsonPointer
 {
     /**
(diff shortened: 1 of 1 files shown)


cowtowncoder:
Ah. At first I thought you were suggesting changing implementation more configurable (which I don't think makes sense since I doubt spec will ever change the separator). But I don't have anything against adding a constant (and possibly also using internally as beneficial). I'll just add a `@since` tag.

PR #321, opened 2016-09-26
Refactored Long Method 

feroozkhanchintu:
Refactored long method : com.fasterxml.jackson.core.json.ReaderBasedJsonParser.java


cowtowncoder:
Why?

feroozkhanchintu:
The method was very long. Broke the method down to improve the readability

cowtowncoder:
Thanks but I'll skip it: this is part of critical path, and without full performance analysis I don't want to make a change that has only aesthetic value.

Issue #323, opened 2016-10-17
Add `JsonParser.ALLOW_TRAILING_COMMA` to work for Arrays and Objects


bdhess:
Not sure if the inconsistency is intentional?

Seems weird that the feature permits `["a","b",]` but not `{"a": 1, "b": 2, }`.

bdhess:
See also  #118

cowtowncoder:
Did you read documentation of this feature?

bdhess:
Yes-- the behavior is consistent with its documentation.  However...

I submitted a pull request in #234 that covers the trailing commas use case for both arrays and objects, which I closed after you mentioned that this feature had been merged.  I'd really like some level of support for trailing commas in objects as well.  I'm happy to reopen the pull request and/or take another approach that you suggest.

cowtowncoder:
@bdhess Hmmh. Ok. So, the issue as I recall is/was that whereas with arrays it is possible to stuff add logical `null`s where missing value is, but with Objects this can not be done. So the question would be what the behavior is then.
I think one possibility considered was to perhaps allow just trailing comma.

I am open to improvements here; perhaps a separate feature would make sense if we can figure out reasonable behavior and implementation for backend.

bdhess:
Agree that there's no logical parallel to inserting `null`s into the Objects, so I think you're right to push this toward a separate feature.

I would propose that a feature `ALLOW_TRAILING_COMMAS` be defined as allowing a single trailing comma at the end of an array or object definition.  Below I've outlined how I it should intersect with `ALLOW_MISSING_VALUES` in practice.

Based on the work I did previously in #234, I think it would be straightforward to implement this way, so if you're okay with this outline in principle, I'd be happy to dust that review off and implement with the below as test cases.

## Default
`["a", "b"]` => `["a", "b"]`
`[,"a","b"]` => invalid
`["a",,"b"]` => invalid
`["a","b",]` => invalid
`["a",,"b",]` => invalid
`["a","b",,]` => invalid

`{"a": true, "b": false}` => `{"a": true, "b": false}`
`{,"a": true, "b": false}` => invalid
`{"a": true,, "b": false}` => invalid
`{"a": true, "b": false,}` => invalid
`{"a": true,, "b": false,}` => invalid
`{"a": true, "b": false,,}` => invalid

## `ALLOW_MISSING_VALUES` enabled
`["a", "b"]` => `["a", "b"]`
`[,"a","b"]` => `[null, "a", "b"]`
[...]

PR #327, opened 2016-10-29
Add JsonParser feature to ignore a trailing comma (fixes #118, #323)

### LINKED ISSUE #118: feature to allow trailing commas (non-compliant JSON)

bdhess:
Design discussion in #323 

Diff of the before/after benchmark results:
https://github.com/bdhess/jackson-benchmark-logs/commit/590fa18fe6312a2af611465fb873bef015008982?diff=split

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/JsonParser.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonParser.java
@@ -227,7 +227,30 @@ public abstract class JsonParser
           * 
           * @since 2.8
           */
-         ALLOW_MISSING_VALUES(false)
+         ALLOW_MISSING_VALUES(false),
+
+         /**
+          * Feature that determines whether {@link JsonParser} will allow for a single trailing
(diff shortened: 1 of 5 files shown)


cowtowncoder:
@bdhess I think this is pretty good; I may make some minor changes but nothing major. Thank you for doing benchmarks too; from quick glance numbers looked quite similar. I can also run tests locally myself to double-check, but from changes I think we should be good.

So: I'd be happy to merge this PR to be included in `master` for 2.9.

One last thing however: before merging the first contribution, we need CLA (unless I already asked and got one; I tried to verify). It's here:

and the usual way is to print, fill & sign, scan/take photo, email to `info` at fasterxml dot com.
Once we get this, I'll go ahead and merge this.

Looking forward to merging!

PR #333, opened 2016-11-13
Fix #332

### LINKED ISSUE #332: get rid of String.intern
[InternCache](https://github.com/FasterXML/jackson-core/blob/fa64390b1bd5f1435daa9d2b17a58594cfb22817/src/main/java/com/fasterxml/jackson/core/util/InternCache.java) is used in jackson for json property names only and solves the problem of extra memory footprint.
In the same time it uses `String.intern` - that is not an issue rather than misusage of it. The purpose of using string.intern is jvm specific: usually it maintains string literal pool (and covers internal jvm cases). 

There are several known drawbacks of using intern: 
- [String pool Hashtable](http://hg.openjdk.java.net/jdk7/jdk7/hotspot/file/924777755fad/src/share/vm/classfile/symbolTable.hpp#l73) is non-resizable table. That means: we suffer of hash code collisions when table will have items more than its size (suffer as much as many items in string.intern pool)
- The other drawback is involve GC into that: string intern pool is subject for minor collection too. So, app suffers of using string.intern implicitly via GC phases.

There is no use cases of check the equality of strings using `==` and therefore no any reasons of using `String.intern` - the biggest profit is to have string deduplication (already achieved). 

## patch
```patch
===================================================================
--- src/main/java/com/fasterxml/jackson/core/util/InternCache.java  (revision 489becbfb28a41980f0d5147d6069b30fa3b5864)
+++ src/main/java/com/fasterxml/jackson/core/util/InternCache.java  (revision )
@@ -58,7 +58,7 @@
                 }
             }
         }
-        result = input.intern();
+        result = input;
         put(result, result);
         return result;
     }
```

## Test case

The general idea behind syntenic json is used in test (a flat json structure with 10000 properties with prefix **someQName**) is provide lots of different property names to trigger usage of `string.intern` within `InternCache` - that is close to real apps use cases where amount of unique property names is on a scale from hundreds to thousands.  
```json
{
    "someQName0": 0,
    "someQName1": 1,
    ....
    "someQName9999": 9999
}
``` 

### JMH test
Let's measure performance of a single json parse with interting and w/o it: [PerfInternCache.java](https://gist.github.com/vladimirdolzhenko/4116b4042f9abff0b03570984cb92346)

```diff
Benchmark                   Mode  Cnt     Score     Error  Units
-PerfInternCache.intern      avgt    5  4098.696 ± 164.484  us/op
+PerfInternCache.noIntern    avgt    5  2320.159 ± 204.301  us/op
```

### GC

Another test is to measure how intern is affecting implicitly via GC: handle 10k the same jsons as we use in previous test (the use case is very very close to real one in real apps like web service / microsevices): 
[InternCache GC timings java test](https://gist.github.com/vladimirdolzhenko/4f50037e09aac8fa72e61021019e1eb5)

Run it with `-verbose:gc -XX:+PrintGCDetails -Xloggc:gc.log` 

and after that get the total GC pause time 
`$ grep "GC" gc.log | grep "Times: " | awk '{S+=$8}END{print S}'`

```diff
-intern      0.1907254 +- 0.00469 sec
+w/o intern  0.07665   +- 0.00498 sec
``` 

# Conclusion
Using **intern** harms application performance as explicitly via more expencive `InternCache.intern` and implicitly via GC. In the same time we keep memory footprint on the same low possible level.

vladimirdolzhenko:
get rid of String.intern

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/util/InternCache.java
+++ b/src/main/java/com/fasterxml/jackson/core/util/InternCache.java
@@ -1,15 +1,13 @@
 package com.fasterxml.jackson.core.util;
 
-import java.util.LinkedHashMap;
 import java.util.concurrent.ConcurrentHashMap;
 
 /**
- * Singleton class that adds a simple first-level cache in front of
- * regular String.intern() functionality. This is done as a minor
- * performance optimization, to avoid calling native intern() method
+ * Singleton class that adds string deduplication functionality.
+ * This is done as a minor performance optimization, to avoid extra memory footprint
  * in cases where same String is being interned multiple times.
  *<p>
- * Note: that this class extends {@link LinkedHashMap} is an implementation
+ * Note: that this class extends {@link ConcurrentHashMap} is an implementation
  * detail -- no code should ever directly call Map methods.
  */
 @SuppressWarnings("serial")
@@ -58,7 +56,7 @@ public final class InternCache
                 }
             }
         }
-        result = input.intern();
+        result = input;
         put(result, result);
         return result;
     }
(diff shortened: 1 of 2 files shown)


cowtowncoder:
No thanks, as per my comments on #332.

PR #342, opened 2016-12-19
Fixes #340

### LINKED ISSUE #340: [QUESTION] Extending WriterBasedJsonGenerator
Any reason for WriterBasedJsonGenerator being final? "Overriding" any of the public/protected methods is possible only by copying the whole class and applying the mod..

rfoltyns:
Removed 'final' keyword from WritedBasedJsonGenerator to allow to extend it and comply with other extensions of JsonGenerator and JsonParser.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/json/WriterBasedJsonGenerator.java
+++ b/src/main/java/com/fasterxml/jackson/core/json/WriterBasedJsonGenerator.java
@@ -11,7 +11,7 @@ import com.fasterxml.jackson.core.io.*;
  * {@link JsonGenerator} that outputs JSON content using a {@link java.io.Writer}
  * which handles character encoding.
  */
-public final class WriterBasedJsonGenerator
+public class WriterBasedJsonGenerator
     extends JsonGeneratorImpl
 {
     final protected static int SHORT_WRITE = 32;


cowtowncoder:
Thanks -- I'll see what to do with this. Will see if addition of `final` for some of the methods makes sense instead.

rfoltyns:
There are some internal methods still being non-final - it differs from UTF8JsonGenerator where they're `protected final`. I'll add `final` there today.

cowtowncoder:
@rfoltyns I think it would make sense to leave `_verifyValueWrite()` overridable too then.

rfoltyns:
@cowtowncoder Would you allow the PR for the overridable `_verifyValueWrite()` before the 2.9 release?

cowtowncoder:
@rfoltyns Possibly. But I would like to know expected usage for this...

rfoltyns:
@cowtowncoder Certainly. It's about changing value separators.

Currently, the only way of doing that is to configure a custom PrettyPrinter. Other ways require changes:
* making the method overridable
* making all separators configurable via API 
* making all separators configurable via extracted class (similar to `CharacterEscapes` with `CharTypes`)

Actually I suddenly became a fan of the third option.. that would require much more changes, hence shouldn't be rushed into 2.9, but seems to be much more flexible.

What do you think?

cowtowncoder:
I would think custom `PrettyPrinter` is superior than sub-classing, from design perspective.
One problem with sub-classing based configuration is its inherent fragility.

I am bit ambivalent on third option as general thing... But! If that would be done via `PrettyPrinter`, I would be +1 for it. That sounds like something worth pursuin.

The reason I don't like the idea of additional separate configuration is that this would complicate code in multiple aspects; additional API, configuration state to keep; but it would also be somewhat of a performance concern -- modern cpus tend to incur significant penalty from additional checks and branching. So I like checks that are coarse, and in this case there are already 2 distinct code paths; one without `PrettyPrinter`, other with it. Making second one more elaborate seems better, and actually may not necessarily complicate things much as we could add all separators as `SerializableString`s. Custom ones would not differ much from default ones.

Would this make sense? It could be done by extending `DefaultPrettyPrinter`, I think; touch `PrettyPrinter` API would be problematic. But I don't think that'd be needed.
`DefaultPrettyPrinter` already allows configuring of root value separator, so adding others seems doable.

I would be interested in PRs, and changes could make it in 2.9.

rfoltyns:
I'm with you on the branching thing, but even with externalized class there's no need to add any more checks. What do you think about [this draft](https://github.com/rfoltyns/jackson-core/commit/902a244ce390b2aca891a3560aae9bf7c519c01f).

* `withCustomSeparators` method added in `DefaultPrettyPrinter` as well as in `MinimalPrettyPrinter`
* current separators extracted to defaults
* all separators as `char` (`SerializableString` or `String` would have to go through more complex `writeRaw(String)` method)
* as it doesn't require any additional checks, it can be added to `JsonGenerator`s in the future without performance concerns
* leaves space for `rootValueSeparator`. I haven't done anything with `rootValueSeparator` to keep it as simple as possible. Plus, it would have to be done in steps - add, deprecate, move completely

Could something like that make a PR?

PR #343, opened 2016-12-27
handle byte[] and char[] sourceRef's specially in JsonLocation.toString()

laszlovandenhoek:
Currently, when attempting to parse a syntactically incorrect byte array as JSON using `ObjectMapper`.`readValue`, the `JsonParseException` that is thrown from `ParserBase`.`_reportMismatchedEndMarker` looks as follows:

> Invalid Json: Unexpected close marker &#x27;}&#x27;: expected &#x27;]&#x27; (for ROOT starting at [Source: [B@4928787c; line: 1, column: 0])
 at [Source: [B@4928787c; line: 1, column: 2]

Since the "Source" in question is a byte array, which in this context can safely be assumed to be a UTF-8 stream, I would much prefer to have the original source that caused this exception included in my Exception, instead of a non-descript default "`[B@deadbeef`" hash. This eases debugging, especially if the source data cannot be deduced otherwise.

I believe the intent of the original code is like that, considering the cover-all `.toString()` call in the final `} else {` block.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/JsonLocation.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonLocation.java
@@ -96,6 +96,10 @@ public class JsonLocation
         sb.append("[Source: ");
         if (_sourceRef == null) {
             sb.append("UNKNOWN");
+        } else if (_sourceRef instanceof byte[]) {
+            sb.append(new String((byte[]) _sourceRef));
+        } else if (_sourceRef instanceof char[]) {
+            sb.append((char[]) _sourceRef);
         } else {
             sb.append(_sourceRef.toString());
         }


laszlovandenhoek:
Also, where do I send my signed CLA? If I release the above code into the public domain anyway, is that even necessary?

cowtowncoder:
Apologies for slow response: I was out on vacation.

Since the patch is small, I think in this specific case I can use it without CLA, if you add a comment stating that you'd like this contribution to  be in public domain.

As to the patch; while I think it makes sense, I have some concern about printing the whole contents.
It might make sense to (for example) truncate contents. I realize that for `String` source this isn't being done anyway, but it seems like a good way. Another question is whether there are security concerns; but in general code probably should not be just printing out `toString()` of location. Challenge here is that it is actually being used by exceptions which do include output as is.

I will bring this up on `jackson-dev` list and see how others feel about it. Changes, regardless, will need to be made in `master` (for 2.9) since this is a non-trivial change to observed API.

cowtowncoder:
@laszlovandenhoek I decided to work a bit more on this one, and result (via #356) will be in 2.9.0.

In addition to indicating content (snippet, limited to first 500 characters), type is now indicated for all sources. May also add content from other source references where sensible.

PR #353, opened 2017-02-09
Alternate approach for issue #17 

uhhhh2:
I've created a possible approach for addressing Issue #17 . 

First, I added the abstract method declaration in JsonGenerator. Next, I added implementations of this method in two of the JsonGenerator implementations. Iin one implementation of the JsonGenerator (WriterBasedJsonGenerator), I created a separate buffer field in the class (using new char[] to avoid calling the same IOContext.allocX method twice which apparently results in an exception). In another implementation of the JsonGenerator (UTF8JsonGenerator), I tried to reuse the (apparently pre-existing) _charBuffer.  I am not sure if these are the best approaches for getting a char[] for use as a temporary buffer for reading from a Reader. Then, I added an implementation in the JsonGeneratorDelegate that  calls the appropriate function of the delegate instance. 

As for a unit test, I copied one of the pre-existing unit test files (StringGenerationTest) and created a new version (StringGenerationReaderTest to avoid deleting the original tests) In StringGenerationReaderTest, the boolean parameters to the methods are removed (as it only tests using a Reader). Also, each function calls 
```
StringReader reader = new StringReader(text); gen.writeString(reader, -1)
```
where it would previously call 
```
gen.writeString(text); 
```
where text is the name of the String variable with the text to write. 

Let me know if there is anything I can do to further test and/or improve this approach.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
@@ -4,7 +4,9 @@
  */
 package com.fasterxml.jackson.core;
 
-import static com.fasterxml.jackson.core.JsonTokenId.*;
+import com.fasterxml.jackson.core.JsonParser.NumberType;
+import com.fasterxml.jackson.core.io.CharacterEscapes;
+import com.fasterxml.jackson.core.util.VersionUtil;
 
 import java.io.*;
 import java.math.BigDecimal;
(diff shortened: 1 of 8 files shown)


uhhhh2:
I've edited the code so that it throws on a null reader or after reading too little. I've added some more unit tests (adapted from GeneratorFailTest and saved as GeneratorFailFromReaderTest) to test these types of failures.

uhhhh2:
Also, where do I send the CLA?

cowtowncoder:
Looks good, and since I got CLA will merge for 2.9.

PR #362, opened 2017-03-21
Use correct length variable

shtratos:
This is probably a typo because previously
the condition would always evaluate to true.

It's unlikely to have caused any issues as it
would only manifest in the unlikely case
when `_charBuffer` has zero length.

Issue found by [lgtm.com](https://lgtm.com/projects/g/FasterXML/jackson-core/snapshot/dist-1118940165-1489497197924/files/src/main/java/com/fasterxml/jackson/core/json/UTF8JsonGenerator.java#L635)

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/json/UTF8JsonGenerator.java
+++ b/src/main/java/com/fasterxml/jackson/core/json/UTF8JsonGenerator.java
@@ -632,7 +632,7 @@ public class UTF8JsonGenerator
             }
             // If this is NOT the last segment and if the last character looks like
             // split surrogate second half, drop it
-            if (len > 0) {
+            if (len2 > 0) {
                 char ch = buf[len2-1];
                 if ((ch >= SURR1_FIRST) && (ch <= SURR1_LAST)) {
                     --len2;


cowtowncoder:
Good catch, yes, `len` makes no sense. And right in fact the whole test seems pointless; `_charBuffer` really should never be empty.

But I think there's the reverse challenge, potentially: deadlock if `len` was 1.
The idea here is, I think, to try to keep surrogate pairs grouped together, by "dropping" last character to be redone next time around. This works except if content just happened to end in an unpaired surrogate char; in this case it would first be pushed from second to last segment, and then end up as the only char.

cowtowncoder:
@shtratos Thank you again -- this was wrong, and I merged similar fix in 2.8 (for 2.8.8).

shtratos:
Cool, thanks!

here's the test to reproduce the deadlock if it helps:
```java
    public void test_reproduce_deadlock() throws Exception {
        StringBuilder sb = new StringBuilder(1000);
        for (int j = 0; j < 4000; ++j) {
            sb.append(' ');
        }
        final int trailingSurrogate = 0xD800;
        sb.appendCodePoint(trailingSurrogate);
        final String text = sb.toString();

        ByteArrayOutputStream out = new ByteArrayOutputStream(1000);
        JsonGenerator g = JSON_F.createGenerator(out);
        try {
            g.writeRawValue(text, 0, text.length());
            fail("must fail!");
        } catch (JsonGenerationException e) {
            assertEquals("Split surrogate on writeRaw() input (last character): first character 0xd800", e.getMessage());
        } finally {
            g.close();
        }
    }
```

PR #365, opened 2017-04-04
Fix javadocs for getValueAsLong/Double

opinali:
Apparently simple copy/paste issue, identical to getValueAsInt but shouldn't be.


cowtowncoder:
Indeed. Thanks!

PR #368, opened 2017-04-12
Implement set/getCurrentValue

ndouba:


cowtowncoder:
Um, no. While I appreciate your suggestion, it makes sense to discuss this through mailing list first: this is implemented by sub-classes and it is not clear it should be done here; and if it should, how to refactor everything. This may be something to do with Jackson 3, but it can not be just added like this.

ndouba:
How come this base class is being used by the XmlMapper for deserialization? Currently there is no way of accessing the current value from custom deserializers?

cowtowncoder:
@ndouba Yes, it is being used. That's how inheritance and extension work. Sub-classes should implement functionality; due to backwards compatibility adding new abstract methods is not possible and as such default implementation is no-op. As per mailing list, missing implementation was recently reported and has been added to xml module impl, but not yet out (will be in  2.9).

PR #373, opened 2017-04-28
Fix #372: JsonParserSequence overrides skipChildren()

### LINKED ISSUE #372: JsonParserSequence#skipChildren() throws exception when current delegate is TokenBuffer.Parser with "incomplete" JSON
`JsonParserSequence` subclasses `JsonParserDelegate` but does not override `skipChildren()`. The default implementation of this just calls current delegate's `skipChildren()` method, which bypasses `JsonParserSequence#nextToken()`. When the current delegate is a `TokenBuffer.Parser `and the buffer contains "unbalanced" JSON (i.e. an incomplete JSON object where the number of start object tokens is more than the number of end object tokens), this violates an invariant of `ParserMinimalBase#skipChildren()` and results in an exception like the following:

```
java.lang.RuntimeException: Internal error: this code path should never get executed
	at com.fasterxml.jackson.core.util.VersionUtil.throwInternal(VersionUtil.java:179)
	at com.fasterxml.jackson.core.base.ParserMinimalBase._throwInternal(ParserMinimalBase.java:566)
	at com.fasterxml.jackson.databind.util.TokenBuffer$Parser._handleEOF(TokenBuffer.java:1662)
	at com.fasterxml.jackson.core.base.ParserMinimalBase.skipChildren(ParserMinimalBase.java:156)
	at com.fasterxml.jackson.core.util.JsonParserDelegate.skipChildren(JsonParserDelegate.java:222)
```
One possible fix for this is to have `JsonParserSequence` override `skipChildren()` in much the same way `FilteringParserDelegate` does, so that its `nextToken()` logic is not bypassed. I can submit a PR if this would be helpful, just let me know.

Thanks!

MichaelSims:
JsonParserSequence should override skipChildren(), otherwise the
delegate's implementation will call it's own nextToken(), bypassing
JsonParserSequence's implementation and potentially causing bugs.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/util/JsonParserSequence.java
+++ b/src/main/java/com/fasterxml/jackson/core/util/JsonParserSequence.java
@@ -154,6 +154,37 @@ public class JsonParserSequence extends JsonParserDelegate
         return t;
     }
 
+    /**
+     * Need to override, re-implement similar to how method defined in
+     * {@link com.fasterxml.jackson.core.base.ParserMinimalBase}, to keep
+     * state correct here.
+     */
+    @Override
+    public JsonParser skipChildren() throws IOException
+    {
+        if ((delegate.currentToken() != JsonToken.START_OBJECT)
+            && (delegate.currentToken() != JsonToken.START_ARRAY)) {
+            return this;
+        }
+        int open = 1;
+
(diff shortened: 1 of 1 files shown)


cowtowncoder:
Thanks!

Issue #330, opened 2016-11-06
`FilteringParserDelegate` seems to miss last closing `END_OBJECT`


cowtowncoder:
(note: adding a failing test for this case)

Looks like with settings like:

```java
        JsonParser p = new FilteringParserDelegate(p0,
               new NameMatchFilter("value"),
                   true, // includePath
                   false // multipleMatches
                );
```

and input

```json
{
  "a":123,
  "array":[1,2],
  "ob": {
    "value0":2,
    "value":3,
    "value2":4
  },
  "b":true
}
```

output will be like:

```json
{"ob":{"value":3}
```

(note the missing trailing `}` for closing `END_OBJECT`)

rfoltyns:
@cowtowncoder seems like `FilteringParserDelegate.nextToken` was missing the `includePath` check when `allowMultipleMatches` was `false`. That's why the it was returning prematurely with null. Will [this fix](https://github.com/rfoltyns/jackson-core/commit/efdbde902a89239fc4dc573a76eb2e425ae87f83) do the trick?

It's not the only issue I've found in the filtering, but.. one thing at a time.

cowtowncoder:
@rfoltyns I'll have to check this out. If this fix works wrt existing test it's a good start. Filtering code is unfortunately rather complicated and somewhat difficult to debug.

cowtowncoder:
@rfoltyns Applying patch, as it does fix tests. Wish I had more time to step through to verify for sure, but will have to do for now.

rfoltyns:
My patch has broken `allowMultipleMatches` functionality.. It doesn't work with `includePath: true` (still fine for false, though..).
Tests passed because they were going through the standard path again, looking for more matches, but provided JSON didn't cover this scenario.
I'll try to fix it as soon as possible, but until then, it might be safer to revert it.

rfoltyns:
I've added more tests and it turned out that it was broken before the patch. I'll create a separate issue for that.

cowtowncoder:
@rfoltyns I undid patch, moved failing tests under `failing`.

PR #381, opened 2017-05-31
#330 FilteringParserDelegate - missing end marker handling fixed

rfoltyns:
The logic here was pretty dodgy and was causing problems not only for #330 but also [here](https://github.com/FasterXML/jackson-databind/issues/1637).
Removing it solved the issue.


cowtowncoder:
Ok, I'm fine with removal since it wasn't clear what purpose it served.

One quick question: is there implied dependency between this and #380? I am planning to merge both, as soon as that one change is made to the other PR, but just wanted to know if ordering matters wrt test failures (so if both need to be applied for all to pass I know this and do not assume regression).

rfoltyns:
After a cosmetic fix (whitespace removal) in [1415e9d](https://github.com/FasterXML/jackson-core/pull/381/commits/1415e9d6db974c8db1151d74013fe48889193eb7) #380 and #381 can be applied in any order - there's no correlation between them.

PR #383, opened 2017-06-02
Fix 382 indexoutofbounds

### LINKED ISSUE #382: ArrayIndexOutOfBoundsException from UTF32Reader.read on invalid input
We are doing some fuzz testing on internal projects and found some input that causes an unexpected exception from JsonParser. While it is throwing an exception and stopping on this, it should probably be some kind of IOException rather than an ArrayIndexOutOfBoundsException.

In our case we are catching the three exception types listed in ObjectMapper.readTree  (JsonParseException | JsonProcessingException | IOException) and this managed to escape.

Simple test case follows
```java
public void testInvalidInput() throws IOException {

        byte[] data = {
                0x00,
                0x00,
                0x00,
                0x20,
                (byte) 0xFE,
                (byte) 0xFF,
                0x00,
                0x01,
                (byte) 0xFB
        };

        JsonFactory FACTORY = new JsonFactory();
        JsonParser parser = FACTORY.createParser(data);

        parser.nextToken();


    }
```

stack trace:
```
java.lang.ArrayIndexOutOfBoundsException: 9

	at com.fasterxml.jackson.core.io.UTF32Reader.read(UTF32Reader.java:138)
	at com.fasterxml.jackson.core.json.ReaderBasedJsonParser._loadMore(ReaderBasedJsonParser.java:243)
	at com.fasterxml.jackson.core.json.ReaderBasedJsonParser._skipWSOrEnd(ReaderBasedJsonParser.java:2331)
	at com.fasterxml.jackson.core.json.ReaderBasedJsonParser.nextToken(ReaderBasedJsonParser.java:646)
	at com.fasterxml.jackson.core.io.TestUTF32Reader.testInvalidInput(TestUTF32Reader.java:32)
```

wselwood:
Fix #382 by checking before reading things out of the buffer that we are not going to run off the end of the array. 

Add test case to check fix.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/io/UTF32Reader.java
+++ b/src/main/java/com/fasterxml/jackson/core/io/UTF32Reader.java
@@ -134,6 +134,10 @@ public class UTF32Reader extends Reader
             int ptr = _ptr;
             int ch;
 
+            if (_buffer.length < ptr + 4) {
+                reportUnexpectedEOF(_buffer.length - ptr, 4);
+            }
+
             if (_bigEndian) {
                 ch = (_buffer[ptr] << 24) | ((_buffer[ptr+1] & 0xFF) << 16)
                     | ((_buffer[ptr+2] & 0xFF) << 8) | (_buffer[ptr+3] & 0xFF);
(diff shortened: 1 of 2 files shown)


cowtowncoder:
Added test, fix (just not via this PR). Thank you again for your help!

PR #387, opened 2017-07-16
Increase code coverage one

TheRealHaui:
I have created Unit Tests which increase test coverage and which I want to contribute.


src/test/java/com/fasterxml/jackson/core/Base64VariantsTest.java
```diff
@@ -0,0 +1,31 @@
+package com.fasterxml.jackson.core;
```
cowtowncoder: Could this be added to one of existing base64 tests, instead of a new test class?

cowtowncoder:
First of all: thank you for contribution.

I am bit torn here: although I like higher test code coverage, I do not like having test whose only value is to increase test coverage. Many tests included here seem to be aimed solely at increasing the coverage.
So I'll have to think about this a bit.

src/test/java/com/fasterxml/jackson/core/Base64VariantsTest.java
TheRealHaui: Done.
Moved to Base64BinaryParsingTest class.

TheRealHaui:
Thank you for your response and you're welcome.

It is of course clearly your decision which code to include in your project and which not!
However, do be honest I am not sure if I really get your point entirely.

What I am trying to do is to create regression tests (https://en.wikipedia.org/wiki/Software_regression) so that in future changes of the software (bug fixes, refactoring and to a certain extend new features) new introduced bugs get detected automatically through test execution (https://en.wikipedia.org/wiki/Regression_testing#Regression_test_selection).

Therefore, to my extend it doesn't make an essential difference if those tests are created entirely during development time of the software or later to avoid regressions.

cowtowncoder:
@TheRealHaui I added some comments that should point out kinds of test cases that to me serve no value except for increasing code coverage. They explicitly do not test either valid behavior (for which regression is most valuable to test), or, specific exception/failure response for cases where it is specified.

And the fundamental thing is this: test code is not free from maintenance -- similar to non-test code, every line of code adds cost to maintenance. So all code has to serve some purpose, add value.

I may have missed some aspects of tests, reason for addition of some of the tests. So feel free to add follow up comments where this is the case.

TheRealHaui:
@cowtowncoder 
Completely agree with you regarding valid behavior, exception/failure response cases and maintenance costs!

Beside that, waiting for your comments that point out kinds of test cases that to me serve no value.

PR #388, opened 2017-07-20
test-previously-uncovered-logic

TheRealHaui:
Refering to #387 a new attempt.

### DIFF
--- a/src/test/java/com/fasterxml/jackson/core/JsonpCharacterEscapesTest.java
+++ b/src/test/java/com/fasterxml/jackson/core/JsonpCharacterEscapesTest.java
@@ -1,0 +1,31 @@
+package com.fasterxml.jackson.core;
+
+import com.fasterxml.jackson.core.io.SerializedString;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Unit tests for class {@link JsonpCharacterEscapes}.
+ *
+ * @date 2017-07-20
+ * @see JsonpCharacterEscapes
+ *
+ **/
+public class JsonpCharacterEscapesTest{
+
+    @Test
+    public void testGetEscapeSequenceOne() {
+        JsonpCharacterEscapes jsonpCharacterEscapes = JsonpCharacterEscapes.instance();
+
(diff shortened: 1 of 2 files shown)


cowtowncoder:
Sounds good, thank you!

One practical thing before I can merge this: unless I have asked for and received, we need Contributor License Agreement (CLA):

but just once -- when it is done all contributions for Jackson components are covered.

The usual way is to print, fill & sign, scan, email to `info` at fasterxml dot com.

TheRealHaui:
Great!
You'll find my eMail in your inbox.

cowtowncoder:
Excellent, thank you! It is good to improve test coverage and while 2.9 makes advances there are still some gaps.

TheRealHaui:
Great, thanks!
Well, why don't you tell me where you see the biggest gaps?
And I might take care of them. :-)
And in exchange you tell me your perspective on the tests from your business side.

cowtowncoder:
@TheRealHaui to be honest I did find code coverage tools useful in finding gaps, so I have nothing against using them. But aside from coverage things usually get tricky with combination of features; either within component or (especially) cross components.

At this point `jackson-core` may be relatively well tested compared to other components; my biggest concerns are with datatype libraries, and their handling of things like polymorphic deserialization.

But as to `jackson-core`, maybe...

1. Location handling: parsers try to keep exact track of location (esp. for tokens). Testing is quite simple, minimal documents. Longer documents, with buffer boundaries, different input sources would probably uncover minor bugs.
2. Handling of non-standard features (comments, optional quotes) in general; and then things like location handling in combination
3. Alternative accessors (access number tokens via `getText()` etc)
4. Base64 decoding, esp. cases where white space is allowed (I think some variants allow this)
5. Beyond location (line, row), handling of parsing/generation context (parent Object/Array, index), that is, `JsonStreamContext` implementations
6. Non-blocking/async parsing is completely new thing in 2.9. I tried adding reasonable testing but... well, it's alternate backend and certainly isn't as well tested as blocking ones.

I hope that gives some idea. Another possibility might be to have a look at fixed bugs, general area of functionality (more bugs fixed, probably more remaining).

PR #390, opened 2017-07-31
add-tests

TheRealHaui:
Added tests that cover previously uncovered logic.

### DIFF
--- a/src/test/java/com/fasterxml/jackson/core/format/DataFormatMatcherTest.java
+++ b/src/test/java/com/fasterxml/jackson/core/format/DataFormatMatcherTest.java
@@ -1,0 +1,46 @@
+package com.fasterxml.jackson.core.format;
+
+import com.fasterxml.jackson.core.JsonFactory;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Unit tests for class {@link DataFormatMatcher}.
+ *
+ * @date 2017-07-31
+ * @see DataFormatMatcher
+ **/
+public class DataFormatMatcherTest {
+
+  @Test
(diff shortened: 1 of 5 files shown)


src/test/java/com/fasterxml/jackson/core/format/DataFormatMatcherTest.java
```diff
@@ -0,0 +1,46 @@
...
+    MatchStrength matchStrength = MatchStrength.WEAK_MATCH;
+    DataFormatMatcher dataFormatMatcher = new DataFormatMatcher(null,
+            byteArray,
+            0,
+            (-2),
+            null,
+            matchStrength);
+    InputStream inputStream = dataFormatMatcher.getDataStream();
+
+    assertEquals((-2), inputStream.available());
```
cowtowncoder: This seems bit odd/wrong actually. `available()` should not, I think, ever return negative value... it's not the next byte (that'd be something like "peek()") but # of bytes that may be read without blocking.

cowtowncoder: could you file a bug for this one?

cowtowncoder:
Looks good, and possibly uncovered one minor flaw even!

src/test/java/com/fasterxml/jackson/core/format/DataFormatMatcherTest.java
TheRealHaui: Sure.

Considered it to be right returning a negative number.
However, read InputStream class comment of the JDK a minute ago and learned that it is not. right.

TheRealHaui: Created: https://github.com/FasterXML/jackson-core/issues/391

PR #403, opened 2017-09-18
add-new-unit-tests

TheRealHaui:
Added new Unit Tests that cover previously uncovered logic.


cowtowncoder:
Sounds good. Going forward, one request: could you make tests extend `BaseTest`, like existing tests do? There are helper methods there which are useful.

TheRealHaui:
Great.
No problem, here it comes.

cowtowncoder:
Much appreciated. WIll merge.

PR #404, opened 2017-09-19
add-tests

TheRealHaui:
Added new Unit Tests.

### DIFF
--- a/src/test/java/com/fasterxml/jackson/core/JsonGeneratorTest.java
+++ b/src/test/java/com/fasterxml/jackson/core/JsonGeneratorTest.java
@@ -1,0 +1,207 @@
+package com.fasterxml.jackson.core;
+
+import com.fasterxml.jackson.core.filter.FilteringGeneratorDelegate;
+import com.fasterxml.jackson.core.filter.TokenFilter;
+import com.fasterxml.jackson.core.io.IOContext;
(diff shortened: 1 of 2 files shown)


src/test/java/com/fasterxml/jackson/core/JsonGeneratorTest.java
```diff
@@ -0,0 +1,207 @@
...
+
+/**
+ * Unit tests for class {@link JsonGenerator}.
+ *
+ * @date 2017-09-19
+ * @see JsonGenerator
+ **/
+public class JsonGeneratorTest extends BaseTest {
+    public void testWriteObjectThrowsIllegalStateException() throws IOException {
+        BufferRecycler bufferRecycler = new BufferRecycler();
```
cowtowncoder: In general I much prefer going through public api, constructing generators and parsers using `JsonFactory`.
Only if that is not possible for some reason it should be necessary to use internal api.
One reason is that internal api will keep on changing and tests that rely on them are quite fragile, needing to be updated if and when call sequences change.

cowtowncoder:
Ok, to me these tests once again seem to be mostly focused on increasing code coverage, but not testing parts that matter. I don't really care if a code path that user should never use results in an NPE or IllegalArgumentException; or what its message is.

So... what is the point of these tests?

PR #438, opened 2018-01-12
Throw exception on attempt of using closed UTF8StreamJsonParser

### LINKED ISSUE #437: `ArrayIndexOutOfBoundsException` in `UTF8StreamJsonParser`
Using of UTF8StreamJsonParser#nextToken for closed JsonParser causes unmeaningfull ArrayIndexOutOfBoundsException exception. 

```
Caused by: java.lang.ArrayIndexOutOfBoundsException: 0
	at com.fasterxml.jackson.core.json.UTF8StreamJsonParser._skipWSOrEnd(UTF8StreamJsonParser.java:2940)
	at com.fasterxml.jackson.core.json.UTF8StreamJsonParser.nextFieldName(UTF8StreamJsonParser.java:968)
	at com.fasterxml.jackson.core.json.UTF8StreamJsonParserTest.testExceptionOnClosedParserOnNextFieldName(UTF8StreamJsonParserTest.java:19)
```

I spent quite a lot of time to debug this and propose to add a meaningfull exception for this case. The code that causes this exception looks like this
```java
// pseudocode
Iterable<SomeData> getData() {
  try {
      jsonParser = jsonFactory.createParser(stream)
      return jsonParser.readValuesAs(SomeData.class);
  } finally {
     jsonParser.close();
  }
}
```
What do you think if we could add a check for this case and throw meaningfull exception?

lislon:
Fixes #437

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/json/UTF8StreamJsonParser.java
+++ b/src/main/java/com/fasterxml/jackson/core/json/UTF8StreamJsonParser.java
@@ -669,6 +669,9 @@ public class UTF8StreamJsonParser
     @Override
     public JsonToken nextToken() throws IOException
     {
+        if (_closed) {
+            _reportIllegalReadOnClosedParser();
+        }
         /* First: field names are special -- we will always tokenize
          * (part of) value along with field name to simplify
          * state handling. If so, can and need to use secondary token:
@@ -861,6 +864,9 @@ public class UTF8StreamJsonParser
     public boolean nextFieldName(SerializableString str) throws IOException
     {
         // // // Note: most of code below is copied from nextToken()
+        if (_closed) {
+            _reportIllegalReadOnClosedParser();
+        }
         _numTypesValid = NR_UNKNOWN;
         if (_currToken == JsonToken.FIELD_NAME) { // can't have name right after name
             _nextAfterName();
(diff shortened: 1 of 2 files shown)


cowtowncoder:
I think that behavior should be identical for `ReaderBasedJsonParser` (character-based source like `Reader` or `String`), so it'd be good to test that too. And possibly `DataInput` backed one as well.

I wish I could say what the exact expected semantic is; given that one obvious alternative is to simply start returning `null`. I'll see if I can find behavior from parsers of other supported formats.

Either way, yes, current behavior is wrong.

lislon:
I noticed that the sibling class returns `null` in the same case, but that behaviour could defer the problem of using closed stream, which will popup in unrelated code later. 

On the other hand it could break BC, so maybe we could leave `null`'s in `2.9` and consider using excepitons in `3.x`?

cowtowncoder:
@lislon Yes, I think `null` is the way to go for 2.x for sure. I think there is also `isClosed()` method that callers can use.

We can discuss alternative for 3.0: I think this is where it would be useful to ask user community.
My intuition about behavior most users would find good is not very accurate.
(but then again, it is not certain it is easy or possible to get representative sample on mailing list either -- what tends to happen is that changes ALWAYS get more negative feedback from users, since those content are less likely to give feedback).

lislon:
@cowtowncoder OK, I created new PR #440 with `null`'s for `2.9`. Please let me know if there is a need to fix something in PR

cowtowncoder:
Closing in favor of #440

PR #440, opened 2018-01-15
Prevent ArrayIndexOutOfBoundsException when using closed JsonParser

lislon:
Solution with returning `null`s for #437


cowtowncoder:
Looks good, thank you!

One more thing before I will merge this in: before the first contribution (but only then), we'll need CLA:

(or its Corporate CLA counterpart, if on behalf of a company)

The usual way is to print it, fill & sign, scan, email to `info` at fasterxml dot com.
If you could do that that would be great: fix will go in `2.9.4` and should improve usability.
We can then consider handling wrt `3.0`, if it should default to exception, or, if not, should there be `JsonParser.Feature` to allow that.

PR #459, opened 2018-05-16
Fixes FasterXML/jackson-databind#2-38 by deserializing codec correctly

243826:


cowtowncoder:
Ah missed the PR while working on the problem. Thank you for reporting & PR!

Issue #57, opened 2013-02-05
Add support for non-blocking ("async") JSON parsing


cowtowncoder:
(migrated from http://jira.codehaus.org/browse/JACKSON-39 -- note, high vote count)

---

(suggested by Dimitri M on user list)

There are use cases where it'd be good to be able to feed input to parser, instead of trying to provide an input stream for parser to read from. This would cover use cases where input comes in chunks; for example, part of logical document in one chunk, then after a delay (perhaps in a separate request) another one and so forth. In these cases it may be difficult to implement InputStream (Reader etc) abstraction; instead, it would be better if application could feed (push) data to parser.
But if so, parser must be able to indicate cases where no data is YET available (but may become available).

This is similar to how Aalto Xml processor (http://www.cowtowncoder.com/hatchery/aalto/index.html) operatesd in its async mode. However, since Json is much simple format than xml, implementation might be simpler.

Based on my experiences with Aalto, implementation is a non-trivial thing however. One problem is that even UTF-8 decoding needs to be somewhat aware of chunk boundaries, so in the end a separate parser may be required: this because current parser uses blocking to handle these split cases. A smaller problem is that of indicating "not-yet-available" case – this can probably be handled by introducing a new member in JsonToken enumeration.

testn:
Is there a way to get this supported? I saw something implemented in jackson-smile.

cowtowncoder:
@testn Yes, by someone with lots of time to spend on implementing and testing it. Smile unfortunately only has skeleton (or, scaffolding), not full implementation. I know how it can be done (see `aalto-xml` for earlier implementation I did for xml), but at this point do not have time or personal need, nor a customer willing to pay for development costs. But anyone else is free to have a stab with it, it is definitely doable.

testn:
Can you describe a bit how it should be done? Maybe I will take a stab at it.

cowtowncoder:
The way I did it with Aalto:

https://github.com/FasterXML/aalto-xml

was to basically do two things:

1. Provide non-blocking content feeder interface, to use instead of blocking `InputStream` or `Reader`
2. Implement state-machine driven parser/decoder to keep byte-accurate state of things; and return `NOT_AVAILABLE` token if there isn't enough content to FULLY decode the event (unlike with blocking, there is no option of lazily decoding things)

[...]

PR #462, opened 2018-05-20
Add failing AsyncTokenFilterTest.

htmldoug:
I've been following your great work in https://github.com/FasterXML/jackson-core/issues/57 and finally found a use-case to try async parsing in one of my apps, but I think I've hit a bug.

## Summary
I'd like to use a `TokenFilter` over a `NonBlockingJsonParser` to do json projection, but wrapping a `FilteringParserDelegate` around `NonBlockingJsonParser` leads to an infinite loop when passing a complete json object over several separate chunks.

I've included a failing test in this PR.

## Infinite loop
This works fine at chunkSize 28 and 27, but we get stuck in an infinite loop at chunkSize 26. Thread point dump points to:
```
  at com.fasterxml.jackson.core.base.ParserMinimalBase.skipChildren(ParserMinimalBase.java:253)
  at com.fasterxml.jackson.core.filter.FilteringParserDelegate.nextToken(FilteringParserDelegate.java:412)
  at com.fasterxml.jackson.core.json.async.AsyncTokenFilterTest.testFilteredNonBlockingParser(AsyncTokenFilterTest.java:46)
```

## Questions
Is my usage supported, or is there another/better way to accomplish my goal?

### DIFF
--- a/src/test/java/com/fasterxml/jackson/core/json/async/AsyncTokenFilterTest.java
+++ b/src/test/java/com/fasterxml/jackson/core/json/async/AsyncTokenFilterTest.java
@@ -1,0 +1,64 @@
+package com.fasterxml.jackson.core.json.async;
+
+import com.fasterxml.jackson.core.*;
+import com.fasterxml.jackson.core.async.AsyncTestBase;
+import com.fasterxml.jackson.core.async.NonBlockingInputFeeder;
+import com.fasterxml.jackson.core.filter.FilteringParserDelegate;
+import com.fasterxml.jackson.core.filter.TokenFilter;
+import com.fasterxml.jackson.core.testsupport.AsyncReaderWrapper;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.Arrays;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertNotEquals;
+
+public class AsyncTokenFilterTest extends AsyncTestBase {
+    private final JsonFactory JSON_F = new JsonFactory();
+
+    private final static String INPUT_STRING = "{'a': 1, 'b': [2, {'c': 3}]}".replace('\'', '"');
+    private final static byte[] INPUT_BYTES = INPUT_STRING.getBytes();
+    private final static TokenFilter TOKEN_FILTER = new TokenFilter() {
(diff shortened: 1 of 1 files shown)


cowtowncoder:
It is not possible to make `skipChildren()` work for all cases of non-blocking parsing, because it may or may not have enough content to complete skipping; and if not, there's no way to retain information on what would still be needed. So skipping logic will have to be done from outside, keeping necessary state information.
However, I think that attempts to try to do that should not end in infinite cycle. Not sure if it should fail for all attempts, or such cases where `JsonToken.NOT_AVAILABLE` is returned. Maybe latter?
I will create a new issue for such improvement.

As to whether non-blocking parser and delegate can work together... that seems more plausible.
I don't think it has been tested, but to degree it is doable it would be nice to improve.
But I think unit tests for failing cases would be useful. It should not be very difficult to write some for simple cases, as variations; especially when using very small buffer (like one byte at a time).

htmldoug:
Having `skipChildren()` throw seems reasonable, or at least less surprising.

I've worked around this in my app by moving the `TokenFilter` to generation-time (`FilteringGeneratorDelegate`) instead of parse-time. I suspect it's less performant than filtering upfront, but it's good enough for now.

> I think unit tests for failing cases would be useful.

You may find the one in this PR helpful. It covers the 1 byte at a time case that you mentioned, along with several others.

> As to whether non-blocking parser and delegate can work together... that seems more plausible.

Agreed. Might need to track more state. Not sure. However, I found `FilteringParserDelegate` a bit daunting. Those changes seem like a bigger investment than I'm able to make right now.

Really appreciate the response, and thank you for making async json processing possible. Way cool. Whatever you want to do with this PR is fine by me.

cowtowncoder:
@htmldoug Excellent. I hope to make some little fixes still before 2.9.6, but will need to publish that relatively soon. Thanks for the PR, I will need to read it with thought. Should be useful.

Also thank you for checking out async parsing: I hesitated for quite a while until implementing it (based on my earlier work on async XML for `aalto-xml`), since adoption of earlier work had been lower than expected. But uptake for json seems to have been bit more rapid -- and as long as there is a steady stream of early adopters, things work out quite nicely: there's a stream of bug reports, fixes, but no torrent of duplicates. :)

Issue #478, opened 2018-09-17
Provide implementation of async JSON parser fed by `ByteBufferFeeder`


poutsma:
Currently, there is no implementation of `com.fasterxml.jackson.core.async.ByteBufferFeeder` provided in Jackson Core; there is only an implementation of `ByteArrayFeeder`. Within Spring Framework, this requires us to covert the (potentially non-heap-allocated) ByteBuffers into byte arrays before feeding them to the parser, resulting in insufficient memory usage.

Please provide an implementation of `ByteBufferFeeder`.

cowtowncoder:
That would be good addition: I always assumed there would be need for it.

Note to potential implementors: beyond feeder, bigger job is to implement actual parser that uses `ByteBuffer` as input source. That should be mostly monkey work in copying array-backed variant, changing accessors (but not logic): unfortunately more modular design that could cover both turns out to have non-negligible amount of overhead (JVM can not quite optimize access patterns to remove all call dispatch, I am guessing, or something similar).

kptfh:
Does it make sense to switch internals of NonBlockingJsonParser to ByteBuffer and add 2 new methods  
`public ByteBufferFeeder getByteBufferNonBlockingInputFeeder() ` and
`public void feedInput(ByteBuffer buf) throws IOException `
while retaining previous byte array methods for backward compatibility?

mizosoft:
Any updates on this? Having to convert `ByteBuffers` to `byte[]` is annoying :( 

It seems that https://github.com/FasterXML/jackson-core/pull/483 is hanging. Are there any performance concerns for it?

kptfh:
Successfully using https://github.com/kptfh/json-reactive in  https://github.com/Playtika/feign-reactive

mizosoft:
@kptfh Thanks. Might use your `NonBlockingJsonParser` port. Wonder why it's not merged :/

cowtowncoder:
Some would need to show that adding a layer of abstraction does not measurably (say, no more than 5% slower) slow down existing `byte[]` backed implementation; or, alternatively, provide implementation that does not require abstraction away single-byte accessors. Since single-byte access is on critical path wrt performance I wouldn't want to simply merge something that could significantly slow existing implementation.

Either approach (show suggested patch has no significant performance overhead; add more sizable complete `ByteBuffer` backed impl) would be acceptable.

On testing, there is:

https://github.com/FasterXML/jackson-benchmarks

[...]

PR #483, opened 2018-10-02
NonBlockingJsonParser implements ByteBufferFeeder

kptfh:
Should fix [https://github.com/FasterXML/jackson-core/issues/478](https://github.com/FasterXML/jackson-core/issues/478)


kptfh:
@cowtowncoder Should it add much performance penalties?

cowtowncoder:
Before considering adding one specific layer of wrapping, I would want actual performance measurements to show it does not hurt performance. Based on past experience I would expect going via `ByteBuffer` to add measurable amount of overhead (work was with `aalto-xml`).

pjfanning:
This is probably unnecessary now due to https://github.com/FasterXML/jackson-core/pull/795 - that latter PR adds dedicated byte buffer support as opposed to changing the byte array support to support buffers and arrays.

cowtowncoder:
Agreed @pjfanning, closing.

PR #491, opened 2018-10-19
in base64 encoding escape newlines only when between quotes

vboulaye:
This is an attempt to fix https://github.com/FasterXML/jackson-dataformats-text/issues/90 which seems to have been introduced when switching the default Base64Variant to MIME instead of MIME_NO_LINEFEEDS (to fix https://github.com/FasterXML/jackson-dataformats-text/issues/62)

When the encoded string is longer than 76 characters, the encoding method adds escaped '\n', instead of "real" newlines, in order to be able to include it in a quoted string. 
But in the case of yaml, the encoding is done without enclosing quotes, and "real" newlines should be used for the decoding to work (as the "\n" is are not unescaped).

With this change, the newlines are only escaped when the encoding is done between quotes.

I tested the change by building jackson-databind and jackson-dataformat-texts (I do not know which other modules could be impacted by this change). 

The additional test for encoding long arrays in jackson-dataformat-texts proposed in https://github.com/FasterXML/jackson-dataformats-text/issues/90 also works with this change.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/Base64Variant.java
+++ b/src/main/java/com/fasterxml/jackson/core/Base64Variant.java
@@ -399,9 +399,13 @@ public final class Base64Variant
             b24 = (b24 << 8) | (((int) input[inputPtr++]) & 0xFF);
             encodeBase64Chunk(sb, b24);
             if (--chunksBeforeLF <= 0) {
-                // note: must quote in JSON value, so not really useful...
-                sb.append('\\');
-                sb.append('n');
+                if (addQuotes) {
+                    // note: must quote in JSON value, so not really useful...
+                    sb.append('\\');
+                    sb.append('n');
+                } else {
+                    sb.append('\n');
+                }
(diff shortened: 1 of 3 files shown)


cowtowncoder:
I'll have to re-read description to fully understand the problem, but unfortunately I am not convinced the fix logic is safe: add-quotes does not have any semantic meaning that should make it safe (or not) of omitting escaping.
All it does is indicate, literally, whether double-quotes should be used to surround value: if not, it is quite possible caller might do enclosing, or it might be they are to be omitted. But it does not really indicate anything specific about expected handling.

Having said that I think that fix probably needs to be applied on YAML backend, if possible, even if it means code duplication (I am ok with cut'n paste as long as comments reflect where code came from).

vboulaye:
Ok, I was also afraid that it was a bit far fetched to do it here!

I'll try to do it directly in the YAMLGenerator in jackson-dataformats-text.

PR #499, opened 2018-11-25
Include Guice Module exposure for utils

GedMarc:


cowtowncoder:
One question first: it seems wrong that `jackson-core` would have to declare other packages it exports packages for. I don't know Java 9+ package system well, but that can't work well in general -- shouldn't packages just expose what they want public, in general? There should be no dependencies from `jackson-core` to other modules, not even for metadata.

So I guess I am not sure of intent to try to limit things this way.

Second: as per earlier discussions, `jackson-core` and `jackson-annotations` must produce JDK 6 bytecode at least for 2.9. Changes seem to remove definitions resulting in Java 7 baseline I think. I recall you mentioning there would be a way to allow Java 6 compatible bytecode.
If this is not possible, change needs to be done for 2.10 and can't go in 2.9.

cowtowncoder:
Reading similar changes to `jackson-databind`, and based on that it seems there is a baseline of truly open, and then limited set of openings for other modules, to try to minimize what is exposed.
I guess that makes sense, and it's just a question of limits: I think I'd want to be bit more permissive, and remove most of special cases.

So I'll see if I have comments specific to this package, first.

src/moditect/module-info.java
```diff
@@ -0,0 +1,11 @@
+module com.fasterxml.jackson.core {
+	exports com.fasterxml.jackson.core;
+
+	exports com.fasterxml.jackson.core.type to com.fasterxml.jackson.databind, com.jwebmp.core, com.fasterxml.jackson.datatype.jsr310;
```
cowtowncoder: Should be open to all: `TypeReference` (for example) is very widely used and essential part of API.

src/moditect/module-info.java
```diff
@@ -0,0 +1,11 @@
...
+	exports com.fasterxml.jackson.core.io to com.fasterxml.jackson.databind, com.fasterxml.jackson.module.afterburner, com.fasterxml.jackson.datatype.jdk8;
```
cowtowncoder: Also should be open to all: widely enough used, esp. something like `JsonEOFException`. Going forward could be refactored (for 3.0) to divide into public and internal types, but for now I think needs to be public.

src/moditect/module-info.java
```diff
@@ -0,0 +1,11 @@
...
+	exports com.fasterxml.jackson.core.util to com.fasterxml.jackson.databind, com.fasterxml.jackson.module.paranamer, com.fasterxml.jackson.module.parameternames, com.fasterxml.jackson.module.osgi, com.fasterxml.jackson.module.mrbean, com.fasterxml.jackson.module.jaxb, com.fasterxml.jackson.module.afterburner, com.fasterxml.jackson.datatype.jsr310, com.fasterxml.jackson.datatype.jdk8, com.fasterxml.jackson.module.guice;
+	exports com.fasterxml.jackson.core.base  to com.fasterxml.jackson.databind;
+	exports com.fasterxml.jackson.core.json  to com.fasterxml.jackson.databind;
+	exports com.fasterxml.jackson.core.filter to com.fasterxml.jackson.databind;
+	exports com.fasterxml.jackson.core.format  to com.fasterxml.jackson.databind;
+}
```
cowtowncoder: In addition to these, `com.fasterxml.jackson.core.async` needs to be exported to all.

src/moditect/module-info.java
```diff
@@ -0,0 +1,11 @@
+module com.fasterxml.jackson.core {
+	exports com.fasterxml.jackson.core;
+
+	exports com.fasterxml.jackson.core.type to com.fasterxml.jackson.databind, com.jwebmp.core, com.fasterxml.jackson.datatype.jsr310;
+	exports com.fasterxml.jackson.core.io to com.fasterxml.jackson.databind, com.fasterxml.jackson.module.afterburner, com.fasterxml.jackson.datatype.jdk8;
+	exports com.fasterxml.jackson.core.util to com.fasterxml.jackson.databind, com.fasterxml.jackson.module.paranamer, com.fasterxml.jackson.module.parameternames, com.fasterxml.jackson.module.osgi, com.fasterxml.jackson.module.mrbean, com.fasterxml.jackson.module.jaxb, com.fasterxml.jackson.module.afterburner, com.fasterxml.jackson.datatype.jsr310, com.fasterxml.jackson.datatype.jdk8, com.fasterxml.jackson.module.guice;
+	exports com.fasterxml.jackson.core.base  to com.fasterxml.jackson.databind;
```
cowtowncoder: Needs to be exported to all as most are needed by dataformat implementations.

src/moditect/module-info.java
```diff
@@ -0,0 +1,11 @@
+module com.fasterxml.jackson.core {
+	exports com.fasterxml.jackson.core;
+
+	exports com.fasterxml.jackson.core.type to com.fasterxml.jackson.databind, com.jwebmp.core, com.fasterxml.jackson.datatype.jsr310;
+	exports com.fasterxml.jackson.core.io to com.fasterxml.jackson.databind, com.fasterxml.jackson.module.afterburner, com.fasterxml.jackson.datatype.jdk8;
+	exports com.fasterxml.jackson.core.util to com.fasterxml.jackson.databind, com.fasterxml.jackson.module.paranamer, com.fasterxml.jackson.module.parameternames, com.fasterxml.jackson.module.osgi, com.fasterxml.jackson.module.mrbean, com.fasterxml.jackson.module.jaxb, com.fasterxml.jackson.module.afterburner, com.fasterxml.jackson.datatype.jsr310, com.fasterxml.jackson.datatype.jdk8, com.fasterxml.jackson.module.guice;
```
cowtowncoder: Some of the types needed widely (about half), so needs to be exported to all. May want to refactor for 3.0 to hide some implementations.

src/moditect/module-info.java
```diff
@@ -0,0 +1,11 @@
...
+	exports com.fasterxml.jackson.core.base  to com.fasterxml.jackson.databind;
+	exports com.fasterxml.jackson.core.json  to com.fasterxml.jackson.databind;
```
cowtowncoder: Eventually might be nice to close, but for now needs to be exported: at least `DupDetector` is needed by other packages, and I think `JsonReadContext` / `JsonWriteContext` may be exposed (until 3.0).
So, export to all.

src/moditect/module-info.java
```diff
@@ -0,0 +1,11 @@
...
+	exports com.fasterxml.jackson.core.filter to com.fasterxml.jackson.databind;
```
cowtowncoder: Filtering expected to be used by all kinds of things, including code outside Jackson modules (custom `TokenFilter` implementation f.ex), so needs to be exported to all.

cowtowncoder:
Ok. So. I think that realistically there's just one package -- `com.fasterxml.jackson.core.json.async` -- that can be hidden from everything else. Everything else needs at least some exemptions, until some refactoring (for 3.0 can and should do that). I don't think there's much point in fine-grained exports, either, since none of remaining cases is really limited to just one package.

There'll have to be bigger discussion (on mailing list) on question of exactly how to change package names for 3.0, but I don't think choices for 2.x will have that much effect given that it will be major (and, mostly, incompatible) version change.

src/moditect/module-info.java
GedMarc: hmm, it also seems to have my personal libraries listed there xD 

Going top down - 
Question 1)
- For internal API's that the library references, in this case, parent/bom/core/databind/annotations etc, these exports list should be specific to the internal references. that jwebmp.com one should definitely not be there. In terms of allowing modules to expose, you would normally never "open" a module, unless absolutely necessary, and definitely not expose all packages (same effect). Yes it seems refactoring would need to be done into a "impl" package of sorts that would be exposed. 
The idea is one exported package per module (I haven't managed to get this right yet anywhere). This is to minimize a) open api classes, and b) restrict public class viewing to only the internal API's that need it, and c) enforce proper object orientation encapsulation, package level.

Question 2) 
The java compiler can build previous class versions, yes the compiler and maven build will run in JDK 8 (hence why I said that Travis build will break on JDK 7), and you set the compiler to read sources and generate targets at a 1.6 JDK level. The pull request does this already :)

Agreed a list of packages to be exported would need to be compiled, but in my experience, it's better to work off a negative list, rather than a positive one. Just now you have a random someone using a class in your utils package which is a massive no-no....

OK, hope I answered most of your JPMS questions! I don't think the packages need to be redefined at all, or any refactoring has to take place. In terms of opening minor packages for known implementations, this can of course be done, when the packages get large with a lot of classes, serious consideration needs to take place with regards to exporting them. Don't break the module road map in the new JVM ;)

src/moditect/module-info.java
GedMarc: Do you mean these are data formats used by internal libraries, or by 3rd parties library implementations?
Easy enough to export 👍

src/moditect/module-info.java
GedMarc: I think doing a wide export (but not an opens all) will work, and then in 3 go "strict"? 
The big things here are just that the module is named so it works with jlink and transitive requires clauses, not really how it is exported (I guess), I tried to get as close to good with what I know, but thinking about, and how it currently is looking, a loose jpms module for 2.x compat rather than a strict may be the way to go

src/moditect/module-info.java
cowtowncoder: Ok. I think I was confused by removal of Java 6 source/target, as I did not see the replacement.
Build on Java 8 is fine, given that I don't think I can do successful Maven Central deployment with pre-Java-8 tools any more (for some ssl certificate thing or another).

As to opening up, I think we agree on idea. I was just looking at proposed changes and they seemed quite specific, but probably they were generated by a tool, for minimal exposure?

GedMarc: Ha! I'll take that as a compliment!!

I've gone through a few routes on this journey of pure module pathing, and I can honestly say it definitely is easier going from nothing up :)

On one of the harder ones (Guice), they attach the classes of asm and cglib in, and you need to expose that as well, but you can't during a regular build, so there are funnies like these that are resolvable, but it feels like we're all starting a new journey xD

When looking at the faster xml module pathing, I came up with the following - 
The really kak one is the java.desktop, and I think perhaps that a datatypes module for sql to keep it perfectly clean is probably a good idea, and the one class that is using it (i can't remember which it is now but removing it will highlight which one)

image

The important thing here is that no paths are duplicated, except for the cores java.logging and java.xml which is perfectly fine 👍 
This is bit different to the current PR (I did a bit of work on it once everything was named) - I'll try PR this all in tonight.

When looking at the "service-modules" the structure stays in tact. This is the benefit of named modules and i've confirmed jlink build compatible.

image
image

GedMarc: hmm see still a few more duplicate dependents xD

GedMarc:
@cowtowncoder Can you please turn off the JDK 7 build? 
As mentioned this would break :)

You can see the built files across the board with the correct class file versions.

Thanks!

GedMarc:
After all the changes I've seen that I've made (for all the data-formats and data types, I think it's better to release with 2.10, and to start again with the pom's updated a bit,

A lot of the enforcer constraints, especially with maven plugins, cause a bit of hassle when trying to update the build.
I think, push till 2.10, then kill the PR's, and I'll start again?

src/moditect/module-info.java
cowtowncoder: (for posterity)

Async is needed by any format implementation that wants to support async api, so needs to be open.
`sym` may not be as necessary, but most fasterxml impls use it and seem valuable for others too.

Issue #470, opened 2018-07-14
Replace toArray(new T[size]) with toArray(new T[0])


hc-codersatlas:
Might be able to be applied to other Jackson projects too.

https://shipilev.net/blog/2016/arrays-wisdom-ancients/#_conclusion (also contains benchmarks).

Could provide some simple performance improvements to Jackson without much work.

cowtowncoder:
I think I saw this article, and findings were interesting. But I must admit I am not convinced it would result in measurable perf improvements since such resizing is quite rare (most storage uses segment-based storage). So I think I'd consider specific changes along with test (jmh benchmark f.ex), if anyone has time and interest.

hc-codersatlas:
Found some time so will create the issues / PR across Jackson. Please review when you can. Thanks for considering!

PR #512, opened 2019-01-23
Optimize toArray (#470)

hc-codersatlas:
replace toArray(new T[size]) with toArray(new T[0]) for better performance


cowtowncoder:
Sure, why not.

PR #513, opened 2019-01-25
 minor startup / perf fixes

hc-codersatlas:
- Don't copy an array section that will be filled later and other minor changes.
- Better sizing for collections. Java 8 defaults to empty collections on creation, which may not be desired to start with.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/util/JsonParserSequence.java
+++ b/src/main/java/com/fasterxml/jackson/core/util/JsonParserSequence.java
@@ -77,7 +77,7 @@ public class JsonParserSequence extends JsonParserDelegate
             return new JsonParserSequence(checkForExistingToken,
                     new JsonParser[] { first, second });
         }
-        ArrayList<JsonParser> p = new ArrayList<JsonParser>();
+        ArrayList<JsonParser> p = new ArrayList<JsonParser>(10);
         if (first instanceof JsonParserSequence) {
             ((JsonParserSequence) first).addFlattenedActiveParsers(p);
         } else {
--- a/src/main/java/com/fasterxml/jackson/core/Base64Variants.java
+++ b/src/main/java/com/fasterxml/jackson/core/Base64Variants.java
@@ -29,10 +29,7 @@ public final class Base64Variants
      * it is <b>not</b> the default for Jackson: no-linefeeds alternative
      * is because of JSON requirement of escaping all linefeeds.
      */
-    public final static Base64Variant MIME;
-    static {
-        MIME = new Base64Variant("MIME", STD_BASE64_ALPHABET, true, '=', 76);
-    }
+    public final static Base64Variant MIME = new Base64Variant("MIME", STD_BASE64_ALPHABET, true, '=', 76);
 
     /**
      * Slightly non-standard modification of {@link #MIME} which does not
@@ -40,10 +37,7 @@ public final class Base64Variants
      * wouldn't work well (possibly in attributes), or for minor space savings
      * (save 1 linefeed per 76 data chars, ie. ~1.4% savings).
      */
-    public final static Base64Variant MIME_NO_LINEFEEDS;
-    static {
-        MIME_NO_LINEFEEDS = new Base64Variant(MIME, "MIME-NO-LINEFEEDS", Integer.MAX_VALUE);
-    }
+    public final static Base64Variant MIME_NO_LINEFEEDS = new Base64Variant(MIME, "MIME-NO-LINEFEEDS", Integer.MAX_VALUE);
 
     /**
      * This variant is the one that predates {@link #MIME}: it is otherwise
--- a/src/main/java/com/fasterxml/jackson/core/io/CharTypes.java
+++ b/src/main/java/com/fasterxml/jackson/core/io/CharTypes.java
@@ -74,7 +74,7 @@ public final class CharTypes
     static {
         final int[] table = new int[256];
         // Default is "not a name char", mark ones that are
-        Arrays.fill(table, 0, 33, -1);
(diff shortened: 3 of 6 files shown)


src/main/java/com/fasterxml/jackson/core/util/JsonParserSequence.java
```diff
@@ -77,7 +77,7 @@ public static JsonParserSequence createFlattened(boolean checkForExistingToken,
             return new JsonParserSequence(checkForExistingToken,
                     new JsonParser[] { first, second });
         }
-        ArrayList<JsonParser> p = new ArrayList<JsonParser>();
+        ArrayList<JsonParser> p = new ArrayList<JsonParser>(10);
```
GedMarc: Default to size 10? :) Any particular reason to this number? :)

hc-codersatlas: It's the normal JDK default before the "start from size 0" change.

GedMarc: Um ok, so unfortunately not anymore, JDK 8 and up does not default to 10

```
An ArrayList has an internal array to store the list elements.
There is a difference between the two constructor calls in Java 7 and 8:
If you do new ArrayList<>(0) the ArrayList creates a new Object array of size 0.
If you do new ArrayList<>() the ArrayList uses a static empty Object array of size 0 and switches to an own private array once you add items to the list.

The Javadoc of the default ArrayList constructor seems to contradict this.
```

Can we set this back to 0 please, 2.10 targets JDK 8

hc-codersatlas: There would be no point in setting it to 0 in this instance since there's an "add" command shortly after. So might as well construct it to begin with. The JDK does these optimizations for cases where it "might not" get used.

GedMarc: Handing over to @cowtowncoder

I'm trying to avoid the initial pitfall, correct initializing with 0 will skip the static performance boost, 
Reason it was put in this way is because instantiating at 10 and popping in 40 items results in a huge performance decrease before it stabilizes out again.
I view the push to instantiate with 10 as a regression for JDK 8

image

image

image

hc-codersatlas: I'm not sure what you're referring to, sorry. Starting at 0 just means it'll slowly grow to 10 anyhow. You can check the latest source code if needed. The only difference is 1 starts at 0 `int newCapacity = oldCapacity + (oldCapacity >> 1);` OR starts at 10 and grows the same. If you're going to grow to 40 as per your example - they'll both cross paths. The only optimization 0 can be better than 10 is if it's for sure < 10.

cowtowncoder:
Hmmh. Ok. So I am not a big fan of optimizing edge cases that do not show up in profilers (although sometimes eagerly over-optimize myself :) ), and ideally improvements would be, conversely, to something where improvement can be shown.
I am guessing maybe these change suggestions came from a tool that just scans code for suspected performance issues? (that would explain why there's one change to test code which really is a red herring).
But most of these seem ok, esp `compareTo()` which is bit more readable.

I'll have to read more about JDK 8 changes as my understanding of how "start with empty" works is bit different than what described here (I thought it mostly avoids initial allocation, benefiting empty case, but then starts with reasonable size).

@hc-codersatlas One thing before merging -- if I haven't yet asked for CLA, we would need one for contributions (just one for any and all for jackson projects):

A simple way is to print, fill & sign, scan, email to `info` at fasterxml dot com.

Looking forward to merging the changes!

cowtowncoder:
@hc-codersatlas @GedMarc Not directly related to this, but something I realized -- there is a quite a bit of interested in performance improvements, and you are couple of most active contributors here. In addition to patch suggestions, I wonder if it'd make sense to figure out some more pro-active way of collaborating? Like discussion group, separate issue tracker/wiki (could maybe use existing `jackson-benchmarks` for those) or something? I haven't shared many of ideas or experiments I have done, and issues (or mailing list) are not always best ways to discuss things in depth and in detail. But I feel that there's so much potential here, especially since individuals like me may end down the wrong rat hole sometimes (there are many attempts that have been fruitless, either for good, or due to benchmark(s) used not showing improvements due to being too narrow).

So anyway. I'd like collaborate more, extend this, and would be interested in ideas you might have?

cowtowncoder:
Hmmh. Not sure what I did but looks like I managed to get discussions... removed? Hmmh. Not a big fan of github ui here.

One note on default size changes: I do NOT think that use of defaults with later JDK 8 changes speed at growth. As per, say:

https://stackoverflow.com/questions/34250207/in-java-8-why-is-the-default-capacity-of-arraylist-now-zero

it is true that initial allocation is deferred by assigning backup storage to `Object[0]`.
But upon first add, calculation of initial size seems to be to default.
If so, I do not see meaningful benefit in forcing default size even if resulting collection is not empty. Using different default in case of parser sequence would be fine (like, say, 4), but all of these changes seem pretty meaningless for actual overall performance.

On the other hand, I like changes like "use `compareTo()` instead of written out code", so +1 for that, and looking forward to similar improvements.

hc-codersatlas:
@cowtowncoder will get over the CLA shortly. Should I update the PR and remove the resizing? Happy to join any disucssion groups. Not sure why no 1 is more active here. Lots are interested in performance and use Jackson extensively (as in bigger frameworks / companies). It's strange, but it's definitely a thing. This patch came from a lot of frameworks all mentioning the "issue" of Jackson startup times. Just took a quick look at minor fixes. "Accidentally" changed a few things whilst at it - bad habit.

PR #518, opened 2019-02-16
Improve performance of parsing long from char array

skjolber:
Somewhat improved long parsing performance. See benchmarks at https://github.com/skjolber/various-jmh-benchmarks

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/io/NumberInput.java
+++ b/src/main/java/com/fasterxml/jackson/core/io/NumberInput.java
@@ -27,30 +27,25 @@ public final class NumberInput
      */
     public static int parseInt(char[] ch, int off, int len)
     {
-        int num = ch[off] - '0';
-
-        if (len > 4) {
-            num = (num * 10) + (ch[++off] - '0');
-            num = (num * 10) + (ch[++off] - '0');
-            num = (num * 10) + (ch[++off] - '0');
-            num = (num * 10) + (ch[++off] - '0');
-            len -= 4;
-            if (len > 4) {
-                num = (num * 10) + (ch[++off] - '0');
-                num = (num * 10) + (ch[++off] - '0');
-                num = (num * 10) + (ch[++off] - '0');
-                num = (num * 10) + (ch[++off] - '0');
-                return num;
-            }
-        }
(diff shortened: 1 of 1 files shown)


cowtowncoder:
Hi @skjolber! Would it be possible to cut'n paste some jmh output, along with JDK/JVM settings used for running differing versions? If I get time, I can try to see I can see differences on my end.
I don't doubt this could improve things, but I have found that improvements are sometimes JDK version specific and reproduction may vary a lot across versions.

I also think this only applies to input sources that are based on chars (Reader) or `String`s, which has not been as much a focus (since that's not as optimal as starting with byte sequence), so it is quite possible this approach could be faster.

Thank you for submitting this!

skjolber:
I guess there is some work to be done on running it on different JVMs, especially OpenJDK 11 - which seems not to like my current JMH gradle project. See the linked projects build.gradle file for JMH details. 

Benchmarks run both on the modified method and on the 2x calling parseLong method. 

OpenJDK 8 build 201 on Linux 4.20: 
https://skjolber.github.io/various-jmh-benchmarks/openjdk8_201/

cowtowncoder:
Ok thank you, that looks legit. And makes sense there's more benefit to longer numbers (if I read results right).

So, just one more thing -- if we haven't yet asked and received a CLA, we'd need that:

and usually it's easiest to print, fill & sign, scan, email to `info` at fasterxml dot com.
Only needs to be done once for all contributions.

With that, I'll be happy to merg this PR.

Thanks againg!

skjolber:
Yuck, paper, is there not a github integration for this CLA stuff? Luckily I was able to use my reMarkable.  :)

cowtowncoder:
Alas, I don't know an easy way nor have anyone to help get that going. I would definitely like an improvement.

FWTW I think it is possible to process pdf to work around need for actual printing, I've received a few cases where this seemed to be done.

skjolber:
@cowtowncoder I used a https://remarkable.com, did you get any email?

cowtowncoder:
@skjolber yeah sorry, I did get it on wednesday, just hadn't time to file it yet (plus couldn't reply).

PR #519, opened 2019-02-26
Prefer ArrayList over LinkedList

BELUGABEHR:
> ArrayList with ArrayDeque are preferable in much more use-cases than LinkedList. If you're not sure — just start with ArrayList.

https://stackoverflow.com/questions/322715/when-to-use-linkedlist-over-arraylist-in-java

Also, there is currently a check for 'isEmpty()' before a call to 'clear'.  This is superfluous because the 'clear' method will perform a no-op if there are no entries to clear... no need to double-check.


cowtowncoder:
I think I'll pass the change since `LinkedList` was chosen on purpose (it's likely lists are short, and access is never by index). But I will remove the emptiness check as that does seem unnecessary.

PR #521, opened 2019-02-26
Remove superfluous toString convenience methods

BELUGABEHR:


cowtowncoder:
Looks like actual use in 2.x was from `IntNode`, `LongNode`, although arguably with little benefit. So makes sense to remove. Thanks!

cowtowncoder:
Annnnd this broke `jackson-databind` master. :-( :-( :-(

I should know better than blindly merge method removal.

BELUGABEHR:
Ugh. Sorry!

Well it passed all the unit tests and this is the 3.0 release so we should
be able to at least deprecate.

cowtowncoder:
@BELUGABEHR yeah no huge problem, I went ahead and replaced uses; no point in optimizing those cases.

Would also be neat to figure out how to make build chains for deps work on Travis or something.

PR #522, opened 2019-02-26
Remove double-check of v value

BELUGABEHR:
Found a place where it checks 'is this value less than 1000'... 'is this value less than 10'.  Bring out the second check so that anything less than 10 does not have to pass through two checks.


cowtowncoder:
Hmmh. Then yet again, for all values from 10 to 999,999, this adds one more check.
I doubt change results in measurable gain (or loss) but fine.

PR #523, opened 2019-02-27
Use simple comparison with equals

BELUGABEHR:
Because... why not?  I think it's easier to read anyway.

https://stackoverflow.com/questions/22466117/is-comparing-to-zero-faster-than-comparing-to-any-other-number


cowtowncoder:
Um, to be honest I would prefer changes that have reason for them, not simply something that has nothing against them.

BELUGABEHR:
Sure, not worries.  Thanks.

PR #526, opened 2019-03-04
Fix typo in javadoc of readBinaryValue

gaganis:
Before this commit the javadoc was describing readBinaryValue as an alternative to getBigInteger which looks like a typo. From what the method does and the rest of the javadoc text it should be referring to getBinaryValue.


cowtowncoder:
yes, thank you!

PR #527, opened 2019-03-05
Add simple module-info for JDK9+, using Moditect

cowtowncoder:
As per title. For 2.x will use `moduleInfoFile` which allows build using JDK 8, but is quite inconvenient as it can not do wildcard exports; will upgrade in 3.0, and probably for components other than `jackson-core` / `jackson-annotations`.

### DIFF
--- a/src/moditect/module-info.java
+++ b/src/moditect/module-info.java
@@ -3,6 +3,7 @@ module com.fasterxml.jackson.core {
     //    So, for 2.x core need to make sure we manually include everything.
     //    Worse, there is only syntactic validation, not contents, so we can both miss
     //    AND add bogus packages.
+    //    However: at least syntax is verified; and this works with JKD8
     exports com.fasterxml.jackson.core;
     exports com.fasterxml.jackson.core.async;
     exports com.fasterxml.jackson.core.base;
--- a/src/moditect/module-info.java
+++ b/src/moditect/module-info.java
@@ -1,0 +1,18 @@
+module com.fasterxml.jackson.core {
+    // 04-Mar-2019, tatu: Ugh. Can not use wildcards, stupid ass JDK 9+ module system...
+    //    So, for 2.x core need to make sure we manually include everything.
+    //    Worse, there is only syntactic validation, not contents, so we can both miss
+    //    AND add bogus packages.
+    exports com.fasterxml.jackson.core;
+    exports com.fasterxml.jackson.core.async;
+    exports com.fasterxml.jackson.core.base;
+    exports com.fasterxml.jackson.core.exc;
+    exports com.fasterxml.jackson.core.filter;
+    exports com.fasterxml.jackson.core.format;
+    exports com.fasterxml.jackson.core.io;
+    exports com.fasterxml.jackson.core.json;
+    exports com.fasterxml.jackson.core.json.async;
+    exports com.fasterxml.jackson.core.sym;
+    exports com.fasterxml.jackson.core.type;
+    exports com.fasterxml.jackson.core.util;
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: java
 # 08-Dec-2018, tatu: While it should be possible to run core streaming on Java 6,
 #   build won't work with anything below Java 8 now
   - openjdk8
+  - oraclejdk11
 
 # Below this line is configuration for deploying to the Sonatype OSS repo
 # http://blog.xeiam.com/2013/05/configure-travis-ci-to-deploy-snapshots.html
--- a/pom.xml
+++ b/pom.xml
@@ -23,15 +23,16 @@
   </scm>
 
   <properties>
-    <!-- 16-Sep-2016, tatu: Retain Java6/JDK1.6 compatibility for streaming for Jackson 2.x -->
+    <!-- 04-Mar-2019, tatu: Retain Java6/JDK1.6 compatibility for annotations for Jackson 2.x,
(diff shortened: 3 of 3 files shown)


cowtowncoder:
@GedMarc similar to annotations one. I think I need to consider JDK 11 for build for other components, to be able to use wildcards for export definitions (and probably for imports too).

GedMarc:
Yeaaahhh, no Wildcards ;) I tried my heart out, but it isn't osgi xD

We found out on this thread (https://github.com/moditect/moditect/issues/90#issuecomment-454790227) that it is actually purely cosmetic from the compiler, the output class knows nothing of it - the output class actually has everything expanded.

I put my module info's on the thread for annotations (https://github.com/FasterXML/jackson-annotations/pull/151)

They seem to be doing the trick quite nicely, and JLink is building like a star

cowtowncoder:
Ok thanks -- will leave output at `base` then, as that did seem to work. Plus planning to do 2.10.0.rc1 (or is it pr1, either way) to give others chance to see if there are problems.
Thank you for adding other module infos; as I said I'll have to think about JDK used to build; maybe only keeping annotations/streaming/databind at JDK 8 (since those probably are most important to avoid adding accidental deps to post-jdk8 types).
Wildcards are a bummer but I can understand why this is, if code is relying on JDK 9+ functionality for heavy lifting (and not reinventing the wheel) of expanding.

cowtowncoder:
Alas, Moditect seems quite half-baked tool, giving obscure errors on about any input :-( :-( :-(
In particular I just can't figure out how to generate package-info without either

```
[ERROR] Failed to execute goal org.moditect:moditect-maven-plugin:1.0.0.Beta2:generate-module-info (default-cli) on project aalto-xml: Execution default-cli of goal org.moditect:moditect-maven-plugin:1.0.0.Beta2:generate-module-info failed. NullPointerException -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/PluginExecutionException
```

or

```
[ERROR] Failed to execute goal org.moditect:moditect-maven-plugin:1.0.0.Beta2:generate-module-info (generate-module-info) on project aalto-xml: Execution generate-module-info of goal org.moditect:moditect-maven-plugin:1.0.0.Beta2:generate-module-info failed: Encountered unexpected token: "-" "-"
[ERROR] at line 1, column 20.
[ERROR] 
[ERROR] Was expecting:
[ERROR] 
[ERROR] <EOF>
[ERROR] 
[ERROR] Problem stacktrace :
[ERROR] com.github.javaparser.GeneratedJavaParser.generateParseException(GeneratedJavaParser.java:10419)
[ERROR] com.github.javaparser.GeneratedJavaParser.jj_consume_token(GeneratedJavaParser.java:10278)
[ERROR] com.github.javaparser.GeneratedJavaParser.NameParseStart(GeneratedJavaParser.java:5785)
[ERROR] com.github.javaparser.JavaParser.parse(JavaParser.java:136)
[ERROR] com.github.javaparser.JavaParser.simplifiedParse(JavaParser.java:342)
[ERROR] com.github.javaparser.JavaParser.parseName(JavaParser.java:503)
```

GedMarc:
Naw Moditect is everywhere and though its beta it is very stable, classgraph guice, primefaces, etc...

Null Pointer is usually a very wrong module file, P.S> rather use an actual module-info.java file, with an attach-sources for JDK11 and up

This lets you build the module file in JDK 11, and simply switch a profile to test the forwards compatibility,

P.P.S.
No import statements, No Wildcards. Always full referencing (Pity you named the SPI "Module" though, perhaps think about giving it a new name in v3)

cowtowncoder:
Including actual `module-info.java` does work; and although there isn't meaningful semantic validation at least syntax is validated by plug-in. So we'll go with this.

cowtowncoder:
Yeah, `Module` was chosen well before Java 9, and is quite baked-in by now. Major compatibility issue to change that, although as usual could change in 3.0. I guess most users do not interact with that type, so base interface could become `JacksonModule` I guess.

As to NPE and other exceptions these are with jackson-core, aalto, so nothing broken about code, only set up of module info and/or plugin configuration. I'm sure there's some reason for failure, but error reporting really is rather bad and I hope maintainers can improve upon it, esp. same useless exceptions were thrown with JDK 11, not just 8.

But be that as it may not much I can do about that.

src/moditect/module-info.java
```diff
@@ -0,0 +1,19 @@
+module com.fasterxml.jackson.core {
```
gunnarmorling: I feel your pain. Note ModiTect would allow you to use wildcards (com.fasterxml.jackson.core.*) if you were to use it on JDK 9 or later. It will then generate the actual descriptor exporting all packages matching the wildcard.

gunnarmorling:
About the NPE, have you figured what is triggering it? If you file an issue ideally containing a small reproducer project, I can take a look.

src/moditect/module-info.java
cowtowncoder: Ok that is good to know going forward. While Jackson 2.x will probably not go beyond JDK 8, 3.x will at some point so that's good info.

cowtowncoder:
No, I didn't. I haven't seen it with other projects, for what that is worth. If I bump into it again, I will try to get reproducible test case & file an issue. Thanks!

Issue #533, opened 2019-04-23
UTF-8 BOM not accounted for in JsonLocation.getByteOffset()


fabienrenaud:
**Version:** Jackson 2.9.8

`parser.getCurrentLocation().getByteOffset()` returns the wrong byte offset for the underlying byte array when the payload start with a [BOM](https://en.wikipedia.org/wiki/Byte_order_mark).
The json parser processes well such json payloads but the `JsonLocation` it returns ignores the offset introduced by the BOM.

Full standalone repro:
```java
package test;

import com.fasterxml.jackson.core.JsonFactory;
import com.fasterxml.jackson.core.JsonParser;
import com.fasterxml.jackson.core.JsonToken;

import java.io.IOException;
import java.util.Arrays;

public class Test {

    private static final JsonFactory JSON_FACTORY;

    static {
        JSON_FACTORY = new JsonFactory();
        JSON_FACTORY.setCharacterEscapes(null);
        JSON_FACTORY.disable(JsonFactory.Feature.INTERN_FIELD_NAMES);
    }

    static byte[] extract(byte[] bytes, JsonParser parser, JsonToken token) throws IOException {
        switch (token) {
            case START_OBJECT:
                int startIndex = (int) parser.getCurrentLocation().getByteOffset() - 1;
                parser.skipChildren();
                int endIndex = (int) parser.getCurrentLocation().getByteOffset();
                return Arrays.copyOfRange(bytes, startIndex, endIndex);
        }
        throw new RuntimeException();
    }

    static byte[] parseAndExtract(byte[] bytes) throws IOException {
        JsonParser parser = JSON_FACTORY.createParser(bytes);
        System.out.println("parser type: " + parser.getClass().getCanonicalName());

        parser.nextToken(); // skip start object
        switch (parser.nextToken()) {
            case FIELD_NAME:
                if ("payload".equalsIgnoreCase(parser.getCurrentName())) {
                    return extract(bytes, parser, parser.nextToken());
                }
                break;
        }
        throw new RuntimeException();
    }

    public static void main(String[] args) throws Exception {
        String json = "{\"payload\":{\"name\":\"foo\"}}";
        byte[] bytes = json.getBytes();

        System.out.println("UTF-8 no BOM:");
        byte[] result = parseAndExtract(bytes);
        System.out.println(new String(result));
        System.out.println();

        System.out.println("UTF-8 no BOM, with leading characters:");
[...]

PR #534, opened 2019-04-24
Account for bytes processed by encoding detection

fabienrenaud:
UTF8StreamJsonParser tracks read pointer (offset) and bytes processed
separately and uses those to generate JsonLocation. When the byte
payload starts with a UTF BOM, ByteSourceJsonBootstrapper processes a
few bytes ahead of the parser, moves/increases the offset and passes the
newly computed offset to the parser without telling it some bytes have
been pre-processed.
With this change, the number of bytes pre-processed for encoding
detection is passed to the parser. JsonLocation instances returned by
the parser now point to the correct byte offset when payload has a BOM.

Issue: https://github.com/FasterXML/jackson-core/issues/533

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/json/ByteSourceJsonBootstrapper.java
+++ b/src/main/java/com/fasterxml/jackson/core/json/ByteSourceJsonBootstrapper.java
@@ -242,7 +242,9 @@ public final class ByteSourceJsonBootstrapper
             ByteQuadsCanonicalizer rootByteSymbols, CharsToNameCanonicalizer rootCharSymbols,
             int factoryFeatures) throws IOException
     {
+        int prevInputPtr = _inputPtr;
         JsonEncoding enc = detectEncoding();
+        int bytesProcessed = _inputPtr - prevInputPtr;
 
         if (enc == JsonEncoding.UTF8) {
             /* and without canonicalization, byte-based approach is not performant; just use std UTF-8 reader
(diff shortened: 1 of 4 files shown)


cowtowncoder:
Ah! Sorry I somehow missed this part. Since we already got your CLA I'll just need to quickly skim over this and should be able to merge. The only blocked is that there's a CVE that I need to handle first, but I should be able to get to this next!

src/main/java/com/fasterxml/jackson/core/json/ByteSourceJsonBootstrapper.java
```diff
@@ -252,7 +254,7 @@ public JsonParser constructParser(ObjectReadContext readCtxt,
                 ByteQuadsCanonicalizer can = rootByteSymbols.makeChild(factoryFeatures);
                 return new UTF8StreamJsonParser(readCtxt, _context,
                         streamReadFeatures, formatReadFeatures, _in, can,
-                        _inputBuffer, _inputPtr, _inputEnd, _bufferRecyclable);
+                        _inputBuffer, _inputPtr, _inputEnd, bytesProcessed, _bufferRecyclable);
```
cowtowncoder: Hmmh. This is bit problematic wrt backwards-compatibility. Need to think about it, needs to go in 2.10.0 not 2.9.x.

fabienrenaud: Problematic because `UTF8StreamJsonParser` no longer has a constructor with 9 arguments (it has 10 now)? I could add a `UTF8StreamJsonParser` public constructor that defaults `bytesProcessed` to 0.

cowtowncoder: Right, only for that reason. So although technically one is supposed to always make sure jackson-core / jackson-databind minor versions align, we try to make sure that "adjacent" versions still have implementations of deprecated methods for at least one version.

So in this case keeping deprecated "old" version around would allow 2.9/2.10 combination to still work (2.9 databind, 2.10 core).

fabienrenaud: Going to add back 10 arg constructors.
Is it really worth deprecating that 10 args constructor in favor of the 11 args one?

fabienrenaud: 10 arg constructor is back!

PR #538, opened 2019-06-06
Extend JsonStringEncoder to support CharSequence

CodingFabian:
Currently JsonStringEncoder requires a String, but it actually works completely on the API of CharSequence internally.
Widening the parameter to CharSequence for example allows passing in a StringBuilder, which can avoid an intermediate toString() copy

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/io/JsonStringEncoder.java
+++ b/src/main/java/com/fasterxml/jackson/core/io/JsonStringEncoder.java
@@ -81,7 +81,7 @@ public final class JsonStringEncoder
      * Method that will quote text contents using JSON standard quoting,
      * and return results as a character array
      */
-    public char[] quoteAsString(String input)
+    public char[] quoteAsString(CharSequence input)
     {
         TextBuffer textBuffer = _text;
         if (textBuffer == null) {
@@ -178,7 +178,7 @@ public final class JsonStringEncoder
      * results as UTF-8, and return result as a byte array.
      */
     @SuppressWarnings("resource")
-    public byte[] quoteAsUTF8(String text)
+    public byte[] quoteAsUTF8(CharSequence text)
     {
         ByteArrayBuilder bb = _bytes;
         if (bb == null) {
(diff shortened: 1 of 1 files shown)


CodingFabian:
if possible I would like to see this in 2.9 (for very selfish reasons)

cowtowncoder:
Ok so I'm not sure I should do that for 2.9 for various reasons (adding functionality in a patch is not suggested, but as importantly, 2.9.10 is probably months away so won't necessarily be available before 2.10.0).

Actually, even worse, this is binary incompatible change so I am not sure it should go in 2.x at all.
Any code that uses methods listed would break unless recompiled against new version; this is ok for direct use but not for transitive dependencies.

Finally there's the question of performance. JIT is less likely to be able to fully optimize code via interface. I don't know how big effect is nor remember exactly where this class is used. But it's bit of a concern.

CodingFabian:
oh yeah, you are right about binary compatibility. I thought of that in my case, and jackson itself is fine, but I forgot other users that might have taken advantage of this class. let me fix this, and then I appreciate you thinking about a possible inclusion.
I doubt that this has an effect on JIT since it will see that you only use it with String

CodingFabian:
Made it binary compatible. thanks for pointing that out.

cowtowncoder:
Added overload for `quoteAsString()`, will handle `encodeAsUTF8()` and `quoteAsUTF8()` later on.

CodingFabian:
@cowtowncoder still looking for `encodeAsUTF8()` - any way I can help making it happen?

cowtowncoder:
Ah. Dropped the ball here it looks like. I think another PR that just adds overload similar to `quoteAsString()`, based on 2.11, would be great? I could merge that, and for 3.0 (`master`) just drop the String-taking one (when backwards compatibility does not matter).

CodingFabian:
No issue - 2.11 is fine. This is more of community benefit. I forked the encoder into our codebase, so the benefit for me would be reduced maintenance.

cowtowncoder:
Actually, let me just add it myself -- simple enough task.

cowtowncoder:
Done. Apologies for late notice -- should have realized it's trivial task. Thank you for reminder!

PR #562, opened 2019-09-17
Add failing test: PointerFromContextTest for NonBlockingJsonParser

htmldoug:
`pathAsPointer()` doesn't return expected `JsonPointer` values in the `NonBlockingJsonParser`. It seems to track descent object keys just fine, but fails to track the array indices. They either appear as `/0` for all indices or come back empty.

I discovered this against 2.9.9, although it seems to be reproducible in master as well.

I'm currently using the `NonBlockingJsonParser` with akka-stream and uPickle. Would be nice to lean on the parser for correct `JsonPointer`s to wrap around downstream visitor exceptions (rather than tracking the path myself in user space).

I haven't started looking into what it'd take to fix this since I'm unfamiliar with the parser internals, although let me know if you'd like for me to take a crack at it.

### DIFF
--- a/src/test/java/com/fasterxml/jackson/core/PointerFromContextTest.java
+++ b/src/test/java/com/fasterxml/jackson/core/PointerFromContextTest.java
@@ -1,8 +1,10 @@
 package com.fasterxml.jackson.core;
 
 import java.io.StringWriter;
+import java.nio.charset.StandardCharsets;
 
 import com.fasterxml.jackson.core.json.JsonFactory;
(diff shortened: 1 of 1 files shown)


cowtowncoder:
Thanks! Makes sense.

cowtowncoder:
I merged the test manually as I wanted to put it under async tests (although initially goes under `failing`). I hope to have a look at this soon, to get fix in 2.10.0, but we'll see.

Thank you for reporting the issue!

PR #591, opened 2020-01-09
Fix NPE in DataFormatMatcher#getMatchedFormatName when no match exists

sleberknight:
* Change DataFormatMatcher#getMatchedFormatName so it does what
  the Javadoc says, because otherwise it throws a NullPointerException
  when there is no match (and the _match field is null)
* Added tests for when there is a match, and when there isn't


cowtowncoder:
Yes, makes total change. Will merge as suggested.

Issue #504, opened 2018-12-13
Add a String Array write method in the Streaming API


feinstein:
In version 2.9.7 there are methods for writing `int[]`, `long[]` and `double[]` arrays. I think it would be a nice addition to add a method for writing `String[]` arrays as well.

cowtowncoder:
Forgot to ask: is this just for convenience or for some other reason? Not against addition, just want to double-check.

feinstein:
Just a convenience. If I recall it correctly (as there has been many years already) I saw the docs and asked myself why `String[]` wasn't included.

cowtowncoder:
Ok. Sounds reasonable. Challenge wrt APIs is regarding how much to include in cases like this; but often just a core set of `boolean`, `int`, `long`, `String` are included so I think adding of `String[]` seems reasonable.

cowtowncoder:
Will be included in 2.11.0.

PR #598, opened 2020-02-10
#504 Add String array write method to Streaming API

alevskyi:


src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
```diff
@@ -962,6 +962,30 @@ public void writeArray(double[] array, int offset, int length) throws IOExceptio
...
+     * @param offset Offset of the first element to write, within array
+     * @param length Number of elements in array to write, from `offset` to `offset + len - 1`
+     */
+    public void writeArray(String[] array, int offset, int length) throws IOException
+    {
+        if (array == null) {
+            throw new IllegalArgumentException("null array");
+        }
+        _verifyOffsets(array.length, offset, length);
+        writeStartArray(array, length);
```
cowtowncoder: Can now use `writeStartArray(array, length)` (added in 2.10). Not much practical difference at this point, but for sake of completeness (since "forValue" is sometimes for `List` values or such, at least for other formats) good to do.

cowtowncoder:
Thank you very much for this & tests @alevskyi ! Just one practical matter before I merge this in: if I have not yet asked for and received (apologies if I have; I did not have it) CLA, we would need that filed once (but only once). It can be found from:

(there is also alternate corporate CLA if this contribution is part of corporate job, but most contributions are by individuals)

and commonly you can just print it, fill & sign, scan/photo, email to `info` at fasterxml dot com.
Once I receive it I'll merge these PRs.

Again, thank you for contributing this improvement!

alevskyi:
@cowtowncoder Thank you for review. It's a pleasure, that I can make small contribution to such wonderful project. Unfortunately, I don't have access to printer at all. Sorry for inconvenience.

cowtowncoder:
@alevskyi Hmmh. That is unfortunate, but I am sure we can figure out some other way. I think digital copy would be acceptable as well: if it was possible to take CLA pdf, convert to image format that can be modified (png, jpeg), then add other textual information with a paint program except ideally scan signature (but I think it is fine, if necessary, to just write that same way as other information too).

Apologies for this complication but some of the users -- big corporations in particular -- are quite serious about projects verifying all contributions. Asking for CLAs has worked out well for that reason, and that is why we do it (I personally would just trust committers but then I am not lawyer and do not have to work on behalf of clients).

Do you think this could work?

Issue #568, opened 2019-10-08
jackson-core-2.9.9 encodeAsUTF8 ERROR: Internal error: total len assumed to be of version 


Stephenkk204:
hi,
now we used jackson-core version 2.9.9. When use JsonStringEncoder.encodeAsUTF8(req) to parse the input request,  Sometimes we get the following error:
java.lang.RuntimeException: Internal error: total len assumed to be 7300, copied 4528 bytes
        at com.fasterxml.jackson.core.util.ByteArrayBuilder.toByteArray(ByteArrayBuilder.java:157)
        at com.fasterxml.jackson.core.util.ByteArrayBuilder.completeAndCoalesce(ByteArrayBuilder.java:202)
        at com.fasterxml.jackson.core.io.JsonStringEncoder.encodeAsUTF8(JsonStringEncoder.java:363)

please help to check this . Thanks.

cowtowncoder:
I would need a reproduction of some kind, such as unit test. Otherwise all I can suggest is to try out `2.10.0` that was released: there are some concurrency improvements that could help.

Stephenkk204:
I don't know how to reproduce it.  I will try out 2.10.0

alevskyi:
It seems that this exception may occur when write or reset method called on `com.fasterxml.jackson.core.util.ByteArrayBuilder` while another thread executing `toByteArray` method. I tried many times, and was able to reproduce it only on very long byte array. Investigated on jackson-core-2.11

cowtowncoder:
@alevskyi Hmmh. If so, that'd be problem with usage as `ByteArrayBuilder` is not designed for multi-threaded access, ever. So while I can definitely believe multi-threaded access could be problematic, such usage should not be done. Neither parsers nor generators are meant to be called from more than one thread.

This does not necessarily mean there is no problem, but just that it'd be necessary to find where such sharing occurs.

alevskyi:
@cowtowncoder I did not find any usages of this class in multiple threads.

cowtowncoder:
@alevskyi Thanks. I think that this is related to something fixed in 2.10: up to 2.9, `JsonStringEncoder` instances were recycled, and they contained references to `ByteArrayBuilder`. With 2.10 I started untangling reuse (in `BufferRecyclers`), and just removed this particular reuse. So I am thinking that problem is likely something that could only occur in 2.9.

[...]

PR #599, opened 2020-02-10
#568 Add failing test

alevskyi:


cowtowncoder:
As per discussion on#568, I think problem has been resolved. I did check this test with 2.11 branch and test passes.

Issue #603, opened 2020-02-25
'JsonParser.getCurrentLocation()` byte/char offset update incorrectly for big payloads


fabienrenaud:
The following test passes with jackson-core-2.10.0 and fails jackson-2.10.1:
```
    public void testBigPayload() throws IOException {
        JsonLocation loc;
        JsonParser p;

        String doc = "{\"key\":\"" + generateRandomAlpha(50000) + "\"}";

        p = createParserUsingStream(JSON_F, doc, "UTF-8");

        assertToken(JsonToken.START_OBJECT, p.nextToken());
        loc = p.getTokenLocation();
        assertEquals(0, loc.getByteOffset());
        assertEquals(-1L, loc.getCharOffset());
        assertEquals(1, loc.getLineNr());
        assertEquals(1, loc.getColumnNr());
        loc = p.getCurrentLocation();
        assertEquals(1, loc.getByteOffset());
        assertEquals(-1L, loc.getCharOffset());
        assertEquals(1, loc.getLineNr());
        assertEquals(2, loc.getColumnNr());

        assertToken(JsonToken.FIELD_NAME, p.nextToken());
        loc = p.getTokenLocation();
        assertEquals(1, loc.getByteOffset());
        assertEquals(-1L, loc.getCharOffset());
        assertEquals(1, loc.getLineNr());
        assertEquals(2, loc.getColumnNr());
        loc = p.getCurrentLocation();
        assertEquals(8, loc.getByteOffset());
        assertEquals(-1L, loc.getCharOffset());
        assertEquals(1, loc.getLineNr());
        assertEquals(9, loc.getColumnNr());

        assertToken(JsonToken.VALUE_STRING, p.nextToken());
        loc = p.getTokenLocation();
        assertEquals(7, loc.getByteOffset());
        assertEquals(-1L, loc.getCharOffset());
        assertEquals(1, loc.getLineNr());
        assertEquals(8, loc.getColumnNr());
        loc = p.getCurrentLocation();
        assertEquals(8, loc.getByteOffset());
        assertEquals(-1L, loc.getCharOffset());
        assertEquals(1, loc.getLineNr());
        assertEquals(9, loc.getColumnNr());

        p.getTextCharacters();
        loc = p.getTokenLocation();
        assertEquals(7, loc.getByteOffset());
        assertEquals(-1L, loc.getCharOffset());
        assertEquals(1, loc.getLineNr());
        assertEquals(8, loc.getColumnNr());
[...]

PR #604, opened 2020-02-25
Fixed issue #603

fabienrenaud:
`_currInputProcessed`, `_currInputRowStart`, and `_nameStartOffset` must be updated before `_inputPtr`

See #603

### DIFF
--- a/src/test/java/com/fasterxml/jackson/core/read/LocationOffsetsTest.java
+++ b/src/test/java/com/fasterxml/jackson/core/read/LocationOffsetsTest.java
@@ -4,9 +4,7 @@ import com.fasterxml.jackson.core.*;
 import com.fasterxml.jackson.core.json.JsonFactory;
 
 import java.io.IOException;
-import java.util.Base64;
 import java.util.Random;
-import java.util.concurrent.ThreadLocalRandom;
 
 public class LocationOffsetsTest extends com.fasterxml.jackson.core.BaseTest
 {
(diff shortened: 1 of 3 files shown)


src/test/java/com/fasterxml/jackson/core/read/LocationOffsetsTest.java
```diff
@@ -3,6 +3,11 @@
 import com.fasterxml.jackson.core.*;
 import com.fasterxml.jackson.core.json.JsonFactory;
 
+import java.io.IOException;
+import java.util.Base64;
```
cowtowncoder: Can not use until 3.0 (part of JDK 8). Same for a few other things, will make necessary changes.

cowtowncoder:
Thank you for the fix(es)!

Due to minor JDK 8 dependencies here I think I'll merge changes manually, starting with 2.10 branch. But keeping essential logic.

fabienrenaud:
@cowtowncoder I can replace the JDK8 dependencies with this function if you'd like:
```
    private String generateRandomAlpha(int length)
    {
        StringBuilder sb = new StringBuilder(length);
        Random rnd = new Random(length);
        for (int i = 0; i < length; ++i) {
            // let's limit it not to include surrogate pairs:
            char ch = (char) ('A' + rnd.nextInt(26));
            sb.append(ch);
        }
        return sb.toString();
    }
```

src/test/java/com/fasterxml/jackson/core/read/LocationOffsetsTest.java
fabienrenaud: I just replaced it with a `generateRandomAlpha` method.

cowtowncoder:
Jackson has full Base64 functionality in itself ....

fabienrenaud:
I don't really need base64 for this test, so it's okay...

PR #605, opened 2020-03-05
Handle case when system property access is restricted

rhernandez35:
I own a system that limits access to system properties. Requiring access to this property to determine if `trackReusableBuffers` should be enabled prevents my users from using Jackson.


cowtowncoder:
@rhernandez35 Thank you for reporting the problem. Oddly enough this is the first time I hear about it... Will backport in 2.10 as well.

rhernandez35:
Thanks for the fast response!

PR #611, opened 2020-04-14
Optionally allow leading decimal in float tokens

jamesagnew:
JSON doesn't allow a leading decimal in a float (e.g. `{ "someval": .123 }` ) and Jackson rightly errors out when you try to parse this.

I have a need however (mostly caused by a previous Gson implementation now replaced with Jackson in [HAPI FHIR](https://hapifhir.io) silently accepting this syntax) to parse JSON containing this invalid format.

The attached pull request adds a read feature enabling the parsing of these leading decimal points.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/json/JsonReadFeature.java
+++ b/src/main/java/com/fasterxml/jackson/core/json/JsonReadFeature.java
@@ -103,7 +103,18 @@ public enum JsonReadFeature
      * this is a non-standard feature, and as such disabled by default.
      */
     ALLOW_LEADING_ZEROS_FOR_NUMBERS(false),
-    
+
+    /**
+     * Feature that determines whether parser will allow
+     * JSON decimal numbers to start with a deciaml point
+     * (like: .123). If enabled, no exception is thrown, and the number
+     * is parsed as though a leading 0 had been present.
+     *<p>
+     * Since JSON specification does not allow leading decimal,
+     * this is a non-standard feature, and as such disabled by default.
+     */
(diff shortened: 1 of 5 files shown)


cowtowncoder:
@jamesagnew I thought I added a comment here but looks like I did not. So, I think this makes sense: I'll have a look and may be able to merge this to be included in 2.11 (will cherry-pick from master commit, should be fine).
Thank you for submitting the patch.

But one thing I would need (if I haven't yet asked for it; I don't think I have) before merging the patch is CLA. It's found from

and is usually easiest to print, fill & sign, scan/photo, email to `info` at fasterxml dot com.
This is only needed once before the first contribution, one is fine for any and all submissions.

Thank you once again!

jamesagnew:
Thanks @cowtowncoder! CLA has been signed and submitted.

cowtowncoder:
Quick note: with CLA, I am proceeding this one -- but merging pieces manually, mostly due to challenges b/w master and 2.11 branches (wrt refactoring of `JsonReadFeature` / `JsonParser.Feature`). Should be done soon.

cowtowncoder:
Finally got it all done; added async-parser tests too (those are separate as input source is different). Thank you again for contributing this! Hoping to now release 2.11.0 by the end of the week.

jamesagnew:
@cowtowncoder Appreciate the fast response to a random PR from a new contributor! You're better at this than I am. :)

cowtowncoder:
@jamesagnew I've had some experience. But the patch was good and made this pretty easy. Adding new support for non-standard parsing is often tricky due to my over-optimization coding habits, but this turned out to be easier than I'd have expected. :)

Thank you once again for contributing this!

PR #632, opened 2020-07-19
Use HTTPS for links in README

Marcono1234:
Where supported uses HTTPS instead of HTTP for the links in the README.
It appears the certificate of your homepage (http://www.cowtowncoder.com) has expired in 2016 so I did not change these links since the user would otherwise see a warning in the browser.

Additionally the link to Codehaus (http://jackson.codehaus.org) under the section "Differences from Jackson 1.x" appears to be dead. Would it make sense to remove that link, or link to the archived versions ([Source overview](https://web.archive.org/web/20150423094711/http://jackson.codehaus.org/Source), [SVN repository](https://web.archive.org/web/20150521190506/http://svn.codehaus.org/jackson/)) instead?


cowtowncoder:
Thanks! Yes, Codehaus has long since gone away so I'll remove/change the link.

Issue #633, opened 2020-07-28
Allow skipping `RS` CTRL-CHAR to support JSON Text Sequences


quaff:
Currently jackson supports reading newline-delimited json such as [JSON Lines](http://jsonlines.org/) and [NDJSON](https://github.com/ndjson/ndjson-spec), see https://github.com/FasterXML/jackson-databind/issues/1304.
There is a proposed standard RFC7464 call [JSON Text Sequences](https://tools.ietf.org/html/rfc7464), it's similar to newline-delimited json by add a leading RS CTRL-CHAR which is not accepted by jackson.
```
com.fasterxml.jackson.core.JsonParseException: Illegal character ((CTRL-CHAR, code 30)): only regular white space (\r, \n, \t) is allowed between tokens
 at [Source: (String)"{"name":"name1"}
{"name":"name2"}
"; line: 1, column: 2]
	at com.fasterxml.jackson.core.JsonParser._constructError(JsonParser.java:1851)
	at com.fasterxml.jackson.core.base.ParserMinimalBase._reportError(ParserMinimalBase.java:707)
	at com.fasterxml.jackson.core.base.ParserMinimalBase._throwInvalidSpace(ParserMinimalBase.java:685)
	at com.fasterxml.jackson.core.json.ReaderBasedJsonParser._skipWSOrEnd(ReaderBasedJsonParser.java:2397)
	at com.fasterxml.jackson.core.json.ReaderBasedJsonParser.nextToken(ReaderBasedJsonParser.java:672)
	at com.fasterxml.jackson.databind.ObjectReader.readValues(ObjectReader.java:1898)
```
It would be nice if jackson simply treat RS as "\t" be default or `JsonReadFeature.ALLOW_UNESCAPED_CONTROL_CHARS` enabled.
here is test case:
```java
import static org.junit.Assert.assertEquals;

import java.util.ArrayList;
import java.util.List;

import org.junit.Test;

import com.fasterxml.jackson.databind.JsonNode;
import com.fasterxml.jackson.databind.MappingIterator;
import com.fasterxml.jackson.databind.ObjectMapper;

public class JSONTextSequencesTest {

	@Test
	public void test() throws Exception {
		String json = "\u001E{\"name\":\"name1\"}\n\u001E{\"name\":\"name2\"}\n";
		ObjectMapper om = new ObjectMapper();
		MappingIterator<JsonNode> it = om.readerFor(JsonNode.class).readValues(json);
[...]

PR #634, opened 2020-07-28
Skip RS CTRL-CHAR to support JSON Text Sequences

quaff:
Fix https://github.com/FasterXML/jackson-core/issues/633


cowtowncoder:
As per my notes on issue, I don't think this should be done by default, as this character is not allowed by JSON specification itself. Its acceptance should be configured separately.

Another note here is that this would only work on one of parser backends (one that takes in `Reader` or similar); there are 4 variants at this point. So changes would need for all of those.

quaff:
@cowtowncoder Just for your information, I will close this and expect your design.

PR #642, opened 2020-10-02
Applying modification pattern in #385 to FilteringParserDelegate#_nextToken2

T45K:
In #385, `FilteringParserDelegate#nextToken` was modified, but its similar method, `_nextToken2` was not modified.
Isn't the same modification needed?


cowtowncoder:
Possibly; I'd need to take a closer look.

But what would be really useful would be a test case to show that it is needed, and problem this would fix.

cowtowncoder:
Ended up making this change manually, as part of fix for #700. I think it is legit, although not sure how to reproduce specific problem.

PR #645, opened 2020-10-10
Fix test so it works on all OS

anneloreegger:
Especially on Windows one test did not pass, so I used the System.lineSeparator()-method, so it works on every Operation System.


cowtowncoder:
Thank you!

Issue #500, opened 2018-11-25
Allow "optional-padding" for `Base64Variant`


cowtowncoder:
(note: follow-up from https://github.com/FasterXML/jackson-databind/issues/2183)

Currently `Base64Variant` instances either require use of specific padding character (and use it for output) or disable use of padding completely (neither written nor accepted).
But it seems that sometimes a more lenient variant would make sense: one that

1. Does write padding (with specific character) and
2. Accepts specific padding but does not require it (i.e. ok to omit)

It seems best to add a "mutant factory" method in `Base64Variant` instance (something like `withLenientPadding(boolean)` maybe?). Whether additional entries should be added `Base64Variants` is an open question.
We'll have to figure out what to do wrt existing "usePadding()" compared to new option; API may be cumbersome if we have 2 booleans but just 3 legal combinations.

Note that we may also change API for configuration for Jackson 3.x, later on.

cowtowncoder:
This might be something for new contributors to tackle -- I can help as necessary.

pavan-kalyan:
I think I can take a shot at this.
Based on preliminary understanding, basically we can either
- use withLenientPadding(true) to return the variant which does not require padding. (factory)
or 
- add a new variant to specifically handle the 2 points you have mentioned.

I'll take a deeper look into the wikipedia article to understand more.

cowtowncoder:
@pavan-kalyan Yes, I think `withLenienPadding()` is needed at any rate (so one can modify all aspects separately, if necessary); but addition of a new variant(s) is optional. Latter because number of permutations can get big so it is not clear if new "standard" option makes sense.

I think starting with new method (and internal settings needed), implementation would be a good step: can then think about second aspect separately (can file a new issue if need be).

pavan-kalyan:
I wanted to get a better idea on the invalid combination you mentioned because the API design would need to not let the user setup an invalid combination.

Considering usesPadding and lenientPadding as internal settings in Base64Variant, the following table is meant to explain how encoding and decoding would work.

| usesPadding | lenientPadding  | Encoding Behaviour | Decoding Behaviour |
| :---: | :---: | :---: | :---: |
| true | true | add padding | does not require padding |
| true | false | add padding | requires padding |
| false | true | do not add padding | should not have  padding | 
| false | false |  do not add padding | does not require padding |

which of the above would be the invalid combination, the 3rd one?

cowtowncoder:
[...]

PR #646, opened 2020-10-17
improvement: allowing more granular control of reading behaviour for base64

pavan-kalyan:
As Discussed in #500.

Allowing configuration of read behaviour via trio methods (withPadddingForbidden, withPaddingAllowed, withPaddingRequired)
existing usage of usesPadding now changes meaning to writePadding. (maybe this should be renamed?)
also there is a method included withWritePadding(boolean) to allow configuration of whether padding should be written derived from the Base64Variants.

Existing constructors set the following read behaviour. if usesPadding is enabled, then Padding on read is required. if usesPadding is disabled then Padding on Read is forbidden
Testing and validation pending.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/Base64Variant.java
+++ b/src/main/java/com/fasterxml/jackson/core/Base64Variant.java
@@ -603,6 +603,9 @@ public final class Base64Variant
                 if (bits != Base64Variant.BASE64_VALUE_PADDING) {
                     _reportInvalidBase64(ch, 2, null);
                 }
+                if (paddingReadBehaviour().equals(PaddingReadBehaviour.PADDING_FORBIDDEN)) {
+                    _reportBase64UnexpectedPadding();
+                }
                 // Ok, must get padding
                 if (ptr >= len) {
                     _reportBase64EOF();
@@ -620,8 +623,8 @@ public final class Base64Variant
             decodedData = (decodedData << 6) | bits;
             // fourth and last base64 char; can be padding, but not ws
             if (ptr >= len) {
-                // but as per [JACKSON-631] can be end-of-input, iff not using padding
-                if (!usesPadding()) {
+                // but as per [JACKSON-631] can be end-of-input, iff padding on read is not required
+                if (!paddingReadBehaviour().equals(PaddingReadBehaviour.PADDING_REQUIRED)) {
                     decodedData >>= 2;
                     builder.appendTwoBytes(decodedData);
                     break;
@@ -633,6 +636,9 @@ public final class Base64Variant
             if (bits < 0) {
                 if (bits != Base64Variant.BASE64_VALUE_PADDING) {
                     _reportInvalidBase64(ch, 3, null);
+                }
+                if (paddingReadBehaviour().equals(PaddingReadBehaviour.PADDING_FORBIDDEN)) {
+                    _reportBase64UnexpectedPadding();
                 }
                 decodedData >>= 2;
                 builder.appendTwoBytes(decodedData);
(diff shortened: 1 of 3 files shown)


cowtowncoder:
Excellent! I'll try to have a look over the weekend (there's a healthy backlog of things but I want to make sure this still gets in 2.12).

One other thing is the CLA; if I haven't asked for one, it's here:

and the easiest way usually is to print it, fill & sign, scan/photo, email to `info` at fasterxml dot com.
Only needs to be done once, and I can merge PRs after we receive and file it.

src/main/java/com/fasterxml/jackson/core/Base64Variant.java
```diff
@@ -87,6 +87,11 @@
      */
     private final transient boolean _usesPadding;
 
+    /**
+     * Whether padding characters should be required or not while decoding
+     */
+    private final transient PaddingReadBehaviour _paddingReadBehaviour;
```
cowtowncoder, line 93: Let's remove `transient` here -- not sure why preceding final field has it, possibly copy-paste issue originally (I will remove it from others too)

src/main/java/com/fasterxml/jackson/core/Base64Variant.java
```diff
@@ -154,6 +165,11 @@ public Base64Variant(Base64Variant base, String name, int maxLineLength)
      * line length) differ
      */
     public Base64Variant(Base64Variant base, String name, boolean usesPadding, char paddingChar, int maxLineLength)
+    {
```
cowtowncoder, line 168: Please add `@since 2.12` for new non-private methods (helps with Javadocs wrt public methods; internal changes for other types).

cowtowncoder:
Looks pretty good; one thing missing (aside from CLA) would be unit tests: just basic demonstration that handling works wrt new settings.

src/main/java/com/fasterxml/jackson/core/Base64Variant.java
pavan-kalyan: This particular highlighted constructor is not new, So I haven't added the @since prefix for this.

pavan-kalyan:
I have made the requested changes (let me know if there are more scenarios/flows that need to be unit tested)
Is there any way to digitally sign the CLA, instead of printing it out?

cowtowncoder:
@pavan-kalyan A few contributors have managed to do it all digitally -- not sure how unfortunately (to share details), but we are fine with digital print as signature, no need to produce physical copy. Just need an image (pdf, png, jpeg) with information and signature of some kind (like ones that are used on online documents, not necessarily based on your handwriting).

pavan-kalyan:
I have sent the CLA, to the email specified

src/main/java/com/fasterxml/jackson/core/Base64Variant.java
```diff
@@ -166,6 +182,47 @@ public Base64Variant(Base64Variant base, String name, boolean usesPadding, char
...
+
+    /**
+     * @since 2.12
+     */
+    public Base64Variant withWritePadding(boolean writePadding) {
+        return new Base64Variant(this, this._name, writePadding, this._paddingChar, this._maxLineLength);
+
+    }
+
+    private enum PaddingReadBehaviour {
```
cowtowncoder, line 221: Since exposed by public accessor, needs to be public type (I can change this post-merge).

cowtowncoder:
Looks good. Could use some more javadocs in general, but one area where I think description is needed is for `Base64Variants`, to explain the logic of "if outputs padding, requires on reading; if does not output padding, does not accept", for existing choices.

I hope to get this merged tomorrow, and did receive CLA (thank you!).

src/main/java/com/fasterxml/jackson/core/Base64Variant.java
pavan-kalyan: Removed transient

src/main/java/com/fasterxml/jackson/core/Base64Variant.java
pavan-kalyan: Changed and added comments and `@since`

pavan-kalyan:
I have added more comments, let me know if there should be more docs.
One doubt I have if these comments are in the right place (https://github.com/FasterXML/jackson-core/pull/646/files#diff-a9b7b3f0fa4f0eb74344f465fc1a1faa6092f4b3a0642809b782c7af68acb923R15-R18). Should these be just above every Variant mentioned?

cowtowncoder:
@pavan-kalyan Thanks! I think that comment makes sense both there AND a brief note in each variant (in latter case can just say that "writes padding on output; does not accept padding when reading (may change latter with a call to [LINK to method to change])"?

pavan-kalyan:
Changed to `writes padding on output; does not accept padding when reading (may change latter with a call to {@link Base64Variant#withWritePadding}`

cowtowncoder (approved):
LGTM.

cowtowncoder:
Merged to 2.12, up to master as well. Thank you very much for getting this through!

pavan-kalyan:
Thanks for patiently guiding me through this!

PR #647, opened 2020-10-25
Fixed NPE in writeNumber(String) method of UTF8JsonGenerator and WriterBasedJsonGenerator

pakru:
I faced NPE in writeNumber(String) of UTF8JsonGenerator when trying to pass null into this method.

`gen.writeNumber(getStringValue());`
where `getStringValue()` method can return null, and this causes NPE in writeRaw(String text) method of generator. 
But it works fine for writeNumber(BigDecimal) and writeNumber(BigInteger) methods if I pass null into those methods.

Thus, I added null check for `writeNumber(String encodedValue)`  the same as it made for BigDecimal and BigInteger types. Furthermore modified some tests for this nullability check.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/json/UTF8JsonGenerator.java
+++ b/src/main/java/com/fasterxml/jackson/core/json/UTF8JsonGenerator.java
@@ -1054,7 +1054,9 @@ public class UTF8JsonGenerator
     public void writeNumber(String encodedValue) throws IOException
     {
         _verifyValueWrite(WRITE_NUMBER);
-        if (_cfgNumbersAsStrings) {
+        if (encodedValue == null) {
+            _writeNull();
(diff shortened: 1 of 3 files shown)


cowtowncoder:
Looks legit, thank you! (and especial thanks for improving tests).

One practical thing before I merge this: unless I have asked for and received CLA (only needs to be done once, good for all PRs for Jackson repositories), I'd need this:

and usually simplest way is to print, fill & sign, scan/photo, email to `info` at fasterxml dot com.
Once I receive it, we are good to go!

Thank you once again for your contribution!

pakru:
Thanks for you feedback! 
I have sent you signed CLA. Check you email, please.

cowtowncoder:
Thanks! Received, will merge.

Issue #649, opened 2020-11-07
Bug in FilteringParserDelegate


jhaber:
There seems to be a bug in `FilteringParserDelegate` that specifically manifests when trying to filter out arrays. Here's a test that reproduces the behavior (can be added to `BasicParserFilteringTest`):
```java
    public void testExcludeArrays() throws Exception
    {
        class NoArraysFilter extends TokenFilter
        {
            @Override
            public TokenFilter filterStartArray() {
                return null;
            }
        }

        String jsonString = aposToQuotes("{'a':123,'array':[1,2]}");
        JsonParser p0 = JSON_F.createParser(jsonString);
        FilteringParserDelegate p = new FilteringParserDelegate(p0,
            new NoArraysFilter(),
            true, // includePath
            true // multipleMatches
        );
        String result = readAndWrite(JSON_F, p);
        assertEquals(aposToQuotes("{'a':123}"), result);
        assertEquals(1, p.getMatchCount());
    }
```

This test fails with:
```
junit.framework.ComparisonFailure: 
Expected :{"a":123}
Actual   :{"a"}
```

I discovered this bug while adding test coverage for #573. However, I confirmed that the bug exists in jackson-core versions 2.8.9, 2.9.10, 2.10.5, and 2.11.3 (and therefore wasn't introduced in #573).

No one has complained yet, and a quick GitHub search suggests there isn't much code in the wild (if any) doing `filterStartArray() { return null; }` so I don't think it's a high priority to fix this. I mainly just wanted to log this issue so that I can link to it (above the commented out unit test)

vincent-simpson:
@jhaber I see you added this test case to `BasicParserFilteringTest` in the failing tests package, but I believe the assert statement is incorrect.
https://github.com/FasterXML/jackson-core/pull/650/files#diff-21191a4efccbbbe22296418ee9163a5a85ff776f4047d3f166916ec54299eaa1R34

The number of expected matches should be 1 here as you mentioned above. So this should be removed as a failing test since the assert is incorrect?

jhaber:
When I opened this issue the test was failing even before it got to that assertion. The filter was somehow causing Jackson to generate invalid JSON (I was getting a JSON object with a key but no value):
```
junit.framework.ComparisonFailure: 
Expected :{"a":123}
Actual   :{"a"}
```

I just checked out the 2.14 branch and seems like it's now generating the expected JSON, so the underlying bug must've been fixed at some point 😄 

[...]

PR #650, opened 2020-11-07
Fix failing tests

jhaber:
Follow-up to #573 to fix 2 failing tests. 

The failing `BasicGeneratorFilteringTest` was just because `multipleMatches` needs to be set to true (and was accidentally flipped to false in #573). The test was updated in 6318aeea3368f8d065535ad9d032de7ca79130be in a way that requires this flag to be true.

The failing `BasicParserFilteringTest` is the issue I described [here](https://github.com/FasterXML/jackson-core/pull/573#issuecomment-546601957) (which appears to have existed since the filtering functionality was introduced). I filed #649 and ignored the test for now.

While I was in there, I also extended `BasicParserFilteringTest#testValueOmitsFieldName2` slightly to make the input more complex

### DIFF
--- a/src/test/java/com/fasterxml/jackson/core/filter/BasicParserFilteringTest.java
+++ b/src/test/java/com/fasterxml/jackson/core/filter/BasicParserFilteringTest.java
@@ -80,14 +80,6 @@ public class BasicParserFilteringTest extends BaseTest
         protected boolean _includeScalar() { return false; }
     }
 
-    static class NoArraysFilter extends TokenFilter
-    {
-        @Override
-        public TokenFilter filterStartArray() {
-            return null;
-        }
-    }
-
     static class NoObjectsFilter extends TokenFilter
     {
         @Override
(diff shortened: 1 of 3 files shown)


src/test/java/com/fasterxml/jackson/core/filter/BasicParserFilteringTest.java
```diff
@@ -423,7 +423,8 @@ public void testNoMatchFiltering6() throws Exception
         assertEquals(0, p.getMatchCount());
     }
 
-    public void testValueOmitsFieldName1() throws Exception
+    // https://github.com/FasterXML/jackson-core/issues/649
```
cowtowncoder, line 426: Could you create a stand-alone test for this, under ".../failing/"? That way it's easy to locate, can be run manually, but won't be run as part of build (same as renaming here)?

jhaber: Sure thing, didn't know about that directory. Fixed in 213b8143eef161096d11667e49de6f7e02ad8bf5

cowtowncoder:
Excellent thanks!

Issue #577, opened 2019-11-06
Improve performance of floating-point number parsing


cowtowncoder:
Jsoniter project (https://github.com/plokhotnyuk/jsoniter-scala) has many impressive performance optimizations; linked f.ex from here:

https://www.reddit.com/r/java/comments/darehu/jackson_release_210/f1ysf1e/

Of ones included, number-parsing would be relevant for this repo.

EDIT: also see (from the comment below)

"Unrelated to jsoniter but this recent port of Lemire's Double parser:

https://github.com/wrandelshofer/FastDoubleParser

and the original paper https://arxiv.org/abs/2101.11408 also relevant"

cowtowncoder:
And specifically, for `BigDecimal` case:

https://github.com/plokhotnyuk/jsoniter-scala/blob/b51f90b87923305b6fa177a4da909ff43da3c7b7/jsoniter-scala-core/src/main/scala/com/github/plokhotnyuk/jsoniter_scala/core/JsonReader.scala#L1477-L1591

anneloreegger:
I'd like to work on this issue, is it still available?

cowtowncoder:
@anneloreegger Yes, this is available! I'll assign it to you (no obligation to work, just a marker so others know you are considering it).

anneloreegger:
Just to be sure I understand the issue correctly. Shall I especially improve the parsingMethods inside the NumberInput-class? (especially parseBigDecimal)
And should this be done on 2.12 or master branch?
Thanks for the answers :)

cowtowncoder:
@anneloreegger Yes, I think `NumberInput` has stubs that would allow for this. On 2.12 vs master -- ideally 2.12, if this seems doable on short term with well contained changes. But `master` if there were need to change API or internal interfaces.
I have not really investigated jsoniter's approach enough to know how involved this gets unfortunately.

cowtowncoder:
NOTE: `BigDecimal` use case covered by #677 (for 2.13), will edit title

LifeIsStrange:
Unrelated to jsoniter but this recent port of Lemire's Double parser is 5.5X faster than the default double parser:
(if the microbenchmark generalize)
https://github.com/wrandelshofer/FastDoubleParser

cowtowncoder:
@LifeIsStrange thank you for sharing this! I hope someone might have time to maybe investigate possibility of PR for improvements here.

abc12345678912345:
[...]

PR #677, opened 2021-02-02
Introduce O(n^1.5) BigDecimal parser implementation

ferenc-csaky:
Covers the `BigDecimal` part of #577. The `double`/`float` implementations would require a much bigger refactor, since the example implementation follows a very different approach, but IMO it's worth to update only the `BigDecimal` without the others as well.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/io/BigDecimalParser.java
+++ b/src/main/java/com/fasterxml/jackson/core/io/BigDecimalParser.java
@@ -1,0 +1,155 @@
+package com.fasterxml.jackson.core.io;
+
+import java.math.BigDecimal;
+
+// Based on a great idea of Eric Obermühlner to use a tree of smaller BigDecimals for parsing really big numbers
+// with O(n^1.5) complexity instead of O(n^2) when using the constructor for a decimal representation from JDK 8/11:
+// https://github.com/eobermuhlner/big-math/commit/7a5419aac8b2adba2aa700ccf00197f97b2ad89f
+public final class BigDecimalParser {
+
+    private final char[] chars;
+    private final int off;
+    private final int len;
+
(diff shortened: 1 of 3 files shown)


cowtowncoder:
@ferenc-csaky Sounds good -- thank you very much for the contribution! -- and it looks like I have CLA from you already.

Just one request: could this be rebased against 2.13? `master` is for Jackson 3.0 and I think this improvement would be great for 2.13 -- I can handle merging to `master`.

ferenc-csaky:
Changed the base branch to `2.13`. My original plan was this as well, my dev. branch also originated from `2.13`, but it seems I missed it in the PR. Corrected now, thanks!

cowtowncoder:
Just realized that none of the tests verifies new code path(s), I think. Will repurpose some from original... should have caught this earlier,.

cowtowncoder:
@ferenc-csaky  Hmmh. I think there may be a bug somewhere, some offset not being carried or used for calculations?
I added a test in 2.13 from the original changes and it looks like offset used for splits is `0` for some reason.
As far as I can see it gets passed correctly to `BigDecimalParser` (first suspected it might not but seems to).

One thing that is tricky, too, is how to propagate underlying secondary fails, if those are possible (not sure if they are after fixing this).

cowtowncoder:
Looks like `int numIdx = 0` in `parseBigDecimal(int splitLen)` is wrong: that should get set to `off`, somewhere. But simple attempts did not produce expected results.

ferenc-csaky:
Oh yes, you are totally right, I didn't follow the source impl. exactly and I messed it up. Sorry about it. I just created a PR with the fix. Now the test you added runs correctly.

cowtowncoder:
@ferenc-csaky np. Just glad this was caught :)

PR #678, opened 2021-02-05
Fix BigDecimalParser

ferenc-csaky:
Fixes a bug introduced in #677.


cowtowncoder:
Thanks! When doing additional verification, actually found a bug that is in similar area (see #679) (but that code here would properly catch).

PR #680, opened 2021-02-22
#2013: add support for reading from / writing to a Path

sdoeringNew:
Is there no need to write Unit tests? The changed ``TextualTSFactory`` has barely coverage.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/TokenStreamFactory.java
+++ b/src/main/java/com/fasterxml/jackson/core/TokenStreamFactory.java
@@ -7,6 +7,8 @@ package com.fasterxml.jackson.core;
 import java.io.*;
 import java.lang.ref.SoftReference;
 import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.List;
 import java.util.Locale;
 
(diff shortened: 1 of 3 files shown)


sdoeringNew:
After getting the permission to add some Unit tests those tests have been added.

Have you noticed that the Unit tests in this project are still JUnit 3?

cowtowncoder:
Junit dependency itself is to 4.x isn't it? As to test style, if it ain't broke... if there are specific improvement ideas those are always good to suggest of course; and if there are questions of why certain things are certain way, can file issues.

cowtowncoder:
Merged; big thank you for increased test coverage! Although methods are quite trivial, there have been cases where even simplest implementations have had some odd edge cases (occasionally silly things like infinite recursion due to refactoring etc).

cowtowncoder:
Quick note: missed the fact that `BinaryTSFactory` also needs new methods... shouldn't be too difficult to add, but just noticed when trying to compile binary format backends.
Just noting for posterity.

sdoeringNew:
> Quick note: missed the fact that `BinaryTSFactory` also needs new methods... shouldn't be too difficult to add, but just noticed when trying to compile binary format backends.
> Just noting for posterity.

Good to know. I'll look at it the next two weeks.

cowtowncoder:
@sdoeringNew no need; I added the methods :)
(plus added "tests" that subclass binary/textual base types, so if adding abstract methods should notice missing implementations [unless IDE tries to be too helpful etc])

PR #682, opened 2021-03-03
Enable caching for travis

Taher-Ghaleb:
Would be interested to know why caching hasn't been enabled in this repository. Thank you.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,10 @@ jdk:
   - openjdk8
   - oraclejdk11
 
+cache:
+  directories:
+  - $HOME/.m2
+
 # Below this line is configuration for deploying to the Sonatype OSS repo
 # https://knowm.org/configure-travis-ci-to-deploy-snapshots/
 before_install: "git clone -b travis `git config --get remote.origin.url` target/travis"


cowtowncoder:
Mostly because caching didn't seem to speed up builds and introduced occasional mysterious fails only solved by clearing caches manually.

Other than that Travis caching was a total smashing success. :-I

Taher-Ghaleb:
Thanks, @cowtowncoder. I appreciate your detailed comment.

cowtowncoder:
Oh, one thing I realized after adding my comment: I think that a big reason why caching might not work as well for `jackson-core` is that it has very few external dependencies (just junit). I am not sure if Maven plugins are cached: it could be that they are not, and if so, caching might help there.

The challenges are probably due to complexities of snapshot dependencies across Jackson components. This may not be a big problem for `jackson-core` as it also only depends on `jackson-base`/`jackson-bom`, but has been problematic for some other components.
For what it is worth, I think caching was enabled at some point -- at least it has been used for other Jackson components.


So maybe caching could be (re)enabled for this repo. I just haven't had enough time to figure best usage patterns. If you are more familiar with its use, or have articles to point to, that'd be great to get.

Taher-Ghaleb:
Thanks, @cowtowncoder, for your follow-up comment. I got your point.

I am currently doing research about caching in CI to investigate current practices and reasons related to enabling/disabling it in software repositories and to understand its potential consequences. I am still approaching information/evidences from different sources to have a big picture of that. Thank you again for your gentle responsiveness.

lastthyme (approved):

cowtowncoder:
Moving to Github actions so this is becoming moot, will close.

PR #744, opened 2022-03-31
Limit size of exception message in BigDecimalParser

pjfanning:
new limit may even be too large

### DIFF
--- a/src/test/java/com/fasterxml/jackson/core/io/BigDecimalParserTest.java
+++ b/src/test/java/com/fasterxml/jackson/core/io/BigDecimalParserTest.java
@@ -1,0 +1,18 @@
+package com.fasterxml.jackson.core.io;
+
+public class BigDecimalParserTest extends com.fasterxml.jackson.core.BaseTest {
+    public void testLongStringParse() {
+        final int len = 1500;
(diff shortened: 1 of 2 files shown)


src/test/java/com/fasterxml/jackson/core/io/BigDecimalParserTest.java
```diff
@@ -0,0 +1,18 @@
...
+        final int len = 1500;
+        final StringBuilder sb = new StringBuilder(len);
+        for (int i = 0; i < len; i++) {
+            sb.append("A");
+        }
+        try {
+            BigDecimalParser.parse(sb.toString());
+            fail("expected NumberFormatException");
+        } catch (NumberFormatException nfe) {
+            assertTrue("exception message starts as expected?", nfe.getMessage().startsWith("Value \"AAAAA"));
```
cowtowncoder, line 14: There is also helper method `verifyException` in test base class for simple inclusion checks.
(this is fine too)

cowtowncoder:
Looks good; I think 1000 is reasonable.

cowtowncoder:
Thank you @pjfanning! Will go in 2.13.3 release.

Issue #514, opened 2019-01-26
Use Ryu for Float/Double to String conversions


re-thc:
Hi, see https://github.com/ulfjack/ryu and the paper / benchmarks / information. Might help a lot in JSON output from Float / Double.

Cheers!

cowtowncoder:
@hc-codersatlas Thank you for the link! Looks potentially interesting. I don't have much time to look into this right now, but maybe someone else has, so it's good to have an issue filed.

hc-codersatlas:
Sounds good. Hopefully it gets picked up. Jsoniter adopted this and is 1 of the reasons it's faster.

vy:
For the records, [jsoniter](/plokhotnyuk/jsoniter-scala) fame @plokhotnyuk kindly shared the following links in [a Reddit comment](https://www.reddit.com/r/java/comments/darehu/jackson_release_210/f1ysf1e):

1. [Fast & medium paths for parsing of doubles](/plokhotnyuk/jsoniter-scala/blob/db689583dc35cd3dd457e9edb0b7fbcc10a5d508/jsoniter-scala-core/src/main/scala/com/github/plokhotnyuk/jsoniter_scala/core/JsonReader.scala#L1243-L1303)
2. [Fast & medium paths for parsing of floats](/plokhotnyuk/jsoniter-scala/blob/db689583dc35cd3dd457e9edb0b7fbcc10a5d508/jsoniter-scala-core/src/main/scala/com/github/plokhotnyuk/jsoniter_scala/core/JsonReader.scala#L1388-L1441)
3. [Fast parsing of BigDecimals and using them to parse BigIntegers](/plokhotnyuk/jsoniter-scala/blob/db689583dc35cd3dd457e9edb0b7fbcc10a5d508/jsoniter-scala-core/src/main/scala/com/github/plokhotnyuk/jsoniter_scala/core/JsonReader.scala#L1611-L1724)
4. [Fast parsing of ISO-8601 representation of java.time.* instances](/plokhotnyuk/jsoniter-scala/blob/db689583dc35cd3dd457e9edb0b7fbcc10a5d508/jsoniter-scala-core/src/main/scala/com/github/plokhotnyuk/jsoniter_scala/core/JsonReader.scala#L1747-L2240)
[...]

PR #749, opened 2022-04-01
fast float/double writer option

pjfanning:
relates to https://github.com/FasterXML/jackson-core/issues/514


src/test/java/com/fasterxml/jackson/core/io/schubfach/FloatToStringTest.java
```diff
@@ -0,0 +1,106 @@
...
+    assertF2sEquals("4103.9004", 4103.9003f);
+    assertF2sEquals("5.3399997E9", 5.3399997E9f);
+    assertF2sEquals("6.0898E-39", 6.0898E-39f);
+    assertF2sEquals("0.0010310042", 0.0010310042f);
+    assertF2sEquals("2.882326E17", 2.8823261E17f);
+    assertF2sEquals("7.038531E-26", 7.038531E-26f);
+    assertF2sEquals("9.223404E17", 9.2234038E17f);
+    assertF2sEquals("6.710887E7", 6.7108872E7f);
+    //TODO investigate
+    assertF2sEquals("9.8E-45", 1.0E-44f);
```
pjfanning, line 92: @plokhotnyuk I took some of the Ryu tests and used them to test Schubfach and this line 92 is one strange diff - the assert works as "1.0E-44f" in Ryu but I had to hack the value here for Schubfach. Any ideas?

plokhotnyuk: Results from Schubfach are rounded with exact precision of 2 decimal digits.

Just see exact representation of float with significant 7:
https://float.exposed/0x00000007

pjfanning:
@cowtowncoder the Ryu and Schubfach algorithms don't output exactly the same values as JDK Double.toString - at least for some edge cases - would there be anyway to have jackson-core NumberOutput support JDK approach as default and to have an option to use Schubfach (which is faster than Ryu) if the user enables a flag?

cowtowncoder:
@pjfanning I think it is reasonable to have a non-default flag to enable "fast but incorrect" (I think? JDK implementations is, from what I recall, quite precisely compatible well-defined criteria of closest value, round-trupping). As with `NumberInput` there is the question of how to handle configurability; I'm pretty sure it should default to JDK for legacy cases where boolean setting cannot be passed.

pjfanning:
@cowtowncoder I'll probably hold back on doing anything with this writer code until the fast double parser work is merged. I think that is getting close enough to something that could be considered for merging.

cowtowncoder:
Ok I think this can be closed?

pjfanning:
@cowtowncoder the recent merges affect parsing (and deserialization) - this PR affects serialization - there is still a little bit of work to do on this but it is close enough to being ready

src/main/java/com/fasterxml/jackson/core/json/JsonWriteFeature.java
```diff
@@ -72,6 +72,15 @@
...
     ESCAPE_NON_ASCII(false, JsonGenerator.Feature.ESCAPE_NON_ASCII),
 
+    /**
+     * Feature that determines whether to use standard Java code to write floats/doubles (default) or
+     * use the Schubfach algorithm which is faster. The latter approach may lead to small
+     * differences in the precision of the float/double that is written to the JSON output.
+     *
+     * @since 2.14
+     */
+    USE_FAST_DOUBLE_WRITER(false, JsonGenerator.Feature.USE_FAST_DOUBLE_WRITER),
```
cowtowncoder, line 82: As with reader side, let's move this to more general `StreamWriteFeature` as it need not be JSON specific but applicable to most if not all textual format backends.

cowtowncoder:
@pjfanning d'oh! I really should read the description in detail before commenting something stupid. Yes, that makes sense.
Just let me know when merging would make sense. Could be split in multiple parts too but may not be necessary.

One thing that makes things easier to merge is that 3.0 (master) has immutable factories, so using builders (JsonFactory.builder().... or equivalent from `BaseTest`) avoids some merging issues. But I can usually handle those quite easily, just cause intermediate merge conflict.

pjfanning:
@cowtowncoder is this PR now a manageable size? It will have some merge issues in master but shouldn't be too bad.

Issue #750, opened 2022-04-19
remove workaround for nasty double


pjfanning:
According to https://www.exploringbinary.com/java-hangs-when-converting-2-2250738585072012e-308/ - the issue with this number was fixed in 2011. Instead of taking the hit to check for this string when parsing doubles, the code could be removed - eg for jackson v2.14. 

https://www.exploringbinary.com/fpupdater-fixes-the-java-2-2250738585072012e-308-bug/

https://github.com/FasterXML/jackson-core/blob/2.14/src/main/java/com/fasterxml/jackson/core/io/NumberInput.java#L302

```
        // [JACKSON-486]: avoid some nasty float representations... but should it be MIN_NORMAL or MIN_VALUE?
        /* as per [JACKSON-827], let's use MIN_VALUE as it is available on all JDKs; normalized
         * only in JDK 1.6. In practice, should not really matter.
         */
        if (NASTY_SMALL_DOUBLE.equals(s)) {
            return Double.MIN_VALUE;
        }
```

If we proceed with https://github.com/FasterXML/jackson-core/issues/746, then maybe this change is safer.

cowtowncoder:
Ok sounds reasonable. But just to make sure; the fix:

* Is in JDK, 1.6.x (x > 0)
* All JDK 8 (and above) versions have the fix

So in theory one might be able to find a JDK 1.6 version with the problem, use it with `jackson-core` -- but that seems highly unlikely.

I think I'd accept the change then. Will add just one suggestion in PR and can then merge in for 2.14.

cowtowncoder:
Ahhh. Need to revert part of this, as `jackson-databind` has dependency: will mark constant as deprecated but not (yet) remove.

PR #751, opened 2022-04-19
Remove workaround for old issue with a particular double

pjfanning:
https://github.com/FasterXML/jackson-core/issues/750


cowtowncoder:
As per my comment on #750, sounds reasonable.

Only one request: could you add a note on `release-notes/VERSION-2.x` for this change?

pjfanning:
added the release note

Issue #753, opened 2022-04-21
Add `NumberInput.parseFloat()`


pjfanning:
For v2.14, NumberInput has become the main way to centralise the code for parsing numbers. One inconsistency is that Floats are still parsed using Float.parseFloat. Having a NumberInput.parseFloat that does that means we can change the implementation later to use the fast-double-parser or something similar. See https://github.com/FasterXML/jackson-core/issues/577

pjfanning:
@plokhotnyuk I think the jackson issue you describe is because in some cases jackson seems to use Double.parseDouble to parse floats and then later to cast that to a float and that seems to lead to the loss of precision with the "1.199999988079071" case. `Float.parse("1.199999988079071")` gives 1.1999999f but `(float)Double.parse("1.199999988079071")` seems to give `1.2f`.

pjfanning:
@plokhotnyuk is 1.2f actually a better answer than 1.999999f - it is actually closer to 1.199999988079071
```
1.2 - 1.199999988079071 = 1.19209289e-8
1.999999 - 1.199999988079071 = -8.80790711e-8
```

plokhotnyuk:
Exact binary representation of `1.199999988079071` is:
```
1.0011001100110011001100101111111111111111111111111111110011000011...
```
When fitting to 32-bit float the properly rounded binary value is:
```
1.00110011001100110011001
```
That is `1.19999992847442626953125` when converted back to exact decimal representation.

You can use the following online calculator to check this:
https://www.exploringbinary.com/binary-converter/

pjfanning:
@cowtowncoder this code in ParserBase (jackson-core) may be contributing to the issue that @plokhotnyuk described with parsing `1.199999988079071` as a float.

```
    @Override
    public float getFloatValue() throws IOException
    {
        double value = getDoubleValue();
        /* 22-Jan-2009, tatu: Bounds/range checks would be tricky
         *   here, so let's not bother even trying...
         */
        /*
        if (value < -Float.MAX_VALUE || value > MAX_FLOAT_D) {
            _reportError("Numeric value ("+getText()+") out of range of Java float");
        }
        */
        return (float) value;
    }
```

From my testing, it seems that casting doubles to floats may not get the same result as parsing the text representation directly with Float.parseFloat(String).

pjfanning:
I created https://github.com/FasterXML/jackson-core/issues/755 for the issue about 1.199999988079071

GedMarc:
[...]

PR #754, opened 2022-04-21
method to parse a float

pjfanning:
https://github.com/FasterXML/jackson-core/issues/753


cowtowncoder:
@pjfanning  I think we should go ahead and merge this in, regardless of the other issues: it seems like a basic building block we need. Issues with coercions can be tackled separately.

So unless you have concerns, please convert to non-draft PR and I'll merge it?

pjfanning:
@cowtowncoder I'm happy to have this merged as a builing block. Some of the other discussions are independent and could be scala specific

Issue #755, opened 2022-04-22
Change ParserBase to track floats independently from doubles


pjfanning:
It seems like in some cases, casting a double to a float may result in a slightly different answer than parsing the original number string using Float.parseFloat.

This code in ParserBase (jackson-core) may be contributing to an issue that @plokhotnyuk described with parsing 1.199999988079071 as a float.

```
    @Override
    public float getFloatValue() throws IOException
    {
        double value = getDoubleValue();
        /* 22-Jan-2009, tatu: Bounds/range checks would be tricky
         *   here, so let's not bother even trying...
         */
        /*
        if (value < -Float.MAX_VALUE || value > MAX_FLOAT_D) {
            _reportError("Numeric value ("+getText()+") out of range of Java float");
        }
        */
        return (float) value;
    }
```

From my testing, it seems that casting doubles to floats may not get the same result as parsing the text representation directly with Float.parseFloat(String).

pjfanning:
@plokhotnyuk could you check the test case in https://github.com/FasterXML/jackson-databind/pull/3463 - result seems correct as is

cowtowncoder:
Hmmh. Yes, I agree.

Unfortunately this may also require more changes for cases where coercion is used (at least for textual formats) -- in case where `double` is requested first, decoded `double` will be used as the base to coerce from (due to assumption that coercion is faster than decoding).

I think what would be really useful is/are test cases to trigger the problem.

One quick question: was the specific textual representation `1.2`? If so, could just verify that the `float` we get is same as `Float.parseFloat("1.2")`, for example.

pjfanning:
@cowtowncoder I still haven't reproduced the problem - so for now, this can be ignored. When I get a chance, I'll try the jsoniter-scala benchmark scenario directly to see if I can debug it there.

pjfanning:
@cowtowncoder I added https://github.com/FasterXML/jackson-databind/pull/3470 - it shows the issue that @plokhotnyuk reported. Would you be able to take a look?

cowtowncoder:
I think this was indeed resolved via #757, closing.

PR #757, opened 2022-04-26
Update ParserBase to support floats directly

pjfanning:
Related to https://github.com/FasterXML/jackson-core/issues/755

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/base/ParserBase.java
+++ b/src/main/java/com/fasterxml/jackson/core/base/ParserBase.java
@@ -892,6 +892,9 @@ public abstract class ParserBase extends ParserMinimalBase
             if (expType == NR_BIGDECIMAL) {
                 _numberBigDecimal = _textBuffer.contentsAsDecimal();
                 _numTypesValid = NR_BIGDECIMAL;
+            } else if (expType == NR_FLOAT) {
+                _numberFloat = _textBuffer.contentsAsFloat();
+                _numTypesValid = NR_FLOAT;
             } else {
                 // Otherwise double has to do
                 _numberDouble = _textBuffer.contentsAsDouble();
--- a/src/main/java/com/fasterxml/jackson/core/base/ParserBase.java
+++ b/src/main/java/com/fasterxml/jackson/core/base/ParserBase.java
@@ -193,6 +193,8 @@ public abstract class ParserBase extends ParserMinimalBase
 
     protected long _numberLong;
 
+    protected float _numberFloat;
+
     protected double _numberDouble;
 
     // And then object types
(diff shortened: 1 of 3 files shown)


src/main/java/com/fasterxml/jackson/core/base/ParserBase.java
```diff
@@ -735,7 +745,15 @@ public float getFloatValue() throws IOException
             _reportError("Numeric value ("+getText()+") out of range of Java float");
         }
         */
-        return (float) value;
+        if ((_numTypesValid & NR_FLOAT) == 0) {
+            if (_numTypesValid == NR_UNKNOWN) {
+                _parseNumericValue(NR_FLOAT);
+            }
+            if ((_numTypesValid & NR_FLOAT) == 0) {
+                convertNumberToFloat();
```
cowtowncoder, line 753: Is this called from anywhere? Or is it meant to be called by databind and/or Scala module?

Also:

cowtowncoder:
Hmmmh. I think it makes sense, although I have a suspicion there might be a conflict with some of binary backends since they add `_numberFloat` on their own. So it'd be good to have a look at `jackson-dataformats-binary` to see if that is a problem (masking of fields is I think legal but very confusing and likely to cause fials).

src/main/java/com/fasterxml/jackson/core/base/ParserBase.java
pjfanning: tbh much of the change is just me copy/pasting the NR_DOUBLE bits and having NR_FLOAT bits that match - the `_parseSlowFloat` change is the one that fixes the broken test case - it was still broken till I made that additional change - some of the other changes that were made in this PR prior to that may not be necessary

pjfanning:
@cowtowncoder with https://github.com/FasterXML/jackson-databind/pull/3470 merged, the jackson-databind build will fail - is it possible to merge this to fix it?

cowtowncoder:
@pjfanning I'm working on re-creating test at `jackson-core` and will either move databind test under `failing` or doing something -- am aware that in the meantime tests fail. Also hoping to merge this soon, yes!

src/main/java/com/fasterxml/jackson/core/base/ParserBase.java
```diff
@@ -678,6 +686,9 @@ public NumberType getNumberType() throws IOException
         if ((_numTypesValid & NR_BIGDECIMAL) != 0) {
             return NumberType.BIG_DECIMAL;
         }
+        if ((_numTypesValid & NR_FLOAT) != 0) {
```
cowtowncoder, line 689: I have some concerns wrt backwards-compatibility here but...  I guess there's already dependency wrt calls made that affect type so maybe this is fine.

cowtowncoder (approved):

cowtowncoder:
Ok decided that it makes more sense to merge this soon; I think it looks good and if there are issues wrt other parsers, can figure these out later on.

I think that there are probably remaining coercion cases where we do NOT want to coerce between `float`, `double` and/or `BigDecimal` (since results differ from actual parsing) but I think this is a solid first step.

cowtowncoder:
Ok so 2.14 version seems to work with changes, as expected. Getting failures from master (3.0) probably due to refactoring of code (some codepaths not updated appropriately).

cowtowncoder:
Ah, found it. A method from `ParserBase` to new `JsonParserBase`.

Issue #736, opened 2021-12-29
`JsonPointer` quadratic memory use: OOME on deep inputs


htmldoug:
Adding https://github.com/nst/JSONTestSuite in https://github.com/rallyhealth/weePickle/pull/106 uncovered that `JsonPointer` memory usage is quadratic over depth. `-Xmx8g` is insufficient to create a `JsonPointer` for [100k opening arrays](https://github.com/nst/JSONTestSuite/blob/d64aefb55228d9584d3e5b2433f720ea8fd00c82/test_parsing/n_structure_100000_opening_arrays.json) without an `OutOfMemoryError`.

`JsonPointer` seems to be a linked list where each node contains an `_asString` field containing the _full_ path to that node.

Minimal test to repro the issue: 3764ff4b1a8dc2090dec57da5dc0cc2c7f8f520b

```java
    // such as https://github.com/nst/JSONTestSuite/blob/master/test_parsing/n_structure_100000_opening_arrays.json
    public void testDeepJsonPointer() throws Exception {
        int DEPTH = 100000;
        String INPUT = new String(new char[DEPTH]).replace("\0", "[");
        JsonParser parser = createParser(MODE_READER, INPUT);
        try {
            while (true) {
                parser.nextToken();
            }
        } catch (Exception e) {
            JsonStreamContext parsingContext = parser.getParsingContext();
            JsonPointer jsonPointer = parsingContext.pathAsPointer(); // OOME
            String pointer = jsonPointer.toString();
            String expected = new String(new char[DEPTH - 1]).replace("\0", "/0");
            assertEquals(expected, pointer);
        }
    }
```

![heap dump filled with `JsonPointer` instances, each containing a String of the full path](https://user-images.githubusercontent.com/663139/146886688-f43fb676-88ca-4b68-863d-dc7318c1be1d.png)

cowtowncoder:
Interesting. I will have to look into this a bit. My immediate thinking is that this might be an unfortunate side effect of attempting to retain lazily computed description for `JsonPointer` for common (if wasteful) use cases where `JsonPointer.toString()` is called multiple times.

But it sounds like here it is more directly called as a side effect, causing every single pointer "node" to create and retain its full path description.

cowtowncoder:
Hmmmh. Unfortunately, looking at code involved, it is not the way I thought -- actual full path is eagerly constructed. This makes changing a whole lot more difficult and probably only doable for a new minor version (2.14).
[...]

PR #762, opened 2022-06-09
Propose that JsonPointer be Serializable

egalpin:
Noting that `Instances [of JsonPointer] are fully immutable and can be cached, shared between threads` it would be nice to be able to pass these between workers that require object serialization to pass instances around.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/JsonPointer.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonPointer.java
@@ -17,8 +17,9 @@ import com.fasterxml.jackson.core.io.NumberInput;
  *
  * @since 2.3
  */
-public class JsonPointer
+public class JsonPointer extends Serializable
 {
+    private static final long serialVersionUID = 1L; 
     /**
      * Character used to separate segments.
      *
--- a/src/main/java/com/fasterxml/jackson/core/JsonPointer.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonPointer.java
@@ -76,13 +76,12 @@ public class JsonPointer implements Serializable
     /* Construction
     /**********************************************************
      */
-    
+
     /**
      * Constructor used for creating "empty" instance, used to represent
-     * state that matches current node. Note that this constructor must be public for
-     * implementing {@link Externalizable}.
+     * state that matches current node.
      */
-    public JsonPointer() {
+    protected JsonPointer() {
         _nextSegment = null;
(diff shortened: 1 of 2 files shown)


cowtowncoder:
Seems reasonable, I like the idea. However, I wonder if this could be optimized to actually serialize as String, re-parse? (implement Externalizable)
I think this would be much more efficient, space-wise, than relying on JDK serialization.

egalpin:
@cowtowncoder good call.  I ran into an issue where class properties that need to be set via readExternal cannot be declared final.  I made 2 commits, each with a different approach to maintaining the `final` properties of JsonPointer as finals.  The 3rd option, not implemented here but very easy to do, is to change the 4 properties (`_nextSegment`, `_matchingPropertyName`, `_matchingElementIndex`, `_asString`) to not be final; I assumed that was less preferable but I'll leave that up to you.

cowtowncoder:
Ok thank you. One thing I also need to consider is that I hope implement fix for #736 which will probably change internal layout of things anyway.

src/main/java/com/fasterxml/jackson/core/JsonPointer.java
```diff
@@ -72,9 +79,10 @@
     
     /**
      * Constructor used for creating "empty" instance, used to represent
-     * state that matches current node.
+     * state that matches current node. Note that this constructor must be public for
+     * implementing {@link Externalizable}.
      */
-    protected JsonPointer() {
+    public JsonPointer() {
```
cowtowncoder, line 85: Does `Externalizable` really require that? TIL

egalpin:
> Does Externalizable really require that? TIL

It turns out it's not 😅 It requires a no-argument constructor, which is already present.

cowtowncoder:
Ok, this looks good. One last thing, CLA: unless you have already sent one, we'd need this:

(the usual way is to print, fill & sign, scan/photo, email to `info` at fasterxml dot com)

If you have already sent one already let me know (sometimes I forget as there are quite a few).

Once there's CLA I'll merge this. CLA works for any and all future contributions too so it's just a one-time hassle.

Looking forward to merging this PR!

egalpin:
Signed and sent 🙂

cowtowncoder:
Merged to be included in 2.14.0. Thank you very much for contributing this, @egalpin !

egalpin:
Thanks @cowtowncoder for the prompt and welcoming review! 🙏

PR #770, opened 2022-06-21
add Schubfach double/float writer

pjfanning:
add Schubfach based code to write floats/doubles - reduces size of #749

based on https://github.com/c4f7fcce9cb06515/Schubfach


pjfanning:
@cowtowncoder this PR should merge cleanly to 2.14 and master branches. I can look tomorrow at further splitting up the rest of #749

PR #771, opened 2022-06-21
add new java packages to module exports

pjfanning:
@cowtowncoder I'm not an expert on java module support but it seems likely that these changes are needed


cowtowncoder:
I think these would only be needed if we wanted other code to access these packages? Core itself can use them, but they are not exposed, and probably should not be?

pjfanning:
cool - I'll leave them out - if someone hits an issue, they can report it

Issue #707, opened 2021-07-09
JsonReadFeature: Introduce missing features for JSON5 compatibility


GrahamLea:
The optional features in the JSON parser allow it be configured to _almost_ be a fully-compatible parser for JSON5.

There's just a few missing features, documented here:
https://stackoverflow.com/questions/68312227/can-the-jackson-parser-be-used-to-parse-json5

Summarised as:
- hexadecimal numbers
- trailing decimal points on numbers
- plus signs before numbers
- extra whitespace characters

I propose that additional Jackson features be added for each of these JSON5 features to allow Jackson to be used as a fully-compliant JSON5 parser.

(#612 was a similar but more specific request, though it seems that feature is already covered)

cowtowncoder:
Sounds reasonable, as long as these are opt-in features (or, possibly, one JSON5 feature), to be added in `JsonReadFeature` and not defaults.
I probably won't have time to work on this in near future but as usual would be happy to help someone else has time for a PR or two!

cowtowncoder:
So, thanks to @pjfanning 's work, 2 number related features needed here are implemented to go in 2.14!

cowtowncoder:
Ok, so looks like we have 2 things missing:

* Hexadecimal numbers
* Extra whitespace 

Of these, hex numbers might be relatively easy to add (esp. for blocking parsers).
White space could be trickier to do efficiently but maybe not (esp. for "common" white space like tabs and other low-numeric code ones); also seems like less commonly needed feature.

cowtowncoder:
If anyone is interested, filing a separate issue for, say, Hexadecimal number support; adding unit tests (failing, under `.../failing/`) would be a way to get things started.

cowtowncoder:
Noting that if added, should be done for 3.x, not 2.x.

seonwooj0810:
@cowtowncoder I'd like to take this on for 3.x, starting with hexadecimal number support
only, per your 2023-05-30 suggestion to "file a separate issue for, say, Hexadecimal
number support" and your 2025-11-17 confirmation that this should target 3.x.

Plan I'd like to validate before opening a PR:

**Scope (this PR)**: Hexadecimal integer support only — `0x1A`, `0X1A`, optionally with
sign (`-0x1A`, `+0x1A` when `ALLOW_LEADING_PLUS_SIGN_FOR_NUMBERS` is also enabled).
Float and exponent forms are out of scope.

[...]

PR #774, opened 2022-06-22
Add a feature to allow leading plus sign (`JsonReadFeature.ALLOW_LEADING_PLUS_SIGN_FOR_NUMBERS`)

pjfanning:
Another of the 'missing' features mentioned in https://github.com/FasterXML/jackson-core/issues/707

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/json/ReaderBasedJsonParser.java
+++ b/src/main/java/com/fasterxml/jackson/core/json/ReaderBasedJsonParser.java
@@ -1479,6 +1479,9 @@ public class ReaderBasedJsonParser
         // First check: must have a digit to follow minus sign
         if (ch > INT_9 || ch < INT_0) {
             _inputPtr = ptr;
+            if (ch == INT_PERIOD) {
+                return _parseFloatThatStartsWithPeriod();
+            }
             return _handleInvalidNumberStart(ch, negative);
         }
         // One special case, leading zero(es):
--- a/src/test/java/com/fasterxml/jackson/core/read/NonStandardNumberParsingTest.java
+++ b/src/test/java/com/fasterxml/jackson/core/read/NonStandardNumberParsingTest.java
@@ -127,6 +127,12 @@ public class NonStandardNumberParsingTest
             assertEquals("0.125", p.getDecimalValue().toString());
             assertEquals("0.125", p.getText());
         }
+        try (JsonParser p = createParser(f, mode, " +.125 ")) {
+            assertEquals(JsonToken.VALUE_NUMBER_FLOAT, p.nextToken());
+            assertEquals(0.125, p.getValueAsDouble());
+            assertEquals("0.125", p.getDecimalValue().toString());
+            assertEquals(".125", p.getText());
+        }
     }
 
     private void _testTrailingDotInDecimalAllowed(JsonFactory f, int mode) throws Exception
--- a/src/main/java/com/fasterxml/jackson/core/json/UTF8DataInputJsonParser.java
+++ b/src/main/java/com/fasterxml/jackson/core/json/UTF8DataInputJsonParser.java
@@ -1099,6 +1099,8 @@ public class UTF8DataInputJsonParser
             // One special case: if first char is 0 need to check no leading zeroes
             if (c == INT_0) {
                 c = _handleLeadingZeroes();
+            } else if (c == INT_PERIOD) {
+                return _parseFloatThatStartsWithPeriod();
             } else {
                 return _handleInvalidNumberStart(c, negative);
             }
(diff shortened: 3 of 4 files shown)


src/test/java/com/fasterxml/jackson/core/read/NonStandardNumberParsingTest.java
```diff
@@ -31,6 +32,22 @@ public void testLeadingDotInDecimal() throws Exception {
...
+     * The format "+NNN" (as opposed to "NNN") is not valid JSON, so this should fail
+     */
+    public void testLeadingPlusSignInDecimal() throws Exception {
+        for (int mode : ALL_MODES) {
+            JsonParser p = createParser(mode, " +123 ");
+            try {
+                p.nextToken();
+                fail("Should not pass");
+            } catch (JsonParseException e) {
+                verifyException(e, "expected digit (0-9) to follow minus sign, for valid numeric value");
```
pjfanning, line 45: @cowtowncoder I've left this in draft because the existing code throws an exception with a confusing message if a plus sign precedes a number - is it ok if I use this PR to change the message to say that plus signs are not allowed in front of JSON numbers?

src/main/java/com/fasterxml/jackson/core/json/ReaderBasedJsonParser.java
```diff
@@ -740,6 +740,10 @@ public final JsonToken nextToken() throws IOException
 
         // Ok: we must have a value... what is it?
 
+        if (i == '+' && isEnabled(JsonReadFeature.ALLOW_LEADING_PLUS_SIGN_FOR_NUMBERS.mappedFeature())) {
+            return nextToken();
```
cowtowncoder, line 744: Ok, this is problematic as it allows something like `++++4` I think. Or, potentially, DoS with 10,000 plus signs. :)

But I think it should be possible to handle this as a new entry in `switch` below, maybe adding case right above `0` ... `9`, reading next character, falling through?
Same goes for all parsers.

src/test/java/com/fasterxml/jackson/core/read/NonStandardNumberParsingTest.java
cowtowncoder: Yes, absolutely, improvements to error messages are good.

cowtowncoder:
Ok, so, improvement(s) to exception message, +1.

But I think handling of leading positive marker should avoid use of recursion. It should be relatively straight-forward to instead add handling in `switch` block with `case '+':` preceding cases for digits `0`...`9`, reading next character (which will then not allow spaces between minus and number), requiring that to be valid number start character, then calling `_parsePosNumber()`.

The one thing that might be tricky is to ensure that `JsonParser.getText()` returns String with leading plus, it should not be trimmed off, ideally (at least I think intent is to represent exact token even if not JSON compliant).

Btw, great work I like that we might be able to support more use cases yet!

pjfanning:
@cowtowncoder I've refactored the code to stop the recursion (so multiple plus signs will not be valid). I've run into an issue with NonStandardParserFeaturesTest and its testAllowInfinity. The problem is that my PR has an explicit check for '+' and if the new parser feature is not enabled, the number is rejected. My check means '+INF' is not allowed unless the new parser feature is enabled.

I'm wondering about Infinity support. In NonStandardParserFeaturesTest, no feature is explicitly enabled to allow infinity - but my understanding is that infinity and NaN are not valid JSON values - if they are not valid, shouldn't they not be supported by default?

src/test/java/com/fasterxml/jackson/core/read/NonStandardNumberParsingTest.java
```diff
@@ -73,23 +90,58 @@ public void testTrailingDotInDecimalAllowedReader() throws Exception {
...
+            assertEquals(JsonToken.VALUE_NUMBER_INT, p.nextToken());
+            assertEquals(125.0, p.getValueAsDouble());
+            assertEquals("125", p.getDecimalValue().toString());
+            assertEquals("125", p.getText());
+        }
+        try (JsonParser p = createParser(f, mode, " +0.125 ")) {
+            assertEquals(JsonToken.VALUE_NUMBER_FLOAT, p.nextToken());
+            assertEquals(0.125, p.getValueAsDouble());
+            assertEquals("0.125", p.getDecimalValue().toString());
+            assertEquals("0.125", p.getText());
```
cowtowncoder, line 128: Ideally here we would get `+0.125`. But that's fine for now.

cowtowncoder:
Quick note: while things work wrt values, as far as I can see, the problem of `getText()` dropping `+` is trickier to resolve.

I think that it mostly/completely works for floating-point numbers but does not -- and unfortunately, cannot -- be easily resolved for integer numbers.
The problem is due to `TextBuffer` being used to store value, and hard-coded assumptions that only negative sign may exist (to be ignored on decoding). So while `+` may be prepended, it would then be used for calculations getting integer/long values badly off.

I'll have to think of a good way to resolve this; until that I may need to accept that plus sign will not be accessible as part of the text value for these cases. Not the worst thing in the world of course :)

PR #786, opened 2022-07-01
Add test to show hexadecimal not supported - but fix issue in UTF8DataInputJsonParser

pjfanning:
UTF8DataInputJsonParser did not behave like the other json parsers - added a change to make it give the same result as the others


cowtowncoder:
Sounds good, thanks!

cowtowncoder:
One other thing that would be good to (eventually) cover is non-blocking async case -- unfortunately as things are, tests are separate due to differences in input feeding.

However, maybe it would be possible to unify testing with two other modes (one where the whole document is fed as one chunk; second that feeds content byte-by-byte -- similar to streaming/throttling cases).
That's a bigger change but I realized that right now functionality/coverage is diverging a bit.
Then again, implementing features for async parser is quite different from blocking ones, generally more work.
Still, would be good to at least show differences with failing tests.

PR #791, opened 2022-07-10
Async pos sign

pjfanning:
The feature for allowing leading plus signs in numbers does not work with async JSON parser. This is my attempt at a fix.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/json/async/NonBlockingJsonParserBase.java
+++ b/src/main/java/com/fasterxml/jackson/core/json/async/NonBlockingJsonParserBase.java
@@ -99,6 +99,8 @@ public abstract class NonBlockingJsonParserBase
 
     protected final static int MINOR_VALUE_TOKEN_NON_STD = 19;
 
+    protected final static int MINOR_NUMBER_PLUS = 230;
+
     protected final static int MINOR_NUMBER_MINUS = 23;
     protected final static int MINOR_NUMBER_ZERO = 24; // zero as first, possibly trimming multiple
     protected final static int MINOR_NUMBER_MINUSZERO = 25; // "-0" (and possibly more zeroes) receive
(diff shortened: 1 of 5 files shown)


src/main/java/com/fasterxml/jackson/core/json/async/NonBlockingJsonParserBase.java
```diff
@@ -99,6 +99,8 @@
 
     protected final static int MINOR_VALUE_TOKEN_NON_STD = 19;
 
+    protected final static int MINOR_NUMBER_PLUS = 230;
```
cowtowncoder, line 102: How about use `22` instead since that's available?

src/test/java/com/fasterxml/jackson/core/json/async/AsyncNonStandardNumberParsingTest.java
```diff
@@ -144,8 +145,57 @@ public void testLeadingPlusSignInDecimalDefaultFail() throws Exception {
...
+    public void testLeadingPlusSignInDecimalDefaultFail2() throws Exception {
+        final String JSON = "[ +0.123 ]";
+
+        // without enabling, should get an exception
+        AsyncReaderWrapper p = createParser(DEFAULT_F, JSON, 1);
+        assertToken(JsonToken.START_ARRAY, p.nextToken());
+        try {
+            p.nextToken();
+            fail("Expected exception");
+        } catch (Exception e) {
```
cowtowncoder, line 163: Should probably catch more specific exception (`StreamReadException`?), but since all other tests for some reason also use `Exception` it's fine here too I suppose.

cowtowncoder:
Looks good, added couple minor comments.

About the only other question I was wondering was whether check for allowing leading plus could be done earlier, but if not that's fine.

src/main/java/com/fasterxml/jackson/core/json/async/NonBlockingJsonParserBase.java
pjfanning: changed

src/test/java/com/fasterxml/jackson/core/json/async/AsyncNonStandardNumberParsingTest.java
pjfanning: changed

pjfanning:
@cowtowncoder I'm not sure it easy to move the + sign checks - I tried moving them earlier in the flow but it affects support for the already supported `+Infinity` value

cowtowncoder:
Ok, looks good, thanks!

PR #795, opened 2022-07-12
add NonBlockingByteBufferJsonParser (and refactor some NonBlockingJsonParser code to make it more extensible)

pjfanning:
Relates to https://github.com/FasterXML/jackson-core/issues/478

* make it feasible to have a separate ByteBuffer based parser - it's a pity to see #483 die on the vine - but I would favour having separate byte[] and ByteBuffer parsers to avoid the overhead of wrapping the byte[] in buffers
* aim would be to move on most NonBlockingJsonParser code to a superclass and that a new NonBlockingByteBufferJsonParser would subclass this superclass and just override some methods like the new getNextByteFromBuffer and getByteFromBuffer methods (would NonBlockingJsonParserBase be ok for this or would a new intermediate class be better?)
* I have run some perf tests and the refactored NonBlockingJsonParser seems to have almost exactly the same perf as before (so far, it seems a tad faster)


src/main/java/com/fasterxml/jackson/core/json/async/NonBlockingJsonParser.java
```diff
@@ -194,7 +194,7 @@ public JsonToken nextToken() throws IOException
         _tokenInputTotal = _currInputProcessed + _inputPtr;
         // also: clear any data retained so far
         _binaryValue = null;
-        int ch = _inputBuffer[_inputPtr++] & 0xFF;
+        int ch = getNextByteFromBuffer() & 0xFF;
```
cowtowncoder, line 197: If we make this change, might as well add that `& 0xFF` part in accessor.
We can document this in Javadoc (requirement to return "unsigned" byte); plus it's only implemented once per backend.

src/main/java/com/fasterxml/jackson/core/json/async/NonBlockingJsonParser.java
```diff
@@ -1272,7 +1290,7 @@ protected JsonToken _finishNonStdTokenWithEOF(int type, int matched) throws IOEx
     protected JsonToken _finishErrorToken() throws IOException
     {
         while (_inputPtr < _inputEnd) {
-            int i = (int) _inputBuffer[_inputPtr++];
+            int i = (int) getNextByteFromBuffer();
```
cowtowncoder, line 1293: ... and if that `& 0xFF` is included, we'd likely require alternative of `getNextSignedByteFromBuffer()` or such, to be used here.

cowtowncoder:
Yeah, I am specifically worried about performance aspects since byte access is, well, once (or more) per input byte.
If this can be shown to have only marginal effect with modern JVMs (say, <10%) I am ok with it.
I don't know if there are any specific tricks we can use to help JVM speculatively inline access for single implementation type.

Alternatively if we could find out hot spots -- maybe a small portion of accesses could be left for concrete classes to implement -- we could perhaps have hybrid model to balance need to re-implement only a small set of methods (with hard-coded access). I haven't checked code to see if that is at all feasible, will have a look next.

It definitely would be nice to support `ByteBuffer`, as well as potentially other library/framework-provided byte-buffer/byte-source abstractions.

pjfanning:
@cowtowncoder this still far from complete and needs testing and more specifically, perf testing.

I have pushed on and created NonBlockingUtf8JsonParserBase which has most of the code from NonBlockingJsonParser. NonBlockingByteBufferJsonParser extends NonBlockingUtf8JsonParserBase.

The idea is to have NonBlockingJsonParser also extend NonBlockingUtf8JsonParserBase but I've left that for a few days so that it is visible how much code has changed in NonBlockingJsonParser in order to not have the methods work directly on the `_inputBuffer` (and to use the getters instead).

pjfanning:
@cowtowncoder does this look a reasonable perf test? 

https://github.com/pjfanning/jackson-number-parse-bench/blob/main/src/jmh/java/org/example/jackson/bench/AsyncParserBench.java

I need to try it with more permutations but on Zulu DK 1.8.0_332, I get these results:

Without refactor:
AsyncParserBench.bigArrayOfFloatPrimitives  thrpt    5  26115.264 ± 626.839  ops/s

With refactor:
AsyncParserBench.bigArrayOfFloatPrimitives  thrpt    5  26964.013 ± 201.662  ops/s

So far, it looks like the refactor has not has a noticeable impact on performance - at least in this test case.

cowtowncoder:
@pjfanning Thanks, I need to have a look. It sounds legit, although ideally test would exercise different token types and access content. But then again, there are so many different aspects (in most tests the whole content is fed as a big chunk, vs. smaller pieces).

But I think at least we can rule out obvious drastic negative effects.

Fwtw, jackson-benchmarks:

also has non-blocking/async test variants so maybe I could find time to test it as well.

pjfanning:
Thanks for pointing me towards jackson-benchmarks. I ran one of the existing benchmarks and got these numbers (Zulu JDK 1.8.0_332):

2.13.3
Benchmark                            Mode  Cnt       Score      Error  Units
JsonStdReadAsync.readPojoMediaItem  thrpt    8  383982.674 ± 5241.035  ops/s

2.14 without refactor
Benchmark                            Mode  Cnt       Score      Error  Units
JsonStdReadAsync.readPojoMediaItem  thrpt    8  386167.559 ± 6744.119  ops/s

2.14 with refactor
Benchmark                            Mode  Cnt       Score      Error  Units
JsonStdReadAsync.readPojoMediaItem  thrpt    8  402986.292 ± 5160.014  ops/s

src/main/java/com/fasterxml/jackson/core/json/async/NonBlockingJsonParser.java
pjfanning: done

src/main/java/com/fasterxml/jackson/core/json/async/NonBlockingJsonParser.java
pjfanning: done

cowtowncoder:
I will try to get this reviewed and merged soon (this week has been hectic but next week I'll have much more time for Jackson again!)

cowtowncoder (approved):

cowtowncoder:
@pjfanning Ok, I probably should have merged PRs in different order, but looks like there's a conflict. Could you resolve that? I'll merge this ASAP when that is done.

pjfanning:
@cowtowncoder thanks for merging the other PRs - the async number parsing test one was always going to conflict with this - I think I've got those changes merged in here now to fix the conflict.

cowtowncoder:
Thank you @pjfanning ! Yeah, minor conflicts are to be expected. Will now merge.

cowtowncoder:
Ugh. This will require some care merging to 3.0 :)

cowtowncoder:
Took a while but merging successful. :)

PR #798, opened 2022-07-23
Avoid copy when parsing BigDecimal

marschall:
Avoid a char[] copy when parsing a BigDecimal reducing intermediate
allocation.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/io/BigDecimalParser.java
+++ b/src/main/java/com/fasterxml/jackson/core/io/BigDecimalParser.java
@@ -33,18 +33,11 @@ public final class BigDecimalParser
     }
 
     public static BigDecimal parse(char[] chars, int off, int len) {
-        if (off > 0 || len != chars.length) {
-            chars = Arrays.copyOfRange(chars, off, off+len);
-        }
-        return parse(chars);
-    }
-
-    public static BigDecimal parse(char[] chars) {
-        final int len = chars.length;
         try {
             if (len < 500) {
-                return new BigDecimal(chars);
+                return new BigDecimal(chars, off, len);
             }
+            chars = Arrays.copyOfRange(chars, off, off+len);
             return new BigDecimalParser(chars).parseBigDecimal(len / 10);
         } catch (NumberFormatException e) {
             String desc = e.getMessage();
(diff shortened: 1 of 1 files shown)


marschall:
While profiling some Jackson code I noted that in order to create a `BigDecimal` we create a copy of the `char[]`. There is no need for this, `BigDecimal` has a `char[],int,int` constructor since Java 5. In the following micro benchmark I get about 20% throughput improvement.

```java
public class JacksonStdBigDecimalReadVanilla
{

    private static final JsonFactory JSON_FACTORY = new JsonFactory();
    
    private static final String JSON;

    static {
        StringBuilder json = new StringBuilder();
        json.append('[');
        for (int i = 0; i < 1_000; i++) {
            if (i > 0) {
                json.append(',');
            }
            json.append("123.456");
        }
        json.append(']');
        JSON = json.toString();
    }

    @Benchmark
    public void parseBigDecimal(Blackhole blackhole) throws IOException {
        try (JsonParser parser = JSON_FACTORY.createParser(JSON)) {
            parser.nextToken(); // start array
            
            JsonToken nextToken = parser.nextToken();
            while (nextToken == JsonToken.VALUE_NUMBER_FLOAT) {
                blackhole.consume(parser.getDecimalValue());
                nextToken = parser.nextToken();
            }
        }
    }

}

```

cowtowncoder:
Thank you for contributing this patch @marschall -- interesting that we had overlooked existing of that constructor, good catch! I'd be happy to merge this, and the only thing I need (if not already done earlier) is to get CLA from:

it's only needed once before the first Jackson contribution, good for all future contributions (hence if you have sent one earlier LMK and I can verify).
The usual way is to print, fill & sign, scan/photo, email to `info` at fasterxml dot com.
Once I get that, I'll merge this and it gets into upcoming 2.14.0.

Thank you again very much for providing this -- performance improvements are always welcome!

marschall:
@cowtowncoder awesome, give me one or two days

marschall:
@cowtowncoder I sent the CLA, let me know if you need anything more. Should I also submit a PR against [FasterXML/jackson-benchmarks](https://github.com/FasterXML/jackson-benchmarks)?

cowtowncoder (approved):

cowtowncoder:
CLA recieved, will merge.

src/main/java/com/fasterxml/jackson/core/io/BigDecimalParser.java
```diff
@@ -33,18 +33,11 @@ public static BigDecimal parse(String valueStr) {
...
-    }
-
-    public static BigDecimal parse(char[] chars) {
-        final int len = chars.length;
         try {
             if (len < 500) {
-                return new BigDecimal(chars);
+                return new BigDecimal(chars, off, len);
             }
+            chars = Arrays.copyOfRange(chars, off, off+len);
```
pjfanning, line 40: Would it make more sense to avoid this Arrays.copyOfRange and change the BigDecimalParser constructor to take (chars, off, len). It looks like the parseBigDecimal can be readily changed to support an offset and explicit len.

cowtowncoder: I can make that, thanks!

cowtowncoder: That could make sense if easy enough to do.

pjfanning: https://github.com/FasterXML/jackson-core/pull/801

cowtowncoder:
@marschall Merged. As to `jackson-benchmarks` not sure; this might be little bit too specific to be tested.

PR #799, opened 2022-07-24
add async parser tests for leading dots (and make changes to get them to pass)

pjfanning:
I've only found a fix partial for the 2 leading dot tests in this PR. I also added a couple of unrelated ones that pass.

The previously failing ones are for +.123 and -.123 -- the combination of a leading sign and a leading dot did not work.

The current 'fix' could do with more work.


cowtowncoder:
Excellent, thank you!

PR #800, opened 2022-07-24
removed unused param on _parseFloatThatStartsWithPeriod

pjfanning:
This version of the _parseFloatThatStartsWithPeriod method on ReaderBasedJsonParser was only introduced in 2.14 branch recently. Seems safe to remove the unused param from the method signature.


cowtowncoder:
Sounds reasonable, thanks!

Issue #768, opened 2022-06-16
Jackson bundles are missing OSGi's osgi.serviceloader metadata


chrisr3:
Bundles need to include extra metadata to support `java.util.ServiceLoader` inside an OSGi framework. Jackson's bundles are all missing this metadata.

In theory, it _should_ just be a matter of applying Bnd's `@ServiceConsumer` and `@ServiceProvider` annotations in the correct places as described [here](https://bnd.bndtools.org/chapters/240-spi-annotations.html), where these annotations can be provided like this:
```diff
diff --git a/base/pom.xml b/base/pom.xml
index 6235a67..4972957 100644
--- a/base/pom.xml
+++ b/base/pom.xml
@@ -30,9 +30,17 @@ of Jackson: application code should only rely on `jackson-bom`
          parent pom, and then also allow override as need be
       -->
     <jackson-bom.version>${project.parent.version}</jackson-bom.version>
+
+    <version.bnd.annotation>6.3.1</version.bnd.annotation>
   </properties>
 
   <dependencies>
+    <dependency>
+      <groupId>biz.aQute.bnd</groupId>
+      <artifactId>biz.aQute.bnd.annotation</artifactId>
+      <version>${version.bnd.annotation}</version>
+      <scope>provided</scope>
+    </dependency>
     <dependency> <!-- all components use junit for testing -->
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
```
I am struggling to wrap my head around your myriad of separate repositories, and so you may have a better way of providing this dependency.

cowtowncoder:
Being bit of OSGi newbie, quick question: is this really needed for `jackson-core` which does not provide Service of any type -- meaning there is no life-cycle or such. It is simply a library with visibility as defined by exports (and theoretically imports but in practice having few).

So what would be the meaning of `ServiceLoader` for this particular component? (or for others, `jackson-databind`)
It sounds like there is a problem to solve, I just want to understand what that is.

chrisr3:
> So what would be the meaning of `ServiceLoader` for this particular component? (or for others, `jackson-databind`) It sounds like there is a problem to solve, I just want to understand what that is.

[...]

PR #805, opened 2022-08-01
Add minimal OSGi metadata as per #768

cowtowncoder:
So, as per #768 we may want to add minimal OSGi metadata here, and eventually across all format backends plus probably `jackson-databind`. But starting with basics.

Also note that if this works, 3.0 (`master`) uses bit different bindings (`TokenStreamFactory` that `JsonFactory` implements) -- but for 2.x we'll stick to `JsonFactory` as being the base class every format backend implements.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -38,7 +38,6 @@
        still use Moditect to get JDK9+ module info support; need newer bundle plugin as well
        (can just defaults from `jackson-parent`)
       -->
-
     <osgi.export>com.fasterxml.jackson.core;version=${project.version},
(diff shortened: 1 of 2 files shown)


cowtowncoder:
@chrisr3 Is this what you meant in #768? Not sure how to test goodness (or lack thereof).

chrisr3 (commented):
Yes, I think that looks right. AFAIK this is the only `jackson-core` class with a `META-INF/services` entry, and it doesn't explicitly use `ServiceLoader.load()` so this should be the only annotation that needs adding.

PR #806, opened 2022-08-06
uptake latest fast double parser code

pjfanning:
Latest code from https://github.com/wrandelshofer/FastDoubleParser


cowtowncoder:
@pjfanning Ok. Merged to 2.14, but unable to merge 2.14->master, unfortunately. Some of it is due to package rename, but there is something else. So I think I need help here. :-(

Issue #809, opened 2022-08-12
Bounds-checks issues with Fast FP parser (placeholder)


cowtowncoder:
(note: this is a placeholder until individual issues are created)

It looks like the new "fast FP parsing" functionality has a few issues uncoved by OssFuzz (https://oss-fuzz.com/) `jackson-core` project. This is not unsurprising as fuzzers are good at finding various edge cases for invalid input.
I will need to figure out how to give access to actual reports; in the meantime, this issue can act as a placeholder for various instances.

cowtowncoder:
@pjfanning I am hoping to give you access to OssFuzz jackson projects -- I don't maintain them (nor written code) but collaborate with author(s). Full information is not by default world-accessible, due to possible security issues as some problems could be considered Vulnerabilities.

The first issue (I think there are more, but first I saw) is reported for invalid input, with stack trace of:

```
== Java Exception: java.lang.StringIndexOutOfBoundsException: String index out of range: 3
--
  | at java.base/java.lang.StringLatin1.charAt(StringLatin1.java:48)
  | at java.base/java.lang.String.charAt(String.java:712)
  | at com.fasterxml.jackson.core.io.doubleparser.AbstractFloatingPointBitsFromCharSequence.parseHexFloatLiteral(AbstractFloatingPointBitsFromCharSequence.java:331)
  | at com.fasterxml.jackson.core.io.doubleparser.AbstractFloatingPointBitsFromCharSequence.parseFloatingPointLiteral(AbstractFloatingPointBitsFromCharSequence.java:215)

```

and oss-fuzz testcase has actual input data (which we'll need to repro and perhaps request a fix or do the fix).
It seems like a missing bounds-check of some kind; I have fixed similar ones from jackson-core multiple times.

pjfanning:
@wrandelshofer AbstractFloatingPointBitsFromCharSequence is copied from your FastDoubleParser project.

@cowtowncoder Jackson doesn't really support hex numbers anyway so it is unlikely that Jackson parsers would run into this issue. Users could use NumberInput class directly but this not really a supported use case. I'd still like to get this fixed - I'm just highlighting that this is relatively low impact for Jackson users.

cowtowncoder:
@pjfanning I could be wrong but I don't think Fuzzer calls these methods directly. So it would seem like code somehow managed to trigger execution. Most likely it'd be root-level JSON value that starts with something like `1.0` or `-0.` that would trigger code path in question.

[...]

PR #810, opened 2022-08-12
fix for issue found by OSS Fuzz

pjfanning:
Relates to https://github.com/wrandelshofer/FastDoubleParser/pull/21 and https://github.com/FasterXML/jackson-core/issues/809

There is a fuller fix scheduled in https://github.com/wrandelshofer/FastDoubleParser/issues/22 - which can be copied here. This existing PR may still be worth using as a partial fix.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/io/doubleparser/AbstractFloatingPointBitsFromCharArray.java
+++ b/src/main/java/com/fasterxml/jackson/core/io/doubleparser/AbstractFloatingPointBitsFromCharArray.java
@@ -338,7 +338,7 @@ abstract class AbstractFloatingPointBitsFromCharArray extends AbstractFloatValue
         // ------------------------
         index = skipWhitespace(str, index, endIndex);
         if (illegal || index < endIndex
-                || digitCount == 0 && str[virtualIndexOfPoint] != '.'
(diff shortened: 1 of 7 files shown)


wrandelshofer:
Please note that this pull request does not fix the problem in all affected classes. You need to do the same changes also in class com.fasterxml.jackson.core.io.doubleparser.AbstractFloatingPointBitsFromCharArray.

Please look at the classes in the https://github.com/wrandelshofer/FastDoubleParser repository. The main branch, from revision 69cdb361224b10ffdd009bd7aadc8c8a0dacf841 onwards, contains code that fixes this problem, and runs optimal on Java 17.

pjfanning:
Thanks @wrandelshofer - I have copied over your latest code into this PR. Thanks for the really fast turnaround.

cowtowncoder:
I was unable to create a reproduction from oss-fuzz test case (input is... weird). But this makes sense anyway so will merge and we'll see if the issue gets resolved.

cowtowncoder:
Was able to merge 2.14 -> master; was bit messy due to package renaming but I think I got it right; all tests passing.

cowtowncoder:
Good news: oss-fuzz issue seems resolved!

PR #817, opened 2022-09-23
get ParserMinimalBase constructor to call super to initiate its super class

pjfanning:
seems more correct because the other ParserMinimalBase constructor calls super(...)


cowtowncoder:
Ok, sure. My understanding is that without it the behavior is the same (no-args supertype constructor is called) but I guess there's no harm in being explicit.

Issue #717, opened 2021-09-16
Hex capitalization for JsonWriter should be configurable


pakoito:
We're seeing inconsistencies between JS's and Jackson's representation of non-printable characters embedded in JSON strings. Where JS sends `\u001b`, once it goes through Jackson the field holds the string `\u001B`, causing all kind of hashing issues.

In https://github.com/FasterXML/jackson-core/blob/efe51e83648ed1b4c33ad1ffd5e2cbee5e0c6091/src/main/java/com/fasterxml/jackson/core/io/CharTypes.java#L7 we find that the hex values are hardcoded as capital `A-F`, whereas other representations such as JS's `JSON.stringify` indicate you should use lowercase `a-f` instead.

These are used in the standard JsonWriter via https://github.com/FasterXML/jackson-core/blob/94a0e1cab04a218cb189e635727ca803dc56a958/src/main/java/com/fasterxml/jackson/core/JsonStreamContext.java#L331

We would like to see a configuration parameter or recommended approach to overcome this issue.

Seen in 2.13, made from before 2.0

cowtowncoder:
This sounds like a good idea, an addition to `JsonWriteFeature` enums (not being universal one, although some other formats may need similar handling).
I wish I had time to work on this; if anyone has time I'd be happy to help but do not have time as is.
2.13.0 is being finalized as well so timing may not work for inclusion there, unless PR was ready within couple of days.

pakoito:
The implementation seems straightforward but I am not familiar enough with the repo to understand how to drill down this configuration, or write the tests. If you could give me an intro I could spend some time on it.

cowtowncoder:
Might be good idea to look into how other configuration settings are accessed. Or where helper method(s) for encoding are called by generators.

cowtowncoder:
On API choice: could either have a simple `JsonWriteFeature` with 2 choices (upper- / lower-case; defaulting to upper-case) -- like `JsonWriteFeature.WRITE_HEX_UPPER_CASE(true)` -- or something more advanced. I am guessing simple choice might work well enough.

Tagging as "good first issue" as this seems reasonable straight-forward (not necessarily super easy of course, just less complicated than many other open issues).

pakoito:
I'm okay with the simpler option. Sadly I'm not at that employer anymore, so tagging @isaacasensio

isaacasensio:
👋 I can take this one. Thanks for the heads up @pakoito :)

[...]

PR #819, opened 2022-10-03
add cfg for hex values if we want upper or lowercase letters

Richie94:
For issue https://github.com/FasterXML/jackson-core/issues/717 I tried to make this configurable.

Its my first PR here, I didnt found a good spot for testing, would be nice if you can point me out where to do this.


src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
```diff
@@ -250,8 +250,21 @@ public enum Feature {
...
+        /**
+         * Feature that specifies that hex values are encoded with capital letters.
+         *<p>
+         * Can be disabled to have a better possibility to compare between other Json
+         * writer libraries, such as JSON.stringify from Javascript.
+         *<p>
+         * Feature is enabled by default.
+         *
+         *
+         * @deprecated Since 2.10 use {@link com.fasterxml.jackson.core.json.JsonWriteFeature#WRITE_HEX_UPPER_CASE} instead
```
cowtowncoder, line 264: 2.10 -> 2.14 since it'll  be added then.

src/main/java/com/fasterxml/jackson/core/io/CharTypes.java
```diff
@@ -284,12 +288,12 @@ public static void appendQuoted(StringBuilder sb, String content)
         }
     }
 
-    public static char[] copyHexChars() {
```
cowtowncoder, line 287: Would require retaining the old method as deprecated, otherwise this is breaking change. Same for the next method.

cowtowncoder:
Looks mostly good, thank you for contributing this!

As to testing, I think it should go somewhere under `com.fasterxml.jackson.core.write`, perhaps `GeneratorBasicTest`, or new test class. Should cover all implementations.

Aside from that, one thing I will need before merging (and about only thing) is CLA.
It can be found here:

(or Corporate CCLA next to it).

the usual way is to print, fill & sign, scan, email to `info` at fasterxml dot com.
With that I can merge this and it can still make it in 2.14.0 release (or -rc2 first maybe).

src/main/java/com/fasterxml/jackson/core/json/JsonWriteFeature.java
```diff
@@ -72,6 +72,17 @@
...
     ESCAPE_NON_ASCII(false, JsonGenerator.Feature.ESCAPE_NON_ASCII),
 
+    /**
+     * Feature that specifies that hex values are encoded with capital letters.
+     *<p>
+     * Can be disabled to have a better possibility to compare between other Json
+     * writer libraries, such as JSON.stringify from Javascript.
+     *<p>
+     * Feature is enabled by default.
+     */
```
cowtowncoder, line 82: Add `@since 2.14)

src/main/java/com/fasterxml/jackson/core/JsonStreamContext.java
```diff
@@ -303,6 +303,11 @@ public JsonLocation startLocation(ContentReference srcRef) {
     public JsonLocation getStartLocation(Object srcRef) {
         return JsonLocation.NA;
     }
+
+    /**
+     * Uppercase Hex letters
+     */
+    public boolean uppercaseHex() { return true; }
```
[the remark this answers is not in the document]
Richie94: Was from some first mistrial, i got rid of it.

src/main/java/com/fasterxml/jackson/core/io/CharTypes.java
Richie94: I kept the old ones as defaulting to uppercase/true

Richie94:
Thx for fast feedback, ill print it out tomorrow and send it to you.

src/test/java/com/fasterxml/jackson/core/write/UTF8GeneratorTest.java
```diff
@@ -114,4 +116,37 @@ public void testFilteringWithEscapedChars() throws Exception
         assertNull(p.nextToken());
         p.close();
     }
+
+    public void testHexLowercase() throws Exception
+    {
+        ByteArrayOutputStream bytes = new ByteArrayOutputStream();
+        IOContext ioc = new IOContext(new BufferRecycler(),
+                ContentReference.rawReference(bytes), true);
+        JsonGenerator gen = new UTF8JsonGenerator(ioc, 0, null, bytes, '"');
```
cowtowncoder, line 125: Would it be possible to change this and other tests to use `JsonFactory.builder().disable(...)...build()` approach instead? The reason is that this makes merge to `master` (Jackson 3.0) easier: there will no longer be mutability of generators.

Or maybe I can change this after merging PR if you want, either is fine.

cowtowncoder (approved):

Richie94:
Okay, Mail should be in your inbox.

cowtowncoder (approved):

cowtowncoder:
CLA received, can merge.

PR #821, opened 2022-10-17
fuzzing: run fuzzers in CI

AdamKorcz:
This sets up CiFuzz for Jackson-core which will run the fuzzers in the CI for 20 minutes when a pull request is made. The ammount of time can be modified if need be.

Signed-off-by: AdamKorcz <adam@adalogics.com>


cowtowncoder:
Ok seems to need minor tweaking. Happy to merge once that's fixed, to see how this might work.

PR #822, opened 2022-10-17
Declare osgi.serviceloader.registrar requirement as optional.

chrisr3:
As with Woodstox [PR#155](https://github.com/FasterXML/woodstox/pull/155), do not require that the OSGi framework also contains Apache SPI-Fly.


cowtowncoder:
First of all, thank you @chrisr3 for submitting this.

Second: I really wish we had gotten this in 2.14.0-rc2 to get some testing. My spider sense is tingling wrt Android users and others soon reporting a problem with "missing" annotation (some tool somewhere freaking out where it really shouldn't).

But I think it is probably best to merge this in, still.

cowtowncoder:
Or, maybe there's no problem since we already have an annotation. We'll see :)

cowtowncoder:
Reverted along with #768 due to #824

Issue #814, opened 2022-09-21
Use `BigDecimalParser` for BigInteger parsing very long numbers


pjfanning:
Maybe something for jackson 2.15. Relates to #813

String to BigInteger support is currently done using `new BigInteger(str)`.

BigDecimalParser has some performance related code for handling numbers with many digits. #813 suggests that we might be able to get improvements in performance and error handling by using some ideas from jsoniter_scala. BigInteger could probably benefit from these too.

@plokhotnyuk fyi

pjfanning:
@cowtowncoder could we consider #826 as a late breaking change for Jackson 2.14.0?

cowtowncoder:
@pjfanning Considering everything else, I think that's safe enough so merged PR to get this improvement in.
I do need to stop merging things of course, but this seemed like non-problematic thing even if last minute.

PR #826, opened 2022-10-21
improve performance with parsing bigints with lots of digits

pjfanning:
Parsing numbers in Java has poor performance characteristics. Parse time increases exponentially as the number of digits increases.
Using BigDecimalParser to parse BigIntegers has better performance when there are a lot of digits. It is slower for more usual numbers. From benchmarking, BigDecimalParser is more performant when you have approx 1250 digits and as you head to 10k, 100k digits, it is much faster. BigDecimalParser is not O(n) but it is better than `new BigInteger(string)`. 

Relates to #814

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/io/BigDecimalParser.java
+++ b/src/main/java/com/fasterxml/jackson/core/io/BigDecimalParser.java
@@ -37,7 +37,7 @@ public final class BigDecimalParser
             return parseBigDecimal(chars, off, len, len / 10);
 
         // 20-Aug-2022, tatu: Although "new BigDecimal(...)" only throws NumberFormatException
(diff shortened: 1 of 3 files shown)


cowtowncoder (approved):

cowtowncoder:
Interesting! Ok, I trust that you have benchmarked this in addition to it being found a useful optimization in other libraries.
Will merge to go on 2.14.0 since it seems like a low-risk improvement.

pjfanning:
@cowtowncoder I benchmarked using https://github.com/pjfanning/jackson-number-parse-bench

cowtowncoder:
@pjfanning Cool. Would it make sense to add simple results on README (everyone can of course run it locally too).

PR #827, opened 2022-10-22
Num size limit

pjfanning:
* introduces a default max length for numbers of 1000 chars - this default can be overridden (eg set to a much higher limit)
* we shouldn't allow super large numbers as a default (I can justify this in private exchanges but don't want to spell it out in a public forum)
* checks the string/char[] length before parsing the text as a number
* this is a bit late but better than not checking at all - because parsing numbers is expensive and slower than O(n)
* the problem with doing an earlier check (while building the char[] is that is not always clear how the TextBuffer is going to be used. Is it being used to get text or is it being used to buffer up the chars needed to parse a number
  * if someone can point out how we can know in advance if the TextBuffer is destined to used for number parsing, then the num size limit can be checked earlier 

To create a JSONFactory that allows larger numbers, you can use this code
```
        JsonFactory f = JsonFactory.builder()
                .streamReadConstraints(StreamReadConstraints.builder().withMaxNumberLength(10000).build())
                .build();
```


src/main/java/com/fasterxml/jackson/core/JsonParser.java
```diff
@@ -437,6 +439,26 @@ protected JsonParser() {
      */
     public abstract void setCodec(ObjectCodec oc);
 
+    /**
+     * Set the max number of characters that are allowed when parsing numbers.
+     *
+     * @param maxNumLen max number of characters that are allowed
+     * @since 2.14
+     */
+    public void setMaxNumLen(int maxNumLen) {
```
cowtowncoder, line 448: Quick comment: I would want to avoid setters (and probably getters too) since 3.0 has parser and generators fully immutable. Also settings should probably be passed in a separate config object to reduce number of methods/parameters being passed: something like `StreamReadConfig` / `StreamWriteConfig` (maybe), sub-classable on per-format basis. Configured using builder()-style approach; starting with default settings.

cowtowncoder: Hmmh. Ok, first things first: I would suggest going via 2.15, not `master`, due to package renaming. It'll probably be tricky either way, but I don't think master->2.15 is any easier than the other way around.

But be that as it may, bigger question is that of passing config objects. My first gut reaction is that making these go through `IOContext` is not clean. But the thing about `TextBuffer` might be relevant -- passing to `JsonParser` / `JsonGenerator` I wouldn't worry as much about (factories have been changed a few times and while unfortunate it's internal interface).
I wouldn't be totally against passing it via `IOContext` but would prefer alternative.

One thing I do not remember, which may be relevant, is whether format backends sub-class `IOContext` or not; I have a vague recollection might do, for recycling or such.

pjfanning: I've introduced a StreamReadConfig - and it can only be set via the TSFBuilder.

Internally, I still haven't found a better way to pass around the StreamReadConfig than using the IOContext.

pjfanning:
@cowtowncoder would you be able to re-review this when you get a chance?

cowtowncoder:
Yes @pjfanning I will try to get back to this as soon as possible.

src/main/java/com/fasterxml/jackson/core/StreamReadConfig.java
```diff
@@ -0,0 +1,28 @@
...
+
+    private int _maxNumLen = DEFAULT_MAX_NUM_LEN;
+
+    /**
+     * Sets the maximum number length (in chars). The default is 1000 (since Jackson 2.14)
+     * @param maxNumLen the maximum number length (in chars)
+     * @return this config
+     * @since 2.15
+     */
+    public StreamReadConfig maxNumberLength(int maxNumLen) {
```
cowtowncoder, line 20: I think these must be immutable, so changes via `withXxx()` methods, or using full builders. Mutable instances are not thread-safe and would require more copying.

src/main/java/com/fasterxml/jackson/core/StreamReadConfig.java
```diff
@@ -0,0 +1,28 @@
+package com.fasterxml.jackson.core;
+
+/**
+ * The configuration for streaming reads.
+ *
+ * @since 2.15
+ */
+public class StreamReadConfig {
```
cowtowncoder, line 8: Probably should have mentioned earlier but since there are limits/constraints, should name the container to indicate that.

src/main/java/com/fasterxml/jackson/core/StreamReadConfig.java
```diff
@@ -0,0 +1,28 @@
...
+
+    private static final int DEFAULT_MAX_NUM_LEN = 1000;
+
+    private int _maxNumLen = DEFAULT_MAX_NUM_LEN;
```
cowtowncoder, line 12: Need to also consider the fact that as-is this would be specific to textual formats. Binary formats would either:

1. Interpret this as byte limit instead of char (which is probably fine), or
2. use different setting

Either way would need to be documented once we figure out way to go.

pjfanning: currently all the existing code takes Strings and parses Strings - everything has been converted to text

src/main/java/com/fasterxml/jackson/core/StreamReadConfig.java
pjfanning: could you suggest a name - you suggested StreamReadConfig and I used that name based on your suggestion

cowtowncoder: Sorry about that.

Maybe `StreamReadLimits` or `StreamReadConstraints`?

src/main/java/com/fasterxml/jackson/core/StreamReadConfig.java
cowtowncoder: I understand this is the case for default handling, and in 2.x. Just noting that these checks would not be applied by, say, CBOR or Smile parsers as they use binary encoding of numbers and do not convert to `String`s in most cases.

But I think it'd make sense to consider length in "units of input", which can then be chars or bytes as appropriate.

src/main/java/com/fasterxml/jackson/core/JsonParser.java
cowtowncoder: Apologies for not thinking this through. As per my CR comments, I think it should be `StreamReadLimits` (or perhaps `StreamReadConstraints`).

But one thing that is also needed is to make sure values are fully immutable so they can be shared by parser instances etc. Changes could be done using `wihtMaxNumberLength()` or such, but that may become unmaintainable when adding values. So perhaps it is necessary to start with builder style construction immediately.

However.... I guess I am ok without doing that right away so we could merge things in sooner.

So it is ok not to add builder as the first thing.

Otherwise I like this; passing limits to/via `IOContext` makes sense, I think.

pjfanning:
@cowtowncoder I think I've addressed most of your issues with some recent commits. I've updated the PR description with how the code is called. The class is now StreamReadConstraints and has a builder().

cowtowncoder:
Ok couple of follow-up question on limits:

1. Is it reasonable to have same length limit for integral and floating-point numbers? Latter are (afaik) much more expensive to handle -- and former might have more legit use cases (for encryption keys, I think?). Or should those be split into 2 settings
2. Is `1000` bit low as a limit? I don't know if there are use case past that, or perhaps limits other parsers use. I know it's to some degree arbitrary, but since we are adding a new constraint it'd be nice to reduce chance of breaking existing use cases

pjfanning:
> Ok couple of follow-up question on limits:
> 
> 1. Is it reasonable to have same length limit for integral and floating-point numbers? Latter are (afaik) much more expensive to handle -- and former might have more legit use cases (for encryption keys, I think?). Or should those be split into 2 settings
> 2. Is `1000` bit low as a limit? I don't know if there are use case past that, or perhaps limits other parsers use. I know it's to some degree arbitrary, but since we are adding a new constraint it'd be nice to reduce chance of breaking existing use cases

1. I think you'd want to want to have some very specialised requirement to want to set a different limit for integers and for BigDecimals. For normal users, it's simpler to have 1 setting. For edge case users, they would have to set the single limit to be the higher of the limits that they have in their heads.
2. 1000 is a very long number under almost any user's definition. From my testing, parsing performance of both integers and BigDecimals tails off dramatically after about 1000 digits - exponential loss in performance.

cowtowncoder:
> > Ok couple of follow-up question on limits:
> > 
> > 1. Is it reasonable to have same length limit for integral and floating-point numbers? Latter are (afaik) much more expensive to handle -- and former might have more legit use cases (for encryption keys, I think?). Or should those be split into 2 settings
> > 2. Is `1000` bit low as a limit? I don't know if there are use case past that, or perhaps limits other parsers use. I know it's to some degree arbitrary, but since we are adding a new constraint it'd be nice to reduce chance of breaking existing use cases
> 
>     1. I think you'd want to want to have some very specialised requirement to want to set a different limit for integers and for BigDecimals. For normal users, it's simpler to have 1 setting. For edge case users, they would have to set the single limit to be the higher of the limits that they have in their heads.
> 
>     2. 1000 is a very long number under almost any user's definition. From my testing, parsing performance of both integers and BigDecimals tails off dramatically after about 1000 digits - exponential loss in performance.

Ok. Besides, I hope we get some feedback during 2.15 RC phase (I'm incurable optimist).
But maybe I should do some googling to see how others set these limits.

cowtowncoder:
@pjfanning Ok one more practical thing. Due to 2.x/3.0 differences, I think it'd be easier to try to merge this in pieces; first adding scaffolding for `StreamReadConstraints` (class itself, passing to `TokenStreamFactory`/`JsonFactory`); merge that 2.15 -> 3.0. And then go back add actual handling. Unfortunately this is kind of change where merging gets bit painful.

I wonder what's the best way to do do that: should I copy parts from this PR and take a first crack; or would you like to do smaller PR based on this, and keep this as the second stage?
I can try to do 2.x -> 3.0 merge for either case.

pjfanning:
I created #845

src/main/java/com/fasterxml/jackson/core/io/IOContext.java
```diff
@@ -115,19 +118,37 @@
...
     @Deprecated // since 2.13
     public IOContext(BufferRecycler br, Object rawContent, boolean managedResource) {
         this(br, ContentReference.rawReference(rawContent), managedResource);
     }
 
+    /**
+     * @return configuration for streaming reads
+     * @since 2.15
+     */
+    public StreamReadConstraints getStreamReadConfig() {
```
cowtowncoder, line 148: let's change to `streamReadConstraints()`.

pjfanning: done

cowtowncoder:
I'll merge this: I have an idea of how to remove need to add parameter for constructing `TextBuffer`, instead taking it to calls to `contentsAsXxx()`

phatblat:
Will this fix be backported to any minor versions before `2.15`?

cowtowncoder:
> Will this fix be backported to any minor versions before `2.15`?

No, this cannot be backported in any earlier versions due to API changes and general complexity of changes.

kingsfleet:
+1 for a backport, 2.15 is not until March

JoshDM:
> +1 for a backport, 2.15 is not until March

+1 for an early release of 2.15.

crahen:
Surely there is a functional patch we can make to existing already released versions of Jackson to limit large numbers. Even if we connect whatever override for length to an environment variable or system property. 

Are the maintainers doing this? When can we expect the patch?

pjfanning:
You're welcome to fork jackson and apply whatever fix you see fit.

cowtowncoder:
@crahen No. Addition of new functionality needs to go in a new minor version. We will not be backporting the changes. So do not expect it at all.

sepatel:
> You can't expect folks to rush out a major version upgrade to protect themselves, very risky.
> 
> Are the maintainers doing this? When can we expect the patch?

They already answered that it cannot be backported. 

> > Will this fix be backported to any minor versions before `2.15`?
> 
> No, this cannot be backported in any earlier versions due to API changes and general complexity of changes.

So yes businesses will have to weigh for themselves the risk of leaving the vulnerability in place or upgrading. Personally, we are going to be upgrading as soon as it is released and already prepping for it. Lot less risk of just upgrading to a new minor release instead.

cowtowncoder:
Jackson 2.15.0 was just released on April 23, 2023 so this is now available for production.

phatblat:
Thanks for the update, @cowtowncoder !

cristinaconstantinc:
Hi! Shouldn't be a CVE created for this?

cowtowncoder:
@cristinaconstantinc There is no real upside from project perspective, but if someone wants to create a reproduction of a problem for older version(s) and file CVE they can go ahead. And if so we would include information about version fixed in and so on.

This based on there not being (to my knowledge) actual published attack against this, work was done pro-actively based on concerns.

PR #828, opened 2022-10-23
Make `BigInteger` parsing lazy

pjfanning:
@cowtowncoder please feel free to reject this - but this relatively small change might be useful for v2.14

This is useful for https://github.com/rallyhealth/weePickle/issues/118

This change means that JsonParser.getNumberType will not need to parse the BigInteger but if you later need the BigInteger (getBigIntegerValue() call), it will then be parsed (just once).

It may also be useful to do something similar with BigDecimal parsing but that change is not as immediately useful.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/base/ParserBase.java
+++ b/src/main/java/com/fasterxml/jackson/core/base/ParserBase.java
@@ -203,6 +203,8 @@ public abstract class ParserBase extends ParserMinimalBase
 
     protected BigDecimal _numberBigDecimal;
 
+    protected String _numberString;
+
     // And then other information about value itself
 
     /**
@@ -607,7 +609,7 @@ public abstract class ParserBase extends ParserMinimalBase
                 return _numberLong;
             }
             if ((_numTypesValid & NR_BIGINT) != 0) {
-                return _numberBigInt;
+                return getBigInteger();
             }
             _throwInternal();
         }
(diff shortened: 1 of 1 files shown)


cowtowncoder:
@pjfanning Excellent! I am thinking I do need to release `2.14.0-rc3`, but these are important improvements I think so that's fine. And yes, testing might show that similar would make sense for `BigDecimal`, but one step at a time.
I hope this merges cleanly to `master` but we'll see.

I hope testing we have is able to find edge cases there might be but... we'll see.

cowtowncoder:
@pjfanning One thing we could do eagerly would to add accessor (I renamed one here to `_getBigInteger()`, minor naming convention thing) for `BigDecimal` too; even if at first it'd still be eagerly accessed.
So something like

```
    protected BigDecimal _getBigDecimal() { return _numberBigDecimal; }
```

in hindsight these probably should have been there from beginning.

pjfanning:
@cowtowncoder I've made big decimal lazy too and renamed the private methods

cowtowncoder:
@pjfanning Is there a PR? I merged this one earlier.

pjfanning:
@cowtowncoder I created https://github.com/FasterXML/jackson-core/pull/830 - I hadn't noticed that this PR was merged and thought that my changes in the related branch would update this PR

cowtowncoder:
Ah. Thank you for creating a new PR, I think updates after merge may not be visible.

PR #837, opened 2022-11-11
uptake latest FastDoubleParser code

pjfanning:
Latest code from https://github.com/wrandelshofer/FastDoubleParser

Has added BigDecimal parsing support that may be useful - but not yet wired in - I can do that in a later PR that optionally allows this parser code instead of BigDecimalUtil


cowtowncoder:
Ok, quite a PR! I'll merge it but may need help with `master` (3.0) merging, will let you know if so.

cowtowncoder:
@pjfanning Ok, unfortunately merge from 2.15 to `master` is... messy. I think I'll need your help with that again.

I also wonder if there's any way we could modularize things somehow, to isolate FastDoubleParser package. Not sure how that'd work; multi-Maven-module set up with shading might work but... that sounds complex and clunky.
And I guess "hiding" packages under unchanging Maven package within project would also defeat the benefits of namespacing -- like, f.ex if we (hypothetically) decided to use "local.foo." as package for 2.x and 3.x we'd have collision again (at least pre-Module system).
If FDP existed in published jar we could use Shade plug-in to add it but I don't know if that exists either?
If so 2.x and 3.x could re-package it to different places.

pjfanning:
Unfortunately, FastDoubleParser is not released as a jar (at least not recently). I can do a new PR for the master branch tomorrow.

cowtowncoder:
Thank you @pjfanning.

One thing that might help with merging would also be staging changes (in case of renames f.ex) but that might not be an option when just syncing with changes from FDP.

PR #845, opened 2022-11-25
add StreamReadConstraints

pjfanning:
preliminary change for #827

some formatting has been updated as a result of copying the text from the original github PR and and my IDE removing stray whitespace - I can try harder to put back the stray whitespace, if necessary


cowtowncoder (approved):

cowtowncoder:
Ok thanks! Np wrt white space; most are good, couple of cases where it'd be nice to have old ... I can change them if need be but probably won't.

Thank  you for doing this change first; 2.x/3.x discrepancy is bit tricky for now.

cowtowncoder:
Oh fudge. Forgot to check naming part. Will need to redo...

PR #847, opened 2022-11-29
Add SLSA provenance to release workflow

### LINKED ISSUE #844: Add SLSA provenance via build script
Hey, I'm Pedro and I'm working on behalf of Google and the [Open Source Security Foundation][ossf] (OpenSSF). Given the significant [increase in supply-chain attacks][sonatype], the OpenSSF is focused on improving the security of the open-source ecosystem as a whole. For example, #792 was based on feedback from Scorecards, an OpenSSF tool.

The OpenSSF has also developed the [SLSA][slsa] specification for projects to attest to a published artifact's provenance, allowing its consumers to ensure that the artifact comes from a trusted source. There are also GitHub workflows to securely generate this provenance and CLI tools to verify an artifact's authenticity.

Given how Jackson is almost synonymous with JSON in the Java ecosystem, the OpenSSF has placed Jackson on its list of the 100 most important open-source projects. I'd therefore like to offer to help jackson-core incorporate SLSA into its deploy workflow.

Would you be interested in a PR to adopt SLSA?

[ossf]: https://openssf.org/
[slsa]: https://slsa.dev/
[sonatype]: https://www.sonatype.com/state-of-the-software-supply-chain/introduction

pnacht:
Closes #844.

As per the linked issue, this PR adds automatic SLSA provenance generation to the release workflow (main.yml). SLSA provenance allows maintainers and consumers to guarantee that the jars came from the expected, trusted source. See the linked issue for more information.

Currently the provenance attestation is only "attached" to the workflow (or added to GitHub releases, if the project used them). Given that Maven allows for arbitrary files in their releases, it would be best if the attestation could also be added to the release. However, I must admit my ignorance as to how precisely to do that via GitHub Actions.

You'll notice the provenance is created in a separate job from the rest of the workflow. This is for security purposes, to ensure the provenance generation is isolated and secure against external interference. If there's interest, I could submit another PR to split the build job into build, deploy, and codecov jobs. This would ensure that things such as credentials cannot leak to other steps (see [GitHub's docs on risks of compromised runners](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#accessing-secrets)).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,20 +2,20 @@ name: Build and Deploy Snapshot
 on:
   push:
     branches:
-      - master
-      - "3.0"
-      - "2.14"
+    - master
+    - "3.0"
+    - "2.14"
     paths-ignore:
-      - "README.md"
-      - "release-notes/*"
+    - "README.md"
+    - "release-notes/*"
   pull_request:
     branches:
-      - master
-      - "3.0"
-      - "2.14"
+    - master
+    - "3.0"
+    - "2.14"
     paths-ignore:
-      - "README.md"
-      - "release-notes/*"
+    - "README.md"
+    - "release-notes/*"
 permissions:
   contents: read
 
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -26,18 +26,22 @@ jobs:
       fail-fast: false
       matrix:
         # As of Jackson 2.14 got Java 8 baseline so can build on later JDKs too
-        java_version: ['8', '11', '17']
-        os: ['ubuntu-20.04']
+        java_version: ["8", "11", "17"]
```
cowtowncoder, line 29: Any particular reason for this change?

.github/workflows/main.yml
```diff
@@ -59,6 +63,27 @@ jobs:
...
         # MAVEN_GPG_PASSPHRASE: ${{ secrets.MAVEN_GPG_PASSPHRASE }}
       run: ./mvnw -B -q -ff -DskipTests -ntp source:jar deploy
+    # Until generator_generic_slsa3@v1.2.3, calculate hashes from ./target
+    # https://github.com/slsa-framework/slsa-github-generator/issues/1225
+    - name: Generate hash
+      id: hash
+      if: |
+        github.event_name != 'pull_request' &&
+        matrix.java_version == '8' &&
+        endsWith(steps.projectVersion.outputs.version, '-SNAPSHOT')
```
cowtowncoder, line 73: This would only calculate hashes for SNAPSHOT builds, not for actual releases?

.github/workflows/main.yml
pnacht: Oh sorry, that was my linter! You'll notice the last commit was fixing some whitespace changes it'd also done, but I missed this change, my apologies. I can undo it, if you prefer.

.github/workflows/main.yml
cowtowncoder: Ah ok.

So here we might have a challenge: releases are not triggered by CI -- I trigger them from my work station using Maven release plug-in (so second option).
But while I know how releases are often triggered via CI (from tagging), I haven't quite found those to my liking, at least the way most corporations do it: they tend to geared towards patch-increase only cases, for (web) applications where there is little semantic information in version numbers.

Work involved would also be quite substantial as well considering none of Jackson repos uses CI releases at this point.
And I do need to be able to release versions from multiple branches.

But longer term I can see why use of CI would be preferable to local work station -- for multiple reasons. So I am not philosophically against it at all. Just more concerned about practical path there.

One thing that might help would be finding good CI workflow to consider adopting: something used for library/framework use case where semantic versioning is used (or rather, where version number increments need to be somewhat more flexible; including occasional release of micro-patches).

I haven't really followed how other OSS projects deal with this so not sure where to look. I guess I could ask on Tidelift OSS slack channels; Joda project might be a good candidate to consider.

pnacht: Let me know if I'm missing something, but GitHub workflows should be able to handle different branches and be indifferent to tag naming conventions.

We could likely create a workflow very similar to this one but that's triggered:
- manually (`on: workflow_dispatch`). You'd pass the branch/tag/commit to use by hand. 
- whenever a new GitHub Release is created (`on: release`). So you define the branch/tag/commit from which to run. For an example (including SLSA provenance generation), see [the release workflow for the `sigstore-python` library](https://github.com/sigstore/sigstore-python/blob/main/.github/workflows/release.yml).

You can also use the `gh` CLI or simply `curl` to perform these actions remotely.

I'm not an expert on releasing on Maven, so I'm unsure if there are additional steps required for actual releases.

(Let me know if you'd rather migrate this conversation somewhere more appropriate)

pnacht: > And I do need to be able to release versions from multiple branches.

It just occurred to me you might've meant you need to be able to release from multiple repos (jackson & jackson-core, for example). In this case, we could write a shell script to make the `gh/curl` calls for all the repos at once.

cowtowncoder: My concern is not that Github workflows can't do things, but rather how awkward their use is. Right now use of

    mvn release:clean release:prepare release:perform

is straight-forward with little git interaction needed. Increases versions etc, handles everything except for final approval on Sonatype OSS repo (which I'll need to do anyway).
Pushing tags etc has always seemed pretty clunky in comparison.

And wrt multi-repo release, there is no workflows for that so no need to worry about that yet.

pnacht: Sorry for the delay in replying, I've been getting more familiar with the Maven release plugin.

As I've come to understand it from [the docs](https://maven.apache.org/maven-release/maven-release-plugin/examples/prepare-release.html), the magic really happens in the `prepare` stage. My understanding is that `prepare`:

- Bumps everything from x-SNAPSHOT to the new release version
- Commits the changed POMs
- Creates a tag for the new version
- Bumps everything to the next y-SNAPSHOT
- Commits the changed POMs again

Meanwhile, `perform` simply builds and deploys. From a security standpoint, however, the most important step to have in CI is actually `perform`.

I've therefore had the idea of splitting the workflow into two parts: we can write a shell script you run on your local machine that runs `:clean :prepare` and pushes everything to GitHub (including the tag created by `prepare`). We can then have a GitHub workflow that triggers on new tags (excluding `*-rc` release-candidate tags) and simply runs the final `:perform`.

Your release process would therefore be almost unchanged. Instead of calling `mvn release:clean release:prepare release:perform`, you'd call `release.sh`, which would give you the same behavior as `mvn release:clean release:prepare` and then automatically trigger the CI `mvn release:perform` by pushing the new commits and tag.

Would you be interested in exploring this?

cowtowncoder: Hi @pnacht ! Apologies for slow response here. But: yes, I think that sounds like a reasonable compromise and I would be interested.
I think we could play with this for core components (jackson-core, jackson-databind, jackson-annotations).

One quick note: I will be now off for vacation until January 4th 2023 so let's sync up early next year.

pnacht: Hey @cowtowncoder, hope you had a good vacation! I've started sketching out the script to replace `mvn release:clean release:prepare` and should have the workflow that'll do the final `mvn release:perform` end of next week. I'll submit those as a separate PR, is that alright?

cowtowncoder: @pnacht That makes sense to me (separate PR)

pnacht: Sorry for the delay! But I've submitted the PR as discussed above: https://github.com/FasterXML/jackson-core/pull/896.

cowtowncoder:
@pnacht just to make sure: can this be closed as I am to merge #896?

pnacht:
Yep!

PR #848, opened 2022-11-29
add tests for writing large numbers

pjfanning:


cowtowncoder:
Ok is the idea here to ensure that the new limits do NOT apply to writer side? Or... ?
If so, may want to try to use for multiple backends (char- and byte-based; `Writer` vs `ByteArrayOutputStream`)?

pjfanning:
Yes - the idea is to ensure the read limits and their defaults are not applied to writes.

I can look at adding similar tests for other backends.

src/test/java/com/fasterxml/jackson/core/write/UTF8GeneratorTest.java
```diff
@@ -183,6 +188,43 @@ public void testWriteLargeIntegerStringWriter() throws Exception
         }
     }
 
+    public void testWriteLargeIntegerDataOutput() throws Exception
```
cowtowncoder, line 191: Maybe add a brief comment on background, testing that large/long numbers can be written fine, without changing default settings?

src/test/java/com/fasterxml/jackson/core/write/UTF8GeneratorTest.java
```diff
@@ -183,6 +188,43 @@ public void testWriteLargeIntegerStringWriter() throws Exception
...
+            gen.writeNumber(bigInt);
+            gen.writeEndObject();
+            gen.close();
+
+            JsonFactory jsonFactory = JsonFactory.builder()
+                    .streamReadConstraints(StreamReadConstraints.builder().maxNumberLength(Integer.MAX_VALUE).build())
+                    .build();
+            try(
+                    ByteArrayInputStream bis = new ByteArrayInputStream(out.toByteArray());
+                    DataInputStream dis = new DataInputStream(bis);
```
cowtowncoder, line 214: Hmm?

cowtowncoder: Oh, I see. That's for DataInput testing.

src/test/java/com/fasterxml/jackson/core/write/UTF8GeneratorTest.java
cowtowncoder: Actually n/m, I can add.

cowtowncoder (approved):

PR #850, opened 2022-11-30
use maven-shade-plugin minimizeJar=true

pjfanning:
this option reduces the size of the built jar (but only a little)


cowtowncoder:
@pjfanning Actually it looks as if this made rather big positive difference? (unless I misread something)

From over 800kB to 640kB, so almost half of earlier size add. That sounds more reasonable to me! :)

PR #852, opened 2022-11-30
latest schubfach code (part of JDK 19)

pjfanning:
The Schubfach code to write floats/doubles as strings is part of JDK 19. https://github.com/openjdk/jdk/tree/master/src/java.base/share/classes/jdk/internal/math

This PR brings in that code but makes it work with JDK 8+.

Users of JDK19 should not enable the  USE_FAST_DOUBLE_WRITER feature because the JDK internalized code should run faster.


cowtowncoder (approved):

cowtowncoder:
Ok looks like some use of `ThreadLocal` was removed, which is good.

PR #853, opened 2022-12-01
go back to using switch statements in schubfach code

pjfanning:
When I ported over #852, I had compile warnings about some switch statements. I rewrote them using if statements.
In hindsight, I should have looked more closely - the warnings were about some specific features, the switch statements themselves are ok.


cowtowncoder (approved):

cowtowncoder:
@pjfanning yes, +1 for that if you have time? I agree, sounds like low risk to me (not minimal but low enough).

PR #855, opened 2022-12-03
make streamReadConstraints public so it can be referenced in downstream modules

pjfanning:
relates to https://github.com/FasterXML/jackson-databind/issues/3686

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/base/ParserBase.java
+++ b/src/main/java/com/fasterxml/jackson/core/base/ParserBase.java
@@ -800,6 +800,11 @@ public abstract class ParserBase extends ParserMinimalBase
         return _getBigDecimal();
     }
 
+    @Override // @since 2.15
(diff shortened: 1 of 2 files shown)


cowtowncoder (approved):

src/main/java/com/fasterxml/jackson/core/base/ParserBase.java
```diff
@@ -800,6 +800,11 @@ public BigDecimal getDecimalValue() throws IOException
         return _getBigDecimal();
     }
 
+    @Override // @since 2.15
```
cowtowncoder, line 803: Ok I merged this but then realized that maybe this is actually needed in base `JsonParser` itself?

pjfanning: It looks like streamReadConstraints will need to move up JsonParser class. I'll do another PR.

PR #856, opened 2022-12-03
move streamReadConstraints() to JsonParser (inherited by ParserMinimalBase)

pjfanning:
follow on from #855


cowtowncoder:
Sounds good: just one more thing: could you also add this in `JsonParserDelegate`?

pjfanning:
> Sounds good: just one more thing: could you also add this in `JsonParserDelegate`?

done

cowtowncoder (approved):

PR #858, opened 2022-12-04
expose streamReadConstraints() on TokenStreamFactory

pjfanning:
useful for TOML case - the Parser class should validate num length and the TomlFactory that inherits from TokenStreamFactory would be a good place to get to get the constraints.


cowtowncoder (approved):

cowtowncoder:
Sounds good. I think I'll change this a little bit, knowing that in 2.x every factory basically must extend `JsonFactory` anyway -- but the idea remains the same. I'll make accessor `abstract` for `TokenStreamFactory` in 2.x.

cowtowncoder:
Done. Needed to make `StreamReadConstraints` serializable as well.

PR #865, opened 2022-12-15
Optimize parsing 19 digit longs

marschall:
Parsing 19 digit long values is significantly slower than parsing 18 digit long values as we are hitting a slow path that does a string allocation.

19 digit long values are quite common when a random number generator is used to generate long values.

We have to pay attention that we only use this fast path if the value is a valid long. This can be achieved by calling `#inLongRange` first.

In the following synthetic microbenchmark a throughput increase of about 33% with a drastic reduction in the allocation rate can be achieved.

```java

public class JacksonStdLongReadVanilla
{

    private static final JsonFactory JSON_FACTORY = new JsonFactory();

    private static final String JSON;

    static {
        StringBuilder json = new StringBuilder();
        Random random = new Random(0L);
        json.append('[');
        for (int i = 0; i < 1_000; i++) {
            if (i > 0) {
                json.append(',');
            }
            json.append(random.nextLong());
        }
        json.append(']');
        JSON = json.toString();
    }

    @Benchmark
    public void parseLong(Blackhole blackhole) throws IOException {
        try (JsonParser parser = JSON_FACTORY.createParser(JSON)) {
            parser.nextToken();

            JsonToken nextToken = parser.nextToken();
            while (nextToken == JsonToken.VALUE_NUMBER_INT) {
                blackhole.consume(parser.getLongValue());
                nextToken = parser.nextToken();
            }
        }
    }

}

```

If we have a look at the allocations before we see that it is dominated by `byte[]` allocations for `String` allocations

allocation-site-before

Resulting in many gcs

gcs-before

With the changes in this PR the  `byte[]` and `String` allocations are gone

allocation-site-after

Resulting in no more gcs

gcs-after

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/io/NumberInput.java
+++ b/src/main/java/com/fasterxml/jackson/core/io/NumberInput.java
@@ -152,6 +152,32 @@ public final class NumberInput
         long val = parseInt(ch, off, len1) * L_BILLION;
         return val + (long) parseInt(ch, off+len1, 9);
     }
+    
+    /**
+     * Parses an unsigned long made up of exactly 19 digits.
+     * <p>
+     * It is the callers responsibility to make sure the input is exactly 19 digits.
+     * and fits into a 64bit long by calling {@link #inLongRange(char[], int, int, boolean)}
+     * first.
+     * <p>
+     * Note that input String must NOT contain leading minus sign (even
+     * if {@code negative} is set to true).
+     * 
+     * @param ch Buffer that contains integer value to decode
+     * @param off Offset of the first digit character in buffer
+     * @param negative Whether original number had a minus sign
+     * @return Decoded {@code long} value
+     */
+    public static long parseLong19(char[] ch, int off, boolean negative)
+    {
+        // Note: caller must ensure length is [10, 18]
+        long num = 0L;
+        for (int i = 0; i < 19; i++) {
+            char c = ch[off + i];
+            num = (num * 10) + (c - '0');
+        }
+        return negative ? -num : num;
+    }
 
     /**
      * Similar to {@link #parseInt(String)} but for {@code long} values.
--- a/src/main/java/com/fasterxml/jackson/core/base/ParserBase.java
+++ b/src/main/java/com/fasterxml/jackson/core/base/ParserBase.java
@@ -939,7 +939,6 @@ public abstract class ParserBase extends ParserMinimalBase
 
     private void _parseSlowInt(int expType) throws IOException
     {
-        final String numStr = _textBuffer.contentsAsString();
         try {
             int len = _intLength;
             char[] buf = _textBuffer.getTextBuffer();
@@ -949,10 +948,10 @@ public abstract class ParserBase extends ParserMinimalBase
             }
             // Some long cases still...
             if (NumberInput.inLongRange(buf, offset, len, _numberNegative)) {
-                // Probably faster to construct a String, call parse, than to use BigInteger
-                _numberLong = NumberInput.parseLong(numStr);
+                _numberLong = NumberInput.parseLong19(buf, offset, _numberNegative);
                 _numTypesValid = NR_LONG;
             } else {
+                String numStr = _textBuffer.contentsAsString();
                 // 16-Oct-2018, tatu: Need to catch "too big" early due to [jackson-core#488]
                 if ((expType == NR_INT) || (expType == NR_LONG)) {
                     _reportTooLongIntegral(expType, numStr);
@@ -970,6 +969,7 @@ public abstract class ParserBase extends ParserMinimalBase
                 }
             }
         } catch (NumberFormatException nex) {
+            String numStr = _textBuffer.contentsAsString();
             // Can this ever occur? Due to overflow, maybe?
             _wrapError("Malformed numeric value ("+_longNumberDesc(numStr)+")", nex);
         }
--- a/src/test/java/com/fasterxml/jackson/core/read/NumberParsingTest.java
+++ b/src/test/java/com/fasterxml/jackson/core/read/NumberParsingTest.java
@@ -278,7 +278,9 @@ public class NumberParsingTest
         for (int mode : ALL_MODES) {
             long belowMinInt = -1L + Integer.MIN_VALUE;
             long aboveMaxInt = 1L + Integer.MAX_VALUE;
-            String input = "[ "+Long.MAX_VALUE+","+Long.MIN_VALUE+","+aboveMaxInt+", "+belowMinInt+" ]";
+            long belowMaxLong = -1L + Long.MAX_VALUE;
+            long aboveMinLong = 1L + Long.MIN_VALUE;
+            String input = "[ "+Long.MAX_VALUE+","+Long.MIN_VALUE+","+aboveMaxInt+", "+belowMinInt+","+belowMaxLong+", "+aboveMinLong+" ]";
             JsonParser p = createParser(jsonFactory(), mode, input);
             assertToken(JsonToken.START_ARRAY, p.nextToken());
             assertToken(JsonToken.VALUE_NUMBER_INT, p.nextToken());
@@ -296,6 +298,14 @@ public class NumberParsingTest
             assertToken(JsonToken.VALUE_NUMBER_INT, p.nextToken());
             assertEquals(JsonParser.NumberType.LONG, p.getNumberType());
             assertEquals(belowMinInt, p.getLongValue());
+
+            assertToken(JsonToken.VALUE_NUMBER_INT, p.nextToken());
+            assertEquals(JsonParser.NumberType.LONG, p.getNumberType());
+            assertEquals(belowMaxLong, p.getLongValue());
+
+            assertToken(JsonToken.VALUE_NUMBER_INT, p.nextToken());
+            assertEquals(JsonParser.NumberType.LONG, p.getNumberType());
+            assertEquals(aboveMinLong, p.getLongValue());
 
             
             assertToken(JsonToken.END_ARRAY, p.nextToken());        


src/main/java/com/fasterxml/jackson/core/io/NumberInput.java
```diff
@@ -152,6 +152,32 @@ public static long parseLong(char[] ch, int off, int len)
         long val = parseInt(ch, off, len1) * L_BILLION;
         return val + (long) parseInt(ch, off+len1, 9);
     }
+    
+    /**
+     * Parses an unsigned long made up of exactly 19 digits.
```
pjfanning, line 157: this looks wrong - the call you are replacing allows signed numbers - if needs be you can change the calling code to only call this if _numberNegative is true

src/main/java/com/fasterxml/jackson/core/io/NumberInput.java
```diff
@@ -152,6 +152,32 @@ public static long parseLong(char[] ch, int off, int len)
...
+    public static long parseLong19(char[] ch, int off, boolean negative)
+    {
+        // Note: caller must ensure length is [10, 18]
+        long num = 0L;
+        for (int i = 0; i < 19; i++) {
+            char c = ch[off + i];
+            num = (num * 10) + (c - '0');
+        }
+        return negative ? -num : num;
+    }
```
pjfanning, line 180: can you provide a jmh test to prove this is faster? you can extend https://github.com/pjfanning/jackson-number-parse-bench if you like

pjfanning: if this is really such a performance drag, shouldn't there be an openjdk issue - or something similar?

pjfanning:
Another thing that I'm concerned with is that with a custom long parser, we would need to have test coverage for a wide range of possible values.

The FastDoubleParser that we allow use of is
* optional (and not the default)
* FastDoubleParser has good test coverage
* it is admittedly still quite experimental and users who opt in to using it will need to do so with caution

marschall:
> Another thing that I'm concerned with is that with a custom long parser, we would need to have test coverage for a wide range of possible values.
> 

`NumberInput` is already full of custom number parsing code, I refer you to the various `#parseInt` methods. There is also the custom `BigDecimalParser`. But I can extend the tests in `NumberParsingTest`.

src/main/java/com/fasterxml/jackson/core/io/NumberInput.java
marschall: > if this is really such a performance drag, shouldn't there be an openjdk issue - or something similar?

Outsiders can't create OpenJDK issues. And even if, the discussions alone would probably take years. The time investment is massive with no guaranteed outcome. To give you a concrete example https://github.com/openjdk/jdk/pull/6275.

Even if everything would go smoothly we would still have to wait years until Jackson updates the minimum version to one that has the fixes. Jackson currently has a base of JDK 8 which was released 8 years ago. So we're looking at a best case of 10 years.

marschall: > can you provide a jmh test project to prove this is faster? you can extend https://github.com/pjfanning/jackson-number-parse-bench if you like - you obviously have the test case, just would be handy to be able to run it independently

I saw there is also https://github.com/FasterXML/jackson-benchmarks let me know what you prefer.

pjfanning: Whichever suits you. I have a vague recollection that jackson-benchmarks is configured to run the benchmarks for a long time. There was something about the project that made me go off and create my own where I could reconfigure the settings more readily.

cowtowncoder:
Although `jackson-benchmarks` is useful for comparing performance of various Jackson backends over relatively simple POJO type, it's not as good for micro-benchmarks for aspects like this. So I think I'd go with https://github.com/pjfanning/jackson-number-parse-bench.

src/main/java/com/fasterxml/jackson/core/io/NumberInput.java
```diff
@@ -152,6 +152,32 @@ public static long parseLong(char[] ch, int off, int len)
...
+     */
+    public static long parseLong19(char[] ch, int off, boolean negative)
+    {
+        // Note: caller must ensure length is [10, 18]
+        long num = 0L;
+        for (int i = 0; i < 19; i++) {
+            char c = ch[off + i];
+            num = (num * 10) + (c - '0');
+        }
+        return negative ? -num : num;
```
cowtowncoder, line 179: Another thing to ensure is that `Long.MIN_VALUE` works (negative range has one more value than positive and this means there's one very special value where negating positive value will not work as expected).

pjfanning:
Could NumberInput.parseLong be changed to call parseLong19 where appropriate - it already uses parseInt for small nums. Would it be simpler just to have parseLong parse all numbers the new way (if the performance was not worse)?

marschall:
> Although `jackson-benchmarks` is useful for comparing performance of various Jackson backends over relatively simple POJO type, it's not as good for micro-benchmarks for aspects like this. So I think I'd go with https://github.com/pjfanning/jackson-number-parse-bench.

There you go https://github.com/pjfanning/jackson-number-parse-bench/pull/2

I had to change `build.gradle`

```groovy
repositories {
    mavenLocal()
}

jmh {
    includes = ['org.example.jackson.bench.LongParserBench']
    profilers = ['gc'] // or whatever you like best like jfr or async
}
```

On a Ryzen 7 PRO 5750GE we go from

```
Benchmark                                                    Mode  Cnt      Score      Error   Units
LongParserBench.parseLong                                   thrpt   25  21860.758 ± 1076.882   ops/s
LongParserBench.parseLong:·gc.alloc.rate                    thrpt   25   1139.800 ±   56.144  MB/sec
LongParserBench.parseLong:·gc.alloc.rate.norm               thrpt   25  57410.238 ±    0.043    B/op
LongParserBench.parseLong:·gc.churn.G1_Eden_Space           thrpt   25   1136.527 ±   59.318  MB/sec
LongParserBench.parseLong:·gc.churn.G1_Eden_Space.norm      thrpt   25  57240.652 ±  766.795    B/op
LongParserBench.parseLong:·gc.churn.G1_Survivor_Space       thrpt   25      0.006 ±    0.002  MB/sec
LongParserBench.parseLong:·gc.churn.G1_Survivor_Space.norm  thrpt   25      0.303 ±    0.079    B/op
LongParserBench.parseLong:·gc.count                         thrpt   25    504.000             counts
LongParserBench.parseLong:·gc.time                          thrpt   25    334.000                 ms
```

to 
```
Benchmark                                                    Mode  Cnt      Score      Error   Units
LongParserBench.parseLong                                   thrpt   25  31230.896 ± 1532.563   ops/s
LongParserBench.parseLong:·gc.alloc.rate                    thrpt   25     14.523 ±    0.713  MB/sec
LongParserBench.parseLong:·gc.alloc.rate.norm               thrpt   25    512.025 ±    0.029    B/op
LongParserBench.parseLong:·gc.churn.G1_Eden_Space           thrpt   25     15.785 ±   19.352  MB/sec
LongParserBench.parseLong:·gc.churn.G1_Eden_Space.norm      thrpt   25    544.799 ±  668.200    B/op
LongParserBench.parseLong:·gc.churn.G1_Survivor_Space       thrpt   25     ≈ 10⁻⁴             MB/sec
LongParserBench.parseLong:·gc.churn.G1_Survivor_Space.norm  thrpt   25      0.005 ±    0.011    B/op
LongParserBench.parseLong:·gc.count                         thrpt   25      7.000             counts
LongParserBench.parseLong:·gc.time                          thrpt   25     10.000                 ms
```

We go from 21860.758 ops/s to 31230.896 ops/s.
We go from an allocation rate of 1139 MB/sec to 14 MB/sec.
We go from 504 gcs taking 334 ms to 7 gc taking 10 ms.

src/main/java/com/fasterxml/jackson/core/io/NumberInput.java
marschall: This is already covered by `NumberParsingTest#testLongRange`

marschall:
> > Another thing that I'm concerned with is that with a custom long parser, we would need to have test coverage for a wide range of possible values.
> 
> `NumberInput` is already full of custom number parsing code, I refer you to the various `#parseInt` methods. There is also the custom `BigDecimalParser`. But I can extend the tests in `NumberParsingTest`.

`NumberParsingTest` already covered a lot of cases, I expanded them a bit.

src/main/java/com/fasterxml/jackson/core/io/NumberInput.java
```diff
@@ -152,6 +152,32 @@ public static long parseLong(char[] ch, int off, int len)
...
+     * if {@code negative} is set to true).
+     * 
+     * @param ch Buffer that contains integer value to decode
+     * @param off Offset of the first digit character in buffer
+     * @param negative Whether original number had a minus sign
+     * @return Decoded {@code long} value
+     */
+    public static long parseLong19(char[] ch, int off, boolean negative)
+    {
+        // Note: caller must ensure length is [10, 18]
```
pjfanning, line 173: can you remove this comment line which doesn't match this code?

src/main/java/com/fasterxml/jackson/core/io/NumberInput.java
marschall: This the same approach that `com.fasterxml.jackson.core.util.TextBuffer#contentsAsInt(boolean)` uses, it calls `com.fasterxml.jackson.core.io.NumberInput#parseInt(char[], int, int)` that parses an optionally singed value unsigned by ignoring the sign and later flips the sign if it had a minus sign.

src/main/java/com/fasterxml/jackson/core/io/NumberInput.java
cowtowncoder: Ok, good, thank you for verifying.

cowtowncoder (approved):

cowtowncoder:
Thank you @marschall -- I'll merge this. May do minor tweaks but looks good & can go in 2.15.0.

marschall:
> > Could NumberInput.parseLong be changed to call parseLong19 where appropriate - it already uses parseInt for small nums. Would it be simpler just to have parseLong parse all numbers the new way (if the performance was not worse)?
> 
> Right now method is quite particular expecting length of 10 - 18 characters and there'd be one more check (and branch). Not sure it would affect optimizer that much; I don't have strong opinion where to modify it and add checks or keep separate method. Separate method would do well to follow the same pattern tho; decoding 2 ints, using `int` multiplications is faster than doing 19 `long` multiplications. I remember benchmarking that difference years ago and it was non-trivial.

I had a look at this, it would require a bit more changes. Currently `#parseInt` only handles `int`s with length 9. `int`s with length 10 are parsed as `long` and then converted to `int`. Therefore we can't call `#parseInt` twice to parse `long` of length 19, only `long` of length 18.

If we we wanted to change this we would have to:
- Introduce `NumberInput#inIntRange`
- Change `ParserBase#_parseIntValue()` and `ParserBase#_parseNumericValue(int)` to call `NumberInput#inIntRange` for length 10 and based on this either call `TextBuffer#contentsAsInt(boolean)` or `TextBuffer#contentsAsLong(boolean)`.
- Change `#parseInt` to handle a length of up to 10

cowtowncoder:
I think we are ok for now; if anyone really wants to optimize things further, a new PR would be fine.
There's fine line between handling 10/19 lengths optimally vs handling shorter ones -- for most use cases I suspect it is more important to have close(r) to optimal parsing for shorter numbers than longer. That is: adding more checks (-> more branching in code) for different cases slows down existing cases.

PR #874, opened 2022-12-20
Fix #873: change the way number maximum length limits are imposed

### LINKED ISSUE #873: Rewrite number-limit changes using `ParserBase.resetInt()/Float()`
Jackson 2.15 will add limits to various entities, including numbers. Initial implementation adds methods in `TextBuffer` but there's a more natural place to impose restrictions: methods in `ParserBase` that set state for number decoding.

cowtowncoder:


cowtowncoder:
@pjfanning Ok so I decided to change the way number limits are imposed: this seems like a better place to do it.

One potential downside is that this does not allow separate handling of `BigDecimal` vs float/double lengths, given that distinction would only be available at point where access is know -- but I suspect it would be difficult to gather anyway due to various conversions. What we might want to do instead is simply truncate longer `double`/`float` inputs before invoking (Fast) DoubleParser?

PR #875, opened 2022-12-20
deprecate method that is removed in v2.15 branch

pjfanning:
This method is not used in main source of jackson-core (in 2.14 branch).

It was removed in recent 2.15 commits. Seems ok to remove it.

Just seems tidiest to flag it for deprecation in 2.14 branch.


cowtowncoder:
Uhhh. Good catch ... actually, my bad. I thought it was being added in 2.15, not that it was already in 2.14 (but seeing it has no `@since` I'm not sure why I thought so).

How about I go and actually restore it as deprecated in 2.15 instead.

PR #877, opened 2022-12-21
private expand method ignores its input param

pjfanning:
We either use the param that's passed in or we stop passing it in.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/util/TextBuffer.java
+++ b/src/main/java/com/fasterxml/jackson/core/util/TextBuffer.java
@@ -661,7 +661,7 @@ public final class TextBuffer
         // Room in current segment?
         char[] curr = _currentSegment;
         if (_currentSize >= curr.length) {
-            expand(1);
+            expand();
             curr = _currentSegment;
         }
         curr[_currentSize++] = c;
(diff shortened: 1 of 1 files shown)


src/main/java/com/fasterxml/jackson/core/util/TextBuffer.java
```diff
@@ -661,7 +661,7 @@ public void append(char c) {
         // Room in current segment?
         char[] curr = _currentSegment;
         if (_currentSize >= curr.length) {
-            expand(1);
+            expand();
```
cowtowncoder, line 664: Here, 1 is fine, only appending one char.

src/main/java/com/fasterxml/jackson/core/util/TextBuffer.java
```diff
@@ -694,7 +694,7 @@ public void append(char[] c, int start, int len)
         // And then allocate new segment; we are guaranteed to now
         // have enough room in segment.
         do {
-            expand(len);
+            expand();
             int amount = Math.min(_currentSegment.length, len);
             System.arraycopy(c, start, _currentSegment, 0, amount);
```
cowtowncoder, line 699: ... and here presumably.... hmmh. I guess it actually isn't strictly required that length. It's just sort of a ... wish?

(for a moment I thought this is where `len` must be obeyed but no, calc of `amount` takes into account that we get what we get).

cowtowncoder (approved):

cowtowncoder:
@pjfanning Thank you. Yes, you are right; code already takes care not to expect any specific expansion. So that param is useless. Will merge.

PR #878, opened 2022-12-21
change parseLong to try to improve performance

pjfanning:
I changed NumberInput.parseLong(String) to work more like NumberInput,parseInt(String). Iterating over the chars and scaling them and added the scaled results.

Recent changes added a parseLong19 but this is unaffected by this change (it is used by a different code path).

This seems to have improved performance based on https://github.com/pjfanning/jackson-number-parse-bench/commit/f86dfcbead34cf60a6aeb0b31f4782dbfd582c23

Tested using Java 8 on my Macbook:
```
Benchmark                               Mode  Cnt      Score     Error  Units
LongParserBench.alternateJacksonParse  thrpt    5  25442.760 ± 212.009  ops/s
LongParserBench.existingJacksonParse   thrpt    5  14697.121 ± 323.988  ops/s
```

With Java 17, the performance results see the existingJacksonParse almost catch up (but there is still a non-trivial advantage to alternateJacksonParse).

Only open for discussion at the moment - @marschall, @cowtowncoder  could you have a look?

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/io/NumberInput.java
+++ b/src/main/java/com/fasterxml/jackson/core/io/NumberInput.java
@@ -190,14 +190,55 @@ public final class NumberInput
      */
     public static long parseLong(String s)
     {
-        // Ok, now; as the very first thing, let's just optimize case of "fake longs";
-        // that is, if we know they must be ints, call int parsing
-        int length = s.length();
-        if (length <= 9) {
-            return (long) parseInt(s);
+        /* Ok: let's keep strategy simple: ignoring optional minus sign,
+         * we'll accept 1 - 19 digits and parse things efficiently;
+         * otherwise just defer to JDK parse functionality.
+         */
+        char c = s.charAt(0);
+        int len = s.length();
+        boolean neg = (c == '-');
+        int offset = 1;
+        // must have 1 - 19 digits after optional sign:
+        // negative?
+        if (neg) {
+            if (len == 1 || len > 20) {
+                return Long.parseLong(s);
+            }
+            c = s.charAt(offset++);
+        } else {
+            if (len > 19) {
+                return Long.parseLong(s);
+            }
         }
-        // !!! TODO: implement efficient 2-int parsing...
-        return Long.parseLong(s);
+        if (c > '9' || c < '0') {
+            return Integer.parseInt(s);
+        }
+        long num = c - '0';
+        if (offset < len) {
+            c = s.charAt(offset++);
+            if (c > '9' || c < '0') {
+                return Long.parseLong(s);
+            }
+            num = (num * 10) + (c - '0');
+            if (offset < len) {
+                c = s.charAt(offset++);
+                if (c > '9' || c < '0') {
+                    return Long.parseLong(s);
+                }
+                num = (num * 10) + (c - '0');
+                // Let's just loop if we have more than 3 digits:
+                if (offset < len) {
+                    do {
+                        c = s.charAt(offset++);
+                        if (c > '9' || c < '0') {
+                            return Long.parseLong(s);
+                        }
+                        num = (num * 10) + (c - '0');
+                    } while (offset < len);
+                }
+            }
+        }
+        return neg ? -num : num;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


cowtowncoder:
I am not against improvements here, but one thing to note is that I think this helper method is not used very often is it? JsonParser's usual decoder uses `char[]` based decoding methods instead and this is (I think?) only used for coercion use cases (number from JSON String), or as fallback for some cases.

pjfanning:
The method in this PR is used in jackson-databind and some of the dataformat libs too.

cowtowncoder:
@pjfanning Yes but only in secondary calls -- primary ones that call, say `JsonParser.getLongValue()` will not use it, as far as I remember. But it would be used when coercion is needed from JSON String; or for backends that offload decoding (YAML, CSV).

Actually now that I wrote that I realize that this could actually be somewhat significant for CSV (and maybe YAML) use cases. So it's just JSON where it is unlikely to have much effect.

I guess that makes this more useful than I first thought.

src/main/java/com/fasterxml/jackson/core/io/NumberInput.java
```diff
@@ -190,14 +190,55 @@ public static long parseLong19(char[] ch, int off, boolean negative)
...
+        int offset = 1;
+        // must have 1 - 19 digits after optional sign:
+        // negative?
+        if (neg) {
+            if (len == 1 || len > 20) {
+                return Long.parseLong(s);
+            }
+            c = s.charAt(offset++);
+        } else {
+            if (len > 19) {
```
marschall, line 209: Wouldn't this always throw `NumberFormatException`? Did you mean

    if (len >= 19) {

src/main/java/com/fasterxml/jackson/core/io/NumberInput.java
```diff
@@ -190,14 +190,55 @@ public static long parseLong19(char[] ch, int off, boolean negative)
...
+         * otherwise just defer to JDK parse functionality.
+         */
+        char c = s.charAt(0);
+        int len = s.length();
+        boolean neg = (c == '-');
+        int offset = 1;
+        // must have 1 - 19 digits after optional sign:
+        // negative?
+        if (neg) {
+            if (len == 1 || len > 20) {
```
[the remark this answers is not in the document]
pjfanning: * the logic is taken from NumberInput.parseInt with the numbers bounds ramped up to match long bounds
* I guess the logic for both parseInt and parseLong could be changed to not call the JDK parse code in as many places
* in the end, it is real edge cases, where the failover to JDK parsing happens so the overall perf impact is low

src/main/java/com/fasterxml/jackson/core/io/NumberInput.java
```diff
@@ -190,14 +190,55 @@ public static long parseLong19(char[] ch, int off, boolean negative)
...
+                }
+                num = (num * 10) + (c - '0');
+                // Let's just loop if we have more than 3 digits:
+                if (offset < len) {
+                    do {
+                        c = s.charAt(offset++);
+                        if (c > '9' || c < '0') {
+                            return Long.parseLong(s);
+                        }
+                        num = (num * 10) + (c - '0');
```
marschall, line 236: I haven't understood all the callers. How do you make sure you don't overflow on something like 9223372036854775808?

src/main/java/com/fasterxml/jackson/core/io/NumberInput.java
```diff
@@ -190,14 +190,55 @@ public static long parseLong19(char[] ch, int off, boolean negative)
...
+            c = s.charAt(offset++);
+        } else {
+            if (len > 19) {
+                return Long.parseLong(s);
+            }
+        }
+        if (c > '9' || c < '0') {
+            return Long.parseLong(s);
+        }
+        long num = c - '0';
```
marschall, line 216: I assume this is taken from the existing `#parseInt` method. I could not find any performance advantage over a simple `for` loop, see https://github.com/pjfanning/jackson-number-parse-bench/pull/4/

src/main/java/com/fasterxml/jackson/core/io/NumberInput.java
pjfanning: this overflows with the new code so this is a bug - I've created https://github.com/FasterXML/jackson-core/pull/879 to extend test coverage

If you'd like to create a PR with the improved code you added to the benchmark project, that would be great - I'll close this PR.

Issue #882, opened 2022-12-29
Allow TokenFIlter to skip last elements in arrays


pgomulka:
When the last element in array is an array or object and that element is skipped, the `FilteringParserDelegate` will end up in a loop from which it cannot exit. This means that the rest of the input will be skipped too. This results in incorrect JSON.

This behaviour exists since 2.9+
I believe this is due to https://github.com/FasterXML/jackson-core/commit/7db467ddec7c2899038249b55695b7e44c7b5c3e#diff-f6642caef61e0c403f51a6150ecf45263034fca5002782fd02eacd01e53fe549L694 where the `if (gotEnd)` conditions where removed.
I think this should be added as currently the logic is:
```
 boolean gotEnd = (_headContext == buffRoot);
                    boolean returnEnd = gotEnd && _headContext.isStartHandled();

                    _headContext = _headContext.getParent();
                    _itemFilter = _headContext.getFilter();

                    if (returnEnd) {
                        return t;
                    }
```
and that means that it can only exit when `_headContext.isStartHandled()` is true. For skipped elements this is false.

This can be easily reproduced with this testcase
```
 @Test
    public void testCustomIncludesWithMultipleObjectsInArrayMissLast() throws Exception {
        var factory = new JsonFactory();
        var baseParser = factory.createParser("{\"foo\":[{\"bar\":\"baz\"},{\"bing\":\"boom\"}]}");
        var filteredParser = getFilteredParser(baseParser, filter("foo", "bar"));
        var writer = new StringWriter();
        var generator = factory.createGenerator(writer);
        Assertions.assertTrue(filteredParser.nextToken().isStructStart());
        generator.copyCurrentStructure(filteredParser);
        generator.flush();
        Assertions.assertEquals("{\"foo\":[{\"bar\":\"baz\"}]}", writer.toString());
        //Expected :{"foo":[{"bar":"baz"}]}
        //Actual   :{"foo":[{"bar":"baz"} 
    }
```
cc @tvernum who coauthored the fix and the testcases

cowtowncoder:
Sounds reasonable. I'd be happy to merge, but need CLA first: will add a note on PR.

pgomulka:
@cowtowncoder should we close this issue manually? or will the team close this once the release is done?

cowtowncoder:
No, I need to close it since it didn't automatically do so. Closing done when fix merged for specific branch.

Thanks!

pgomulka:
[...]

PR #883, opened 2022-12-29
Allow FilteringParserDelegate to skip last elements in array

pgomulka:
FilteringParserDelegate when parsing ID_END_ARRAY or ID_END_OBJECT of the skipped element (when `_headContext.isStartHandled() == false` should be able to exit the loop and continue parsing next tokens. 

This logic was changed accidentally removed in https://github.com/FasterXML/jackson-core/commit/7db467ddec7c2899038249b55695b7e44c7b5c3e#diff-f6642caef61e0c403f51a6150ecf45263034fca5002782fd02eacd01e53fe549L694

closes https://github.com/FasterXML/jackson-core/issues/882

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/filter/FilteringParserDelegate.java
+++ b/src/main/java/com/fasterxml/jackson/core/filter/FilteringParserDelegate.java
@@ -754,6 +754,11 @@ public class FilteringParserDelegate extends JsonParserDelegate
                     if (returnEnd) {
                         return t;
                     }
+                    if (gotEnd /*|| (_headContext == buffRoot)
+                    this makes BasicParserFilteringTest.testNotAllowMultipleMatchesWithPath4 fail
+                    */) {
+                        return null;
+                    }
                 }
                 continue main_loop;
             case ID_END_OBJECT:
@@ -780,6 +785,11 @@ public class FilteringParserDelegate extends JsonParserDelegate
 
                 if (returnEnd) {
                     return t;
+                }
+                if (gotEnd /*|| (_headContext == buffRoot)
+                    this makes BasicParserFilteringTest.testNotAllowMultipleMatchesWithPath4 fail
+                */) {
+                    return null;
                 }
             }
             continue main_loop;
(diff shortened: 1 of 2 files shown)


cowtowncoder:
Ok, one quick question first: would it make sense to rebase against `2.14` instead of `2.15`, or would this be too risky a change? It seems like it might be safe for a patch, and that way would be released earlier (2.15.0 is still bit out).
I am fine either way.

Also: before merging I'd need CLA (unless I have received one before -- apologies if I missed it).

CLA is from here: https://github.com/FasterXML/jackson/blob/master/contributor-agreement.pdf
(and there is also alternative Corporate CLA that is sometimes used although most contributions are with individual CLA).

The usual way is to print the doc, fill & sign, scan/photo, email to `info` at fasterxml dot com.
Once I receive it I can merge this to whichever branch we want (I can also merge it to 2.15 and cherry-pick in 2.14 if necessary).

tvernum:
@cowtowncoder I believe you have a corporate CLA for Elastic.
I'll arrange to have @pgomulka and I added to schedule A.

pgomulka:
> Ok, one quick question first: would it make sense to rebase against 2.14 instead of 2.15, or would this be too risky a change? It seems like it might be safe for a patch, and that way would be released earlier (2.15.0 is still bit out).

@cowtowncoder I can either rebase this PR against 2.14 and forward-port it to 2.15
or we could backport this commit to 2.14 after it is merged to 2.15. 

What is the preferred way in fasterxml project?

pjfanning:
target 2.14 and then we'll merge to 2.15 and master

cowtowncoder:
> @cowtowncoder I believe you have a corporate CLA for Elastic. I'll arrange to have @pgomulka and I added to schedule A.

I had a feeling I should have CCLA for Elastic but for some reason I don't seem to be able to locate it.
If you could send an updated version (with schedule A list) I'll make sure to check this into repository.

cowtowncoder (approved):

cowtowncoder:
Given that there should be CLA in effect & there will be updated one sent, I will merge this now.

PR #892, opened 2023-01-12
Feature Request - allow all array elements in JsonPointerBasedFilter

DemonicTutor:


src/main/java/com/fasterxml/jackson/core/filter/JsonPointerBasedFilter.java
```diff
@@ -13,25 +13,40 @@
...
     public JsonPointerBasedFilter(String ptrExpr) {
-        this(JsonPointer.compile(ptrExpr));
+        this(JsonPointer.compile(ptrExpr), false);
     }
 
     public JsonPointerBasedFilter(JsonPointer match) {
+        this(match, false);
+    }
+
+    public JsonPointerBasedFilter(String ptrExpr, boolean includeAllElements) {
```
cowtowncoder, line 26: Would it be possible to add Javadocs here, for the new argument.

src/main/java/com/fasterxml/jackson/core/filter/JsonPointerBasedFilter.java
```diff
@@ -13,25 +13,40 @@
...
+    public JsonPointerBasedFilter(JsonPointer match, boolean includeAllElements) {
         _pathToMatch = match;
+        _includeAllElements = includeAllElements;
     }
 
     @Override
     public TokenFilter includeElement(int index) {
-        JsonPointer next = _pathToMatch.matchElement(index);
+        JsonPointer next;
+        if (_includeAllElements && -1 == _pathToMatch.getMatchingIndex()) {
```
cowtowncoder, line 38: Use `!_pathToMatch.mayMatchElement()` instead of checking `getMatchinIndex()` (same thing just bit more readable)

cowtowncoder (approved):

src/main/java/com/fasterxml/jackson/core/filter/JsonPointerBasedFilter.java
```diff
@@ -13,25 +13,40 @@
 public class JsonPointerBasedFilter extends TokenFilter
 {
     protected final JsonPointer _pathToMatch;
+    protected final boolean _includeAllElements;
```
cowtowncoder, line 16: Javadoc here (I realize there's lack of javadocs in this class prior to changes but.. hopefully can improve)... or
maybe the logic wrt "include all elements" needs to actually go in class Javadoc?

cowtowncoder:
Ok, I realized that I don't quite know what is the intended logic here, so would it be possible to add a brief description?
Looks like there isn't much testing for this part (although since this seems to be based off the other PR, I may be missing something; may be easier to see once I merge #891), so I think tests would be useful (it looked like one of tests did exercise new code but most not).

cowtowncoder:
@DemonicTutor I could proceed merging this next, if you have a chance to look at my suggestions. Also need to resolve the test conflict (merge from 2.15 I think)

YASH-3999:
/
[Data.txt](https://github.com/FasterXML/jackson-core/files/11183112/Data.txt)
how to work in hts java please send code

PR #899, opened 2023-01-27
make float and double parsing lazy

pjfanning:
* matches what we do for BigDecimal/BigInteger
* also helps with cases where we might initially choose wrong number type and then need a different type later
  * for instance, if we initially choose 'double' and then need a BigDecimal - in the past, we would have done a lossy conversion to double - so the BigDecimal could have a rounding issue 
  * leaving the underlying value as a string, means that in he example above, the BigDecimal can be parsed from the original text and no double parsing will happen

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/base/ParserBase.java
+++ b/src/main/java/com/fasterxml/jackson/core/base/ParserBase.java
@@ -929,7 +929,7 @@ public abstract class ParserBase extends ParserMinimalBase
         return _numberInt;
     }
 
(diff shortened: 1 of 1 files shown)


cowtowncoder (approved):

cowtowncoder:
Cool, thank you @pjfanning ! I'll try to get this merged to `master`, see if I have any trouble.

cowtowncoder:
No problems merging, straight-forward! Will now ensure dataformat modules also show no issues for 2.15/3.0.

PR #903, opened 2023-01-28
Implement JsonParser.getNumberValueDeferred(), add tests

cowtowncoder:
So here is my suggested alternative for #893. Instead of adding wrappers here, use bit of polymorphism and let caller handle distinctions: in case of databind `TokenBuffer` I think it works fine and allows deferring decoding which is beneficial for 2 things:

1. Deferring decision of "right" floating-point type to use (`BigDecimal` vs `Double`/`Float`) until we know what that is
2. Avoiding decoding altogether in case number value is to be skipped (case for `@JsonCreator` handling, esp. affects Kotlin and Scala modules), only holding on to `String` until value is actually needed

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/JsonParser.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonParser.java
@@ -1661,6 +1661,37 @@ public abstract class JsonParser
     }
 
     /**
+     * Method similar to {@link #getNumberValue} but that returns
+     * <b>either</b> same {@link Number} value as {@link #getNumberValue()}
+     * (if already decoded), <b>or</b> {@code String} representation of
+     * as-of-yet undecoded number.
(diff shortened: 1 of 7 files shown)


cowtowncoder:
@pjfanning WDYT?

pjfanning:
@cowtowncoder I just want to sort out the unnecessary and possibly exploitable BigInteger parse that happens for unexpected long integers seen in https://github.com/FasterXML/jackson-module-scala/issues/609

If a TokenBuffer solution can be built on top of this change then I guess it's fine with me.

cowtowncoder:
@pjfanning Yes, the intent is for this to work for TokenBuffer (and if for some reason that didn't work, we'd undo it etc).

I'll go ahead and merge this and hope to have time tomorrow to play with `TokenBuffer` side (but you can work concurrently too of course).

PR #906, opened 2023-01-29
build(deps): bump junit to 5.9.2

yeikel:
It also adds the junit bom and replaces the harcoded versions

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -57,7 +57,6 @@ com.fasterxml.jackson.core.*;version=${project.version}
 
     <!-- for Reproducible Builds -->
     <project.build.outputTimestamp>2022-11-05T20:35:16Z</project.build.outputTimestamp>
-    <junit.version>5.9.2</junit.version>
   </properties>
 
   <!-- Alas, need to include snapshot reference since otherwise can not find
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -57,6 +57,7 @@ com.fasterxml.jackson.core.*;version=${project.version}
 
     <!-- for Reproducible Builds -->
     <project.build.outputTimestamp>2022-11-05T20:35:16Z</project.build.outputTimestamp>
+    <junit.version>5.9.2</junit.version>
```
yeikel, line 60: The property might be an overkill if used only once

cowtowncoder:
Is there some specific benefit here? All units test are written against JUnit 4, would this "just work"?

yeikel:
> Is there some specific benefit here? All units test are written against JUnit 4, would this "just work"?

We are using 5.8.2 even if the tests are written against Junit 4 

And everything should work as expected. Is the CI running the tests currently?

cowtowncoder:
> > Is there some specific benefit here? All units test are written against JUnit 4, would this "just work"?
> 
> We are using 5.8.2 even if the tests are written against Junit 4

Ok that's good. I wouldn't mind starting to convert JUnit tests to JUnit 5, just wasn't sure if it could be done incrementally or not.

> And everything should work as expected. Is the CI running the tests currently?

Yes, CI does run test suite against couple of JDK versions.

I think we can start with this here, and if all goes well can start rolling out to all projects and eventually upgrade in `jackson-base` bom.

cowtowncoder (approved):

PR #911, opened 2023-02-02
Remove ThreadLocals from schubfach code

pjfanning:
In perf testing, they have a small benefit but not worth it for the maintenance overhead.

```
Benchmark                                 Mode  Cnt        Score       Error  Units
SchubfachBench.doubleWithThreadLocal     thrpt    5  1076597.929 ± 26758.597  ops/s
SchubfachBench.doubleWithoutThreadLocal  thrpt    5   997088.673 ± 15197.149  ops/s
SchubfachBench.floatWithThreadLocal      thrpt    5  1395289.534 ±  5535.617  ops/s
SchubfachBench.floatWithoutThreadLocal   thrpt    5  1353461.663 ±  5967.389  ops/s
```

Also removes some unused code with the benefit of removing the char[] initialized when we parse numbers. Only the byte[] remains.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/io/schubfach/DoubleToDecimal.java
+++ b/src/main/java/com/fasterxml/jackson/core/io/schubfach/DoubleToDecimal.java
@@ -22,8 +22,6 @@
 
 package com.fasterxml.jackson.core.io.schubfach;
 
-import java.io.IOException;
-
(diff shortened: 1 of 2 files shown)


cowtowncoder (approved):

cowtowncoder:
Sounds good, I like this.

Remains in 2.14 but that's probably safer all things considered.

PR #921, opened 2023-02-14
Add `JsonFactory.Feature` to disable charset detection

yawkat:


cowtowncoder:
Ok this looks nice and reasonable. But let me think about this for just a bit before merging. :)

Thank you again!

cowtowncoder:
Oh, one idea/suggestion: although `JsonFactory.Feature` is easier to add, making is `JsonReadFeature` would allow per-call changes (at least when used via `ObjectReader`).
In case someone wanted to change this more dynamically.

I don't have strong opinion but this just occurred to me and thought I'd share.

yawkat:
@cowtowncoder i used a JsonFactory.Feature because it was easier in this case. it looks like JsonReadFeature also requires a JsonParser.Feature at the moment, which did not seem appropriate here, the other JsonParser.Features seem more like they operate to all formats (not just bytes). Do you think it should be a JsonParser.Feature as well, or should I figure out something to add a JsonReadFeature without a JsonParser.Feature?

cowtowncoder:
> @cowtowncoder i used a JsonFactory.Feature because it was easier in this case. it looks like JsonReadFeature also requires a JsonParser.Feature at the moment, which did not seem appropriate here, the other JsonParser.Features seem more like they operate to all formats (not just bytes). Do you think it should be a JsonParser.Feature as well, or should I figure out something to add a JsonReadFeature without a JsonParser.Feature?

Good point. Unfortunately coupling for 2.x remains (untangling is not super easy), so let's just leave things the way they are as `JsonFactory.Feature`.
We can reconsider this for 3.0 if there are reasons to change it.

cowtowncoder (approved):

yawkat:
ok thanks!

PR #923, opened 2023-02-20
Add constraint validation for TextBuffer.append

yawkat:
This fixes https://github.com/FasterXML/jackson-dataformats-text/issues/384, the tests now fail in the append call, not just on contentsAsString.

FYI @pjfanning

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/util/TextBuffer.java
+++ b/src/main/java/com/fasterxml/jackson/core/util/TextBuffer.java
@@ -711,11 +711,10 @@ public class TextBuffer
         _resultString = null;
         _resultArray = null;
 
-        validateAppend(1);
-
         // Room in current segment?
         char[] curr = _currentSegment;
         if (_currentSize >= curr.length) {
+            validateAppend(1);
             expand();
             curr = _currentSegment;
         }
@@ -731,8 +730,6 @@ public class TextBuffer
         _resultString = null;
         _resultArray = null;
 
-        validateAppend(len);
-
         // Room in current segment?
         char[] curr = _currentSegment;
         int max = curr.length - _currentSize;
@@ -742,6 +739,9 @@ public class TextBuffer
             _currentSize += len;
             return;
         }
+
+        validateAppend(len);
+
         // No room for all, need to copy part(s):
         if (max > 0) {
             System.arraycopy(c, start, curr, _currentSize, max);
@@ -769,8 +769,6 @@ public class TextBuffer
         _resultString = null;
         _resultArray = null;
 
-        validateAppend(len);
-
         // Room in current segment?
         char[] curr = _currentSegment;
         int max = curr.length - _currentSize;
@@ -779,6 +777,9 @@ public class TextBuffer
             _currentSize += len;
             return;
         }
+
+        validateAppend(len);
+
         // No room for all, need to copy part(s):
         if (max > 0) {
             str.getChars(offset, offset+max, curr, _currentSize);
--- a/src/main/java/com/fasterxml/jackson/core/util/TextBuffer.java
+++ b/src/main/java/com/fasterxml/jackson/core/util/TextBuffer.java
@@ -710,6 +710,9 @@ public class TextBuffer
         }
         _resultString = null;
         _resultArray = null;
+
+        validateAppend(1);
+
         // Room in current segment?
         char[] curr = _currentSegment;
         if (_currentSize >= curr.length) {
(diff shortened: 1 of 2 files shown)


pjfanning (approved):
lgtm - thanks

src/main/java/com/fasterxml/jackson/core/util/TextBuffer.java
```diff
@@ -790,6 +797,15 @@ public void append(String str, int offset, int len)
         } while (len > 0);
     }
 
+    private void validateAppend(int toAppend) {
```
pjfanning, line 800: it may be more efficient to have a protected empty implementation in TextBuffer and to have this implementation as `protected final` in ReadConstrainedTextBuffer. SegmentedStringWriter is the only use case that I can think of for the plain TextBuffer though. Everything else uses ReadConstrainedTextBuffer.

pjfanning: SegmentedStringWriter uses the append(char) method on TextBuffer and this method and its calculations will happen for every character.

yawkat: i think it should inline to the same code either way

pjfanning: raw TextBuffer doesn't actually validate - only the ReadContrained one does. So this PR adds extra processing overhesd in the raw TextBuffer when it is not needed.

yawkat: no i think for raw TextBuffer it will simply inline validateAppend and the empty validateStringLength, detect that newTotalLength is unused, and eliminate the code inside validateAppend.

i'm 99% sure youd see this happen in a benchmark at least. however it's a bit less clear what will happen when there is a polymorphic call site.

pjfanning:
To give some background on why the current code doesn't validate in append methods: I had added checks there but @cowtowncoder was worried about the performance overhead. For instance, when append(char) is used, we will now be checking the length on every char.

The idea was to try to reduce the number of checks by only trying to check when the underlying char array in TextBuffer was increased in size. The current checks in TextBuffer are lax because the first time we increase the underlying char array size above the limit, the calling code may not fill the extra capacity and the resulting string might still end up smaller than the limit.

Is there any way that the append(char) checks could be changed so that they only happen if the underlying char array size is being updated?

yawkat:
the checks in append need to happen every time to get 100% compliance, because the resize op may not coincide with reaching the max string length (i.e. the error would trigger too early or too late). If this is fine it could be done of course.

However is .append(char) really perf-critical? it already does a lot of "duplicate work" like checking the input buffers each time. at least for toml, .append(char) is a relatively rare call (e.g. for escape chars), so imo it's not perf-critical.

pjfanning:
I guess I am suggesting the validateAppend calls could be moved to just about the `do` loops that handle the expand calls.

yawkat:
hmm, i see, so basically the failure would be "too late" but it's fine because between the time the limit is actually hit, and the failure is triggered, there are no additional allocations or anything. that makes sense. i will move them and try to figure out how to fix the test :)

pjfanning:
These are the javadocs for the maxStringSize.
```
         * Sets the maximum string length (in chars or bytes, depending on input context).
         * The default is 1,000,000. This limit is not exact, the limit is applied when we increase
         * internal buffer sizes and an exception will happen at sizes greater than this limit. Some
         * text values that are a little bigger than the limit may be treated as valid but no text
         * values with sizes less than or equal to this limit will be treated as invalid.
```

We are willing to allow strings that are a little above the limit if it allows us to avoid making the size checks too often.

src/main/java/com/fasterxml/jackson/core/util/TextBuffer.java
cowtowncoder: Right, the issue is that there are actually usages of polymorphic nature when `SegmentStringWriter` is used (unbounded). But I think bigger concern for me is actual usage of `append()` methods for validating case.

It does look like there are ideas of how to defer detection, however; and I agree that we do not require 100% byte/char accurate limits but rather order-of-magnitude matching.

yawkat:
@pjfanning wdyt

pjfanning:
@yawkat thanks for moving the validateAppend checks. I think this will eliminate the issue where appending single chars or small arrays/strings will lead to a lot of validation.

src/main/java/com/fasterxml/jackson/core/util/TextBuffer.java
yawkat: since the check is deferred now, i dont think it matters either way.

cowtowncoder (approved):

cowtowncoder:
LGTM. @pjfanning I assume you are ok with this so will merge!

Thank you again @yawkat !

PR #926, opened 2023-02-25
[DRAFT] Deep nesting check

pjfanning:
* a proposal for how we might police the depth f input docs
* if this approach is generally ok, I can add more javadoc and refactor the new test


src/main/java/com/fasterxml/jackson/core/StreamReadConstraints.java
```diff
@@ -12,6 +12,11 @@
 {
     private static final long serialVersionUID = 1L;
 
+    /**
+     * Default setting for maximum depth: see {@link Builder#maxDepth(int)} for details.
+     */
+    public static final int DEFAULT_MAX_DEPTH = 1000;
```
cowtowncoder, line 18: Yes, I was thinking of 1_000 as the max as well. Reading/writing may default to same value.

src/main/java/com/fasterxml/jackson/core/StreamReadConstraints.java
```diff
@@ -22,16 +27,37 @@
      */
     public static final int DEFAULT_MAX_STRING_LEN = 1_000_000;
 
+    protected final int _maxDepth;
```
cowtowncoder, line 30: I suspect we need separate settings for reading and writing; this is for reading.

src/main/java/com/fasterxml/jackson/core/base/ParserBase.java
```diff
@@ -1547,4 +1549,19 @@ protected void loadMoreGuaranteed() throws IOException {
 
     // Can't declare as deprecated, for now, but shouldn't be needed
     protected void _finishString() throws IOException { }
+
+    protected final void createChildArrayContext(final int lineNr, final int colNr) {
+        _streamReadConstraints.validateDepth(++_depth);
```
cowtowncoder, line 1554: Would be good to measure performance impact; if there is some, we could instead read max depth into local setting, inline check (that is, method local to parser).

cowtowncoder:
Ah. So far I had been thinking of storing the depth layer in actual parsing context object (since it's always +1 of parent).
But it is true that we could instead just hold on to dynamic nesting in parser/generator, and avoid extra storage.
There is slightly higher risk of missed updates (everywhere where child context created/parent referenced we MUST update depth) but maybe it's worth it.

I think this approach makes sense; added some comments.

src/main/java/com/fasterxml/jackson/core/StreamReadConstraints.java
pjfanning: This is StreamReadConstraints. We'll eventually need a StreamWriteConstraints but I think that should be done later.

src/main/java/com/fasterxml/jackson/core/base/ParserBase.java
pjfanning: yes - I'll set up some benchmark testing

src/main/java/com/fasterxml/jackson/core/StreamReadConstraints.java
cowtowncoder: Ah yes. I realized that last night, right you are.

src/main/java/com/fasterxml/jackson/core/StreamReadConstraints.java
```diff
@@ -22,16 +27,37 @@
...
+         * `{` and `[` respectively.
+         *
+         * @param maxDepth the maximum depth
+         *
+         * @return this builder
+         * @throws IllegalArgumentException if the maxDepth is set to a negative value
+         *
+         * @since 2.15
+         */
+        public Builder maxDepth(final int maxDepth) {
```
cowtowncoder, line 53: `maxDepth` could work; another possibility `maxNesting`. Not sure what terminology is most commonly used with JSON parsers.

pjfanning: I've changed this to maxNestingDepth

pjfanning:
@cowtowncoder I've done a little benchmark testing using https://github.com/pjfanning/jackson-nest-bench - the results are a little erratic. I don't really have dedicated hardware for doing accurate benchmark testing. I'll do some more runs.

One side issue is that I think I made mistake with using IllegalStateException in StreamReadConstraints. I used it in the validateString method and am using for consistency on the method for validating the nesting depth. Would it be ok to use a checked exception (one that does extend RuntimeException)? Maybe, JsonParseException. validateString is new in jackson 2.15. Changing the exception type will break some code in jackson-databind and the jackson-dataformat* jars but can be quickly fixed.

cowtowncoder:
> @cowtowncoder I've done a little benchmark testing using https://github.com/pjfanning/jackson-nest-bench - the results are a little erratic. I don't really have dedicated hardware for doing accurate benchmark testing. I'll do some more runs.

Ok. I should probably try building and running on my end too, on my desktop, to get another datapoint.
 
> One side issue is that I think I made mistake with using IllegalStateException in StreamReadConstraints. I used it in the validateString method and am using for consistency on the method for validating the nesting depth. Would it be ok to use a checked exception (one that does extend RuntimeException)? Maybe, JsonParseException. validateString is new in jackson 2.15. Changing the exception type will break some code in jackson-databind and the jackson-dataformat* jars but can be quickly fixed.

Ah yes, I have been thinking about this. I think that `IllegalStateException` is not optimal and we should throw a `JsonParseException` instead, yes.
For all violations.

pjfanning:
Looks like it would be difficult in Jackson v2 to use a checked exception, like JsonParseException, for the string length checks. This is because the exceptions are thrown in TextBuffer class and that class does not have declared exceptions - on methods like getContentsAsString() for instance. Maybe, we can change this in jackson-core v3?

src/main/java/com/fasterxml/jackson/core/base/ParserBase.java
```diff
@@ -1554,12 +1554,12 @@ protected void loadMoreGuaranteed() throws IOException {
     // Can't declare as deprecated, for now, but shouldn't be needed
     protected void _finishString() throws IOException { }
 
-    protected final void createChildArrayContext(final int lineNr, final int colNr) {
+    protected final void createChildArrayContext(final int lineNr, final int colNr) throws JsonParseException {
```
cowtowncoder, line 1557: With `JsonParser`, Jackson 2.x, I have moved mostly to declaring `IOException` so let's do that with this method and everything else -- there were often cases where some methods need to throw `IOException` (ones that may need to load more content) and then there were mismatches. External API of `JsonParser` exposes it anyway so there is no benefit from internal methods from trying to be more selective.

src/main/java/com/fasterxml/jackson/core/exc/StreamReadException.java
```diff
@@ -52,6 +52,11 @@ protected StreamReadException(String msg, JsonLocation loc, Throwable rootCause)
         super(msg, loc, rootCause);
     }
 
+    // @since 2.15
+    protected StreamReadException(String msg) {
```
cowtowncoder, line 56: I will merge this and similar change in `JsonParseException` separate from this PR.

src/main/java/com/fasterxml/jackson/core/base/ParserBase.java
pjfanning: changed

pjfanning:
@cowtowncoder in https://github.com/pjfanning/jackson-nest-bench, there seems no benefit in moving the validate method so that the limit is stored locally (the V2 implemention in jackson-nest-bench). For some reason, there does appear to be a little bit of a slow down in the Reader based JsonParser but not much proportionally. The byte stream parsers seem to not slow down at all.

cowtowncoder:
Ok, one non-trivial concern: not all format implementations extend `ParserBase` -- some use `MinimalParserBase`.
For those, I think it's necessary to duplicate much of handling at layer above.
Not sure if there is any reasonable way around that, just noting the concern.

Good to know performance might not be affected: I should really try building and running `jackson-benchmarks` for 2.15 with and without modifications.

pjfanning:
The parsing context implementations seem quite different - so different that what I have put in ParserBase won't work for the use cases where ParserMinimalBase is used instead. I think I'll end up having to write specific implementations of the validations for each class that extends ParserMinimalBase.

I have moved the `_nestingDepth` to ParserMinimalBase.

cowtowncoder:
@pjfanning Correct, different implementations are needed for various reasons, for better or worse.

One thing I managed to do is to run preliminary checks on `jackson-benchmarks` vs 2.14.2 and 2.15.0-SNAPSHOT (last full runs are from year ago for 2.13.3), and it does not look like performance of 2.15.0-SNAPSHOT before these changes is measurably slower for the benchmark(s) in question. For what that's worth.
But at least it is possible to isolate effects of changes; I can build from this branch/PR locally, see if I spot any immediate differences.

pjfanning:
close in favout of #943

PR #933, opened 2023-03-01
change string len validation to use JsonParseException

pjfanning:
* TextBuffer methods now throw IOExceptions (specifically JsonParseException)
* In some code paths, the code to get numbers calls the code to get the latest charArray or String and parses that, and this can fail with a JsonParseException. This is now caught and NumberFormatExceptions thrown instead.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/io/JsonStringEncoder.java
+++ b/src/main/java/com/fasterxml/jackson/core/io/JsonStringEncoder.java
@@ -1,5 +1,6 @@
 package com.fasterxml.jackson.core.io;
 
+import java.io.IOException;
 import java.util.Arrays;
 
 import com.fasterxml.jackson.core.util.ByteArrayBuilder;
@@ -97,7 +98,12 @@ public final class JsonStringEncoder
                     if (textBuffer == null) {
                         textBuffer = TextBuffer.fromInitial(outputBuffer);
                     }
-                    outputBuffer = textBuffer.finishCurrentSegment();
+                    try {
+                        outputBuffer = textBuffer.finishCurrentSegment();
+                    } catch (IOException e) {
+                        // IOException won't happen here, can only occur when ReadConstrainedTextBuffer is used
+                        throw new IllegalStateException(e);
+                    }
                     outPtr = 0;
                 }
                 outputBuffer[outPtr++] = c;
@@ -123,7 +129,12 @@ public final class JsonStringEncoder
                 if (textBuffer == null) {
                     textBuffer = TextBuffer.fromInitial(outputBuffer);
                 }
-                outputBuffer = textBuffer.finishCurrentSegment();
+                try {
+                    outputBuffer = textBuffer.finishCurrentSegment();
+                } catch (IOException e) {
+                    // IOException won't happen here, can only occur when ReadConstrainedTextBuffer is used
+                    throw new IllegalStateException(e);
+                }
                 int second = length - first;
                 System.arraycopy(qbuf, first, outputBuffer, 0, second);
                 outPtr = second;
@@ -137,7 +148,12 @@ public final class JsonStringEncoder
             return Arrays.copyOfRange(outputBuffer, 0, outPtr);
         }
         textBuffer.setCurrentLength(outPtr);
-        return textBuffer.contentsAsArray();
+        try {
+            return textBuffer.contentsAsArray();
+        } catch (IOException e) {
+            // IOException won't happen here, can only occur when ReadConstrainedTextBuffer is used
+            throw new IllegalStateException(e);
+        }
     }
 
     /**
@@ -178,7 +194,12 @@ public final class JsonStringEncoder
                     if (textBuffer == null) {
                         textBuffer = TextBuffer.fromInitial(outputBuffer);
                     }
-                    outputBuffer = textBuffer.finishCurrentSegment();
+                    try {
+                        outputBuffer = textBuffer.finishCurrentSegment();
+                    } catch (IOException e) {
+                        // IOException won't happen here, can only occur when ReadConstrainedTextBuffer is used
+                        throw new IllegalStateException(e);
+                    }
                     outPtr = 0;
                 }
                 outputBuffer[outPtr++] = c;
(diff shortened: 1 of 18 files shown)


cowtowncoder:
Ok, I think I have a counter-proposal here: as per #935 (part of which this PR actually adds), I think we can actually eliminate length check from `contentAsString()` by ensuring checks are done in other places where buffer is actually modified.

I also do not like the idea of exposing `IllegalStateException` but would rather that `JsonParseException` was exposed, if at all possible. I am not as worried about cost (although there is some definitely) of this exceptional case, but rather due to combination of checked vs unchecked exceptions.
Fortunately I think that `JsonParser` does expose `IOException` in its API, even if `TextBuffer` doesn't yet. There's definitely possibility of backwards-compatibility issues wrt `TextBuffer` for source compatibility... but I _think_ exceptions not being part of signature means there's no binary incompatibility (oddly enough), if and when we need to start exposing `IOException` / `JsonProcessingException`

pjfanning:
Adding exceptions to methods is API incompatible. It will break existing code that uses TextBuffer APIs.

pjfanning:
removing the validateStringlen calls from contentsAsString breaks existing tests - eg AsyncStringArrayTest.testLongAsciiStringsSmallLimit:153

I'm pretty sure, it will break tests in other Jackson modules too.

cowtowncoder:
@pjfanning yes, unless checks are added in places where I suggest: the idea is to essentially just ensure that by the time we get there, checks have been made already.

cowtowncoder:
Sorry, did minor refactoring, moving `TextBuffer.finishAndReturn()` as it was not next to other related methods (I think it was added to help make CSV module use standard `TextBuffer`).

cowtowncoder:
@pjfanning Ok, I tested this out and could get all but 1 test passing. The reason for last failure is since `append(char)` does not validate length unless buffer gets full.

Thinking about this again I think that I have come to conclusion that maybe check on `contentsAsString()` makes sense after all, and we can leave the combination of checks in place. This assuming it is ok that we expose `JsonParseException` from it.
Put another way I might not want to proceed with #935 after all & current checks make sense.

I do however still want to expose checked exceptions as planned instead of `IllegalStateException`; that seems like less of a surprise for calling code.

src/main/java/com/fasterxml/jackson/core/util/TextBuffer.java
```diff
@@ -958,7 +1079,14 @@ public char[] expandCurrentSegment(int minSize) {
      * {@link #contentsAsString}, since it's not guaranteed that resulting
      * String is cached.
      */
-    @Override public String toString() { return contentsAsString(); }
+    @Override public String toString() {
+        try {
+            return contentsAsString();
+        } catch (IllegalStateException e) {
+            // can fail if the string is too large, use the default Object `toString()` if this happens
+            return super.toString();
```
cowtowncoder, line 1087: Or return constant "ERROR" or whatever. Not 100% sure why I even implemented `toString()` here originally, to be honest. It should not be used for any production thing.

pjfanning: I created https://github.com/FasterXML/jackson-core/pull/938

pjfanning:
I guess v2.15 can be regarded as a relatively major release. As long as we document that the TextBuffer APIs have changed in the release notes (in that they now throw explicit exceptions) - I think we can change this code to drop use of IllegalStateException.

cowtowncoder:
> I guess v2.15 can be regarded as a relatively major release. As long as we document that the TextBuffer APIs have changed in the release notes (in that they now throw explicit exceptions) - I think we can change this code to drop use of IllegalStateException.

Agreed. And I don't think `TextBuffer` is (and I hope it isn't) used by developers outside context of parser implementations.
Well, except possibly for generators but that is covered.

src/main/java/com/fasterxml/jackson/core/io/JsonStringEncoder.java
```diff
@@ -97,7 +98,12 @@ public char[] quoteAsString(String input)
                     if (textBuffer == null) {
                         textBuffer = TextBuffer.fromInitial(outputBuffer);
                     }
-                    outputBuffer = textBuffer.finishCurrentSegment();
+                    try {
+                        outputBuffer = textBuffer.finishCurrentSegment();
+                    } catch (IOException e) {
```
cowtowncoder, line 103: Oh, I see... yeah, this is unfortunate side effect. But it is what it is.

src/main/java/com/fasterxml/jackson/core/io/JsonStringEncoder.java
```diff
@@ -137,7 +148,12 @@ public char[] quoteAsString(String input)
             return Arrays.copyOfRange(outputBuffer, 0, outPtr);
         }
         textBuffer.setCurrentLength(outPtr);
```
cowtowncoder, line 150: I think we can replace combination with 

   return textBuffer.setCurrentAndReturn();

src/main/java/com/fasterxml/jackson/core/io/SegmentedStringWriter.java
```diff
@@ -37,34 +36,84 @@ public Writer append(char c) {
     @Override
     public Writer append(CharSequence csq) {
         String str = csq.toString();
-        _buffer.append(str, 0, str.length());
```
cowtowncoder, line 40: Since basic `Writer` exposes `IOException`, is there a reason we can't do that here? Does caller not (yet) expose them?

cowtowncoder: Quick look suggests that we can just throw `IOException` since it's `ObjectMapper` and `ObjectWriter` that use it but only like regular `Writer` (at least during write path).

cowtowncoder:
Ok, cool, getting there. I hope to work on this extensively tomorrow -- I might apply some of changes from here in a separate PR to make it easier to merge in `master`.

src/main/java/com/fasterxml/jackson/core/io/SegmentedStringWriter.java
pjfanning: changed

cowtowncoder (approved):

cowtowncoder:
Looks good. I will merge some of these changes separately first, before merging, to ease merge into `master` now.

cowtowncoder (approved):

cowtowncoder:
Merged PR, also into `master`. Wasn't too bad after all.

PR #936, opened 2023-03-02
for a PR, only 1 concurrent run (CIFuzz job)

pjfanning:
* job is slow
* this is copied from jackson-module-scala
* ensures for a PR that if you do a new commit and the CIFuzz job is running for the last commit, that gets cancelled - the new run for the new commit will proceed


cowtowncoder (approved):

cowtowncoder:
Thanks! I will do the same for databind

PR #939, opened 2023-03-04
Add `StreamConstraintsException`, use for text token length checks

cowtowncoder:


cowtowncoder:
cc @pjfanning -- will first merge to master, then see if number-length checks should migrate as well.

Issue #937, opened 2023-03-02
Add `StreamConstraintsException` to indicate `StreamReadConstraints` violation


cowtowncoder:
With addition of constraints checking, failures, for length of input tokens and so on, there is need to throw exceptions.
Although it seems best to use `JsonProcessingException`, possibly `JsonParseException`, there is the question of whether we should rather use a specific subtype to allow more generic handling, and/or inclusion of additional information.

But although we might at first consider sub-class of `JsonParseException`, it's good to consider the likelihood of getting `StreamWriteConstraints` checks in near future too. Given this, I think we have 2 different paths:

1. Create new type as sibling of `JsonParseException`, like `JsonConstraintsException` -- usable for reading and writing
2. Create 2 new types -- one for reading, another for writing -- as sub-type, like `JsonReadConstraintsException`

Either way, I think addition of a new exception type would make sense, and should be done ASAP for 2.15.

@pjfanning WDYT?

pjfanning:
Maybe easier to just have one JsonConstraintsException. Users will know the read or write aspect from their own context.

cowtowncoder:
Yes, one exception would seem better. My concern is not that they wouldn't know direction but more about 2 other things:

1. No way to pass `JsonParser` (... although we may not have access to it) since that requires `JsonParseException` (or its supertype StreamReadException) -- although I guess we coudl add generic "processor" field if we wanted to expose it
2. If someone catch just `JsonParseException` this woudl be surprising. But then again, we always expose `JsonProcessingException` (or `IOException`) anyway

So yeah. I think `JsonConstraintsException` makes sense. It being separate type is more of a feature than bug, not a "regular" parsing issue.

cowtowncoder:
Ended up using name `StreamConstraintsException`: trying to (over time) move away from using "Json" in names where applicable to all formats.

PR #940, opened 2023-03-04
Make "too-long-number" use new StreamConstraintsException too

cowtowncoder:
More work for #937: make all validation fails use `StreamConstraintsException`


cowtowncoder:
cc @pjfanning

PR #943, opened 2023-03-05
apply constraint to nesting depth

pjfanning:
Basically #926 but depth is tracked on the JsonStreamContext instead. I have built a jackson-databind enhancement based on this.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/StreamReadConstraints.java
+++ b/src/main/java/com/fasterxml/jackson/core/StreamReadConstraints.java
@@ -15,6 +15,11 @@ public class StreamReadConstraints
     private static final long serialVersionUID = 1L;
 
     /**
+     * Default setting for maximum depth: see {@link Builder#maxNestingDepth(int)} for details.
+     */
+    public static final int DEFAULT_MAX_DEPTH = 1000;
+
+    /**
      * Default setting for maximum number length: see {@link Builder#maxNumberLength(int)} for details.
      */
     public static final int DEFAULT_MAX_NUM_LEN = 1000;
@@ -24,15 +29,36 @@ public class StreamReadConstraints
      */
     public static final int DEFAULT_MAX_STRING_LEN = 1_000_000;
 
+    protected final int _maxNestingDepth;
     protected final int _maxNumLen;
     protected final int _maxStringLen;
 
     private static final StreamReadConstraints DEFAULT =
-        new StreamReadConstraints(DEFAULT_MAX_NUM_LEN, DEFAULT_MAX_STRING_LEN);
+        new StreamReadConstraints(DEFAULT_MAX_DEPTH, DEFAULT_MAX_NUM_LEN, DEFAULT_MAX_STRING_LEN);
 
     public static final class Builder {
+        private int maxNestingDepth;
         private int maxNumLen;
         private int maxStringLen;
+
+        /**
+         * Sets the maximum nesting depth. The depth is a count of objects and arrays that have not
+         * been closed, `{` and `[` respectively.
+         *
+         * @param maxNestingDepth the maximum depth
+         *
+         * @return this builder
+         * @throws IllegalArgumentException if the maxNestingDepth is set to a negative value
+         *
+         * @since 2.15
+         */
+        public Builder maxNestingDepth(final int maxNestingDepth) {
+            if (maxNestingDepth < 0) {
+                throw new IllegalArgumentException("Cannot set maxNestingDepth to a negative value");
+            }
+            this.maxNestingDepth = maxNestingDepth;
+            return this;
+        }
 
         /**
          * Sets the maximum number length (in chars or bytes, depending on input context).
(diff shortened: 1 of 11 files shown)


cowtowncoder:
Yes; this is along lines of what I was originally thinking. I think there are some benefits in avoiding possibility of accidentally forgotten "closing" of nesting levels, to avoid bugs,

src/main/java/com/fasterxml/jackson/core/StreamReadConstraints.java
```diff
@@ -212,4 +251,23 @@ public void validateStringLength(int length) throws StreamConstraintsException
...
+     * is thrown.
+     *
+     * @param depth count of unclosed objects and arrays
+     *
+     * @throws StreamConstraintsException If depth exceeds maximum
+     */
+    public void validateNestingDepth(int depth) throws StreamConstraintsException
+    {
+        if (depth > _maxNestingDepth) {
+            throw new StreamConstraintsException(String.format("Depth (%d) exceeds the maximum allowed nesting depth (%d)",
```
cowtowncoder, line 269: Not related to this PR specifically, but one thing we may want to do afterwards is to add something to note that `StreamReadConstraints` settings control this -- just add some more verbiage in exception message.
But I think that makes sense as a separate PR once we have nesting depth checks covered.

cowtowncoder:
@pjfanning Let me know if and when you are content with the base -- I think this is the last "must-have" feature before 2.15 release candidates.

src/main/java/com/fasterxml/jackson/core/StreamReadConstraints.java
pjfanning: StreamConstraintsException has Javadoc that describes what the exception is about. I'm not sure there is much to be gained by making the message more verbose.

pjfanning:
> @pjfanning Let me know if and when you are content with the base -- I think this is the last "must-have" feature before 2.15 release candidates.

@cowtowncoder I'm happy to get this merged more or less as it is. There is uptake needed in numerous other Jackson modules so it would be good to make a start on that.

src/main/java/com/fasterxml/jackson/core/StreamReadConstraints.java
cowtowncoder: My perspective is that when someone sees a stack trace in log files, it is nice to have slightly more information on context. But I guess minimum level certainly is to have information on exception class itself, pointing to constraints settings.

src/main/java/com/fasterxml/jackson/core/json/JsonReadContext.java
```diff
@@ -61,6 +61,7 @@ public JsonReadContext(JsonReadContext parent, DupDetector dups, int type, int l
         _lineNr = lineNr;
         _columnNr = colNr;
         _index = -1;
+        _nestingDepth = parent == null ? 0 : parent._nestingDepth + 1;
```
cowtowncoder, line 64: After merging, I may add an overload constructor to avoid null check, called from `createXxxContext()`, can mark this as deprecated.

src/test/java/com/fasterxml/jackson/core/read/ArrayParsingTest.java
```diff
@@ -106,6 +107,20 @@ public void testNotMissingValueByEnablingFeature() throws Exception
         _testNotMissingValueByEnablingFeature(false);
     }
 
+    public void testDeepNesting() throws Exception
```
cowtowncoder, line 110: Works for now: in future we probably want to create `...jacksonc.core.limits` (or `constraints`) package or something for (read) constraints

cowtowncoder (approved):

cowtowncoder:
@pjfanning Merged into 2.15, up to master -- will make some smaller changes (increase test coverage, add that one constructor).

Note: `master` will require quite a bit more work other formats as array/object context creation has moved from `ParserBase` (used by non-JSON backends) into `JsonParserBase` -- this to support custom context subtypes better.

src/main/java/com/fasterxml/jackson/core/json/JsonReadContext.java
cowtowncoder: Done.

wolfc:
It looks like the CVE makes a bad assertion. How can the tokenizer run into a StackOverflowError?

cowtowncoder:
@wolfc It can't, this does require use of databind (or similar) to read. But since the fix is in streaming level (prevent reading of highly nested content), so is CVE. Put another way, to fix the issue that is triggered via `jackson-databind`, one has to upgrade `jackson-core` (and format-specific backend component for non-JSON content).

So yes, it is possible to use just `JsonParser` (and `jackson-core` in general) without any problems.

wolfc:
So technically it is a feature introduced in `jackson-core` to counter the CVE in `jackson-databind`.
Ideally the depth check should have been a cross-cutting concern on all `JsonDeserializer`s. Looking at branch 2.12 I found only `UntypedObjectDeserializer` having it. (Don't ask why 2.12 :smile: ) I can see why it was not beneficial to do so.
https://github.com/wolfc/jackson-databind/commit/1bb211e2ed16a3e206473282806069a8b94cda0d

cowtowncoder:
@wolfc Yeah, adding depth checks for every deserializer (...and serializer for another limit) would be lots of overhead on writing & maintaining code, as well as for processing overhead.
Plus difficult to get full coverage anyway, esp. with 3rd party datatype libraries.
So doing it at lower level has many benefits.

Nesting checks for `UntypedObjectDeserializer` (`java.lang.Object`) and `JsonNodeDeserializer` (or whatever class name is; for `JsonNode`, added after 2.12 I think) pre-date streaming-level checks and are sort of redundant by now.

PR #945, opened 2023-03-06
remove outdated javadoc

pjfanning:
InternCache does not use LinkedHashMap


cowtowncoder (approved):
Yeah. Could change to actual impl; but given that this is an internal thing, not part of public API, removal of comment makes sense.

PR #947, opened 2023-03-07
make Parser StreamReadConstraints more null safe

pjfanning:
I was testing the TOML code and ran into trouble with TomlMappers created from TomlFactories - in some tests, the Parser StreamReadConstraints was null and it was causing tests to fail.

This code seems a little safer.


cowtowncoder:
Hmmh. I think this is sort of more a bug in TOML factory tests but I guess that's ok, will merge.

PR #948, opened 2023-03-14
use checked exception in canonicalizer

pjfanning:
due to https://oss-fuzz.com/testcase-detail/6531057062641664

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/sym/ByteQuadsCanonicalizer.java
+++ b/src/main/java/com/fasterxml/jackson/core/sym/ByteQuadsCanonicalizer.java
@@ -4,6 +4,7 @@ import java.util.Arrays;
 import java.util.concurrent.atomic.AtomicReference;
 
 import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.exc.StreamConstraintsException;
 import com.fasterxml.jackson.core.util.InternCache;
 
 /**
@@ -863,7 +864,13 @@ public final class ByteQuadsCanonicalizer
     /**********************************************************
      */
 
-    public String addName(String name, int q1) {
+    /**
+     * @param name
+     * @param q1
+     * @return name (possibly interned)
+     * @throws StreamConstraintsException if the constraint exceptions
+     */
+    public String addName(String name, int q1) throws StreamConstraintsException {
         _verifySharing();
         if (_intern) {
             name = InternCache.instance.intern(name);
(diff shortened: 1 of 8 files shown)


pjfanning:
If this change is ok for 2.15 and 2.15 already has some small exception throwing changes to the APIs - then maybe a similar change shoul go into the byte canonicalizer.

cowtowncoder:
Definitely should behave the same for byte-backed symbol table too.

One thing to verify: that all 2.15 format backends are aligned -- I think this may be source incompatibility but bytecode compatible (since exceptions are not part of method signature).

One question/thought -- is there any Jackson `IOException` subtype we could/should use?
Perhaps this could be constraints failure (not per-document, globally)?

pjfanning:
StreamConstraintsException? Or a new class with similar definition to that?

cowtowncoder:
@pjfanning Was thinking of `StreamConstraintsException`. While not 100% accurate, has sort of similar domain.

pjfanning:
> @pjfanning Was thinking of `StreamConstraintsException`. While not 100% accurate, has sort of similar domain.

Change made. Also applied similar changes in the ByteQuadsCanonicalizer

src/main/java/com/fasterxml/jackson/core/sym/ByteQuadsCanonicalizer.java
```diff
@@ -961,16 +990,16 @@ public String addName(String name, int[] q, int qlen)
...
-    private void _verifySharing()
+    private void _verifySharing() throws StreamConstraintsException
     {
         if (_hashShared) {
             // 12-Mar-2021, tatu: prevent modifying of "placeholder" and
             //   parent tables
             if (_parent == null) {
                 if (_count == 0) { // root
-                    throw new IllegalStateException("Cannot add names to Root symbol table");
+                    throw new StreamConstraintsException("Cannot add names to Root symbol table");
```
cowtowncoder, line 1000: Ok, I'll change this back, actually, since it's internal sanity check (not something users can trigger). Same for next one.

cowtowncoder (approved):

cowtowncoder:
Thank you @pjfanning! I think this will now resolve one last OSS-Fuzz issue that I had not tackled in 2.x (only 3.0).

mail4csm:
After trying out "2.15.0-rc1" error pops up

java.lang.NoClassDefFoundError: com/fasterxml/jackson/core/exc/StreamConstraintsException

	at com.fasterxml.jackson.dataformat.xml.XmlFactory._createParser(XmlFactory.java:670)
	at com.fasterxml.jackson.dataformat.xml.XmlFactory._createParser(XmlFactory.java:30)
	at com.fasterxml.jackson.core.JsonFactory.createParser(JsonFactory.java:1123)
	at com.fasterxml.jackson.databind.ObjectMapper.readTree(ObjectMapper.java:3138)

cowtowncoder:
@mail4csm You have a version mismatch somewhere: somehow you have older `jackson-core` than `jackson-dataformat-xml` -- minor versions need to match (specifically you need `2.15.0-rc1` of `jackson-core`, not just `jackson-dataformat-xml`).

mail4csm:
I'm using the same version "2.15.0-rc1" for both:
- com.fasterxml.jackson.core:jackson-databind
- com.fasterxml.jackson.dataformat:jackson-dataformat-xml

PR #966, opened 2023-03-24
Fix #965 by adding GMM declaration in shade-processed pom.xml

### LINKED ISSUE #965: 2.15.0-rc1 missing Gradle module metadata marker in pom.xml
Looks like addition of maven shade plugin removes XML comment from `pom.xml` during processing.
Similar problem was recently resolved for Afterburner module.

cowtowncoder:


cowtowncoder:
cc @jjohannes

PR #969, opened 2023-04-01
Address issue found wrt #967 by improving BigDecimalParser

cowtowncoder:
So, looks like there is important performance difference between `BigDecimal.movePointRight()` and `BigDecimal.scaleByPowerOfThen()` -- latter avoiding issues with former.

PR #971, opened 2023-04-04
[DRAFT] try to reduce edge cases where BigInts are parsed as BigDecimals first

pjfanning:
* seems like there are edge cases where BigDecimal conversion to BigInteger is very slow
* this change reduces the chance that a BigInteger is first parsed as a BigDecimal
* doesn't completely disallow this path - just tries to avoid it
* causes #968 failing test to actually fail fast instead of timing out


cowtowncoder:
@pjfanning Would be happy to merge this (after #969 I think), but is this still WIP?

pjfanning:
@cowtowncoder this breaks the parsing of 'e' notation BigInts - I think we'll need to address https://github.com/FasterXML/jackson-core/issues/968#issuecomment-1496648960 instead

PR #972, opened 2023-04-04
avoid converting strings to char arrays

pjfanning:
* it is inefficient to convert strings to char arrays if we don't need to


pjfanning:
looking at BigDecimal code, this change doesn't really help (BigDecimal String constructor converts to a char array anyway)

PR #981, opened 2023-04-06
Add debug output wrt #844

cowtowncoder:
As per notes on #844, adding some debug output to help troubleshoot.


cowtowncoder:
@pnacht Like so?

pnacht:
Yep, that should do it!

PR #983, opened 2023-04-06
try to avoid precision loss

pjfanning:
* see https://github.com/FasterXML/jackson-core/issues/730
* prefer BigDecimal/BigInteger to avoid losing precision

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
@@ -2494,8 +2494,6 @@ public abstract class JsonGenerator
             Number n = p.getNumberValueExact();
             if (n instanceof Float) {
                 writeNumber((Float) n);
-            } else if (n instanceof BigInteger) {
-                writeNumber((BigInteger) n);
             } else if (n instanceof BigDecimal) {
                 final BigDecimal bd = (BigDecimal) n;
                 p.streamReadConstraints().validateBigIntegerScale(bd.scale());
@@ -2643,8 +2641,6 @@ public abstract class JsonGenerator
                 Number n = p.getNumberValueExact();
                 if (n instanceof Float) {
                     writeNumber((Float) n);
-                } else if (n instanceof BigInteger) {
-                    writeNumber((BigInteger) n);
                 } else if (n instanceof BigDecimal) {
                     final BigDecimal bd = (BigDecimal) n;
                     p.streamReadConstraints().validateBigIntegerScale(bd.scale());
--- a/src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
@@ -2482,22 +2482,18 @@ public abstract class JsonGenerator
             NumberType n = p.getNumberType();
             if (n == NumberType.INT) {
                 writeNumber(p.getIntValue());
-            } else if (n == NumberType.BIG_INTEGER) {
-                writeNumber(p.getBigIntegerValue());
             } else {
-                writeNumber(p.getLongValue());
+                writeNumber(p.getBigIntegerValue());
             }
             break;
         }
         case ID_NUMBER_FLOAT:
         {
             NumberType n = p.getNumberType();
-            if (n == NumberType.BIG_DECIMAL) {
-                writeNumber(p.getDecimalValue());
-            } else if (n == NumberType.FLOAT) {
+            if (n == NumberType.FLOAT) {
                 writeNumber(p.getFloatValue());
             } else {
-                writeNumber(p.getDoubleValue());
+                writeNumber(p.getDecimalValue());
             }
             break;
         }
@@ -2627,22 +2623,18 @@ public abstract class JsonGenerator
                 NumberType n = p.getNumberType();
                 if (n == NumberType.INT) {
                     writeNumber(p.getIntValue());
-                } else if (n == NumberType.BIG_INTEGER) {
-                    writeNumber(p.getBigIntegerValue());
                 } else {
-                    writeNumber(p.getLongValue());
+                    writeNumber(p.getBigIntegerValue());
                 }
                 break;
             }
             case ID_NUMBER_FLOAT:
             {
                 NumberType n = p.getNumberType();
-                if (n == NumberType.BIG_DECIMAL) {
-                    writeNumber(p.getDecimalValue());
-                } else if (n == NumberType.FLOAT) {
+                if (n == NumberType.FLOAT) {
                     writeNumber(p.getFloatValue());
                 } else {
-                    writeNumber(p.getDoubleValue());
+                    writeNumber(p.getDecimalValue());
                 }
                 break;
             }
--- a/src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
@@ -2497,7 +2497,7 @@ public abstract class JsonGenerator
             } else if (n instanceof BigDecimal) {
                 writeNumber((BigDecimal) n);
             } else {
-                writeNumber(n.longValue());
+                writeNumber(n.doubleValue());
             }
             break;
         }
@@ -2642,7 +2642,7 @@ public abstract class JsonGenerator
                 } else if (n instanceof BigDecimal) {
                     writeNumber((BigDecimal) n);
                 } else {
-                    writeNumber(n.longValue());
+                    writeNumber(n.doubleValue());
                 }
                 break;
             }
--- a/src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
@@ -2491,11 +2491,17 @@ public abstract class JsonGenerator
         }
         case ID_NUMBER_FLOAT:
         {
-            NumberType n = p.getNumberType();
-            if (n == NumberType.FLOAT) {
-                writeNumber(p.getFloatValue());
+            Number n = p.getNumberValueExact();
+            if (n instanceof Float) {
+                writeNumber((Float) n);
+            } else if (n instanceof BigInteger) {
+                writeNumber((BigInteger) n);
+            } else if (n instanceof BigDecimal) {
+                final BigDecimal bd = (BigDecimal) n;
+                p.streamReadConstraints().validateBigIntegerScale(bd.scale());
+                writeNumber(bd);
             } else {
-                writeNumber(p.getDecimalValue());
+                writeNumber(n.longValue());
             }
             break;
         }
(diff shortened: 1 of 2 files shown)


src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
```diff
@@ -2482,22 +2482,18 @@ public void copyCurrentEvent(JsonParser p) throws IOException
             NumberType n = p.getNumberType();
             if (n == NumberType.INT) {
                 writeNumber(p.getIntValue());
-            } else if (n == NumberType.BIG_INTEGER) {
```
cowtowncoder, line 2485: I don't think this is necessary: `NumberType.LONG` is only indicated if value fits -- and we do not want to "upgrade" long to `BigInteger` unless it is absolutely necessary.

The reason this matters is that buffering is used quite a lot, for Constructor/Factory method creators etc.

src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
```diff
@@ -2482,22 +2482,18 @@ public void copyCurrentEvent(JsonParser p) throws IOException
...
         case ID_NUMBER_FLOAT:
         {
             NumberType n = p.getNumberType();
-            if (n == NumberType.BIG_DECIMAL) {
-                writeNumber(p.getDecimalValue());
-            } else if (n == NumberType.FLOAT) {
+            if (n == NumberType.FLOAT) {
                 writeNumber(p.getFloatValue());
             } else {
-                writeNumber(p.getDoubleValue());
```
cowtowncoder, line 2500: This is the fix, but only need if `StreamReadCapability.EXACT_FLOATS` is false. The reason we want to check that is to prevent unnecessary coercion when there is true 64-bit `double` under the hood.

So I think something like 

    else if (n == NumberType.DOUBLE && p.getReadCapabilities().isEnabled(StreamReadCapability.EXACT_FLOATS)) {
       writeNumber(p.getDoubleValue());
    } else {
       writeNumber(p.getDecimalValue());

might work. Although, maybe even better... `parser.getNumberValueExact()`?

src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
pjfanning: I can remove the int/long change - just did this to keep logic consistent with the decimal change

src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
cowtowncoder: Ah. Whether `getNumberValueExact()` would work or not, it's inconvenient...

pjfanning: I changed the code to use `getNumberValueExact()`

src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
```diff
@@ -2491,13 +2491,17 @@ public void copyCurrentEvent(JsonParser p) throws IOException
...
         {
-            NumberType n = p.getNumberType();
-            if (n == NumberType.BIG_DECIMAL) {
-                writeNumber(p.getDecimalValue());
-            } else if (n == NumberType.FLOAT) {
-                writeNumber(p.getFloatValue());
+            Number n = p.getNumberValueExact();
+            if (n instanceof Float) {
+                writeNumber((Float) n);
+            } else if (n instanceof BigInteger) {
```
cowtowncoder, line 2497: Never happens, it's floating point numbers.

src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
```diff
@@ -2491,13 +2491,17 @@ public void copyCurrentEvent(JsonParser p) throws IOException
...
+                writeNumber((BigInteger) n);
+            } else if (n instanceof BigDecimal) {
+                final BigDecimal bd = (BigDecimal) n;
+                p.streamReadConstraints().validateBigIntegerScale(bd.scale());
```
cowtowncoder, line 2501: .... huh?

pjfanning: in the 'failing' test case, we end up with getNumberExact returning this exact case (a BigDecimal)

src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
cowtowncoder: For better or worse, integer numbers are much easier: their magnitude can be reliably handled. As such we do want to use the most efficient form since that's what we decode.

But for FP there is that `Double` vs `BigDecimal` distinction which is retained for some backends (binary), but not for others (textual).

So int/FP shouldn't be unified in that sense: there are different constraints, optimal ways.

src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
cowtowncoder: Hmmh. Yeah there's something wrong there (diff?) -- there's combination of Int and FP handling.

But I also realized that use of `getNumberValueExact()` might not be the best way unfortunately, returning `Number`, requiring casting. I could try to see if what I suggested above works better.

src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
pjfanning: I've now removed this BigInteger check

src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
cowtowncoder: `BigDecimal` makes sense, that validation is what I don't understand (as we are not converting).

src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
```diff
@@ -2636,13 +2638,15 @@ protected void _copyCurrentContents(JsonParser p) throws IOException
...
-                if (n == NumberType.BIG_DECIMAL) {
-                    writeNumber(p.getDecimalValue());
-                } else if (n == NumberType.FLOAT) {
-                    writeNumber(p.getFloatValue());
+                Number n = p.getNumberValueExact();
+                if (n instanceof Float) {
+                    writeNumber((Float) n);
+                } else if (n instanceof BigDecimal) {
+                    final BigDecimal bd = (BigDecimal) n;
+                    p.streamReadConstraints().validateBigIntegerScale(bd.scale());
```
cowtowncoder, line 2646: Same as above, this validation does not make sense.

src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
```diff
@@ -2636,13 +2638,15 @@ protected void _copyCurrentContents(JsonParser p) throws IOException
...
+                    writeNumber(bd);
                 } else {
-                    writeNumber(p.getDoubleValue());
+                    writeNumber(n.longValue());
```
cowtowncoder, line 2645: Also shouldn't this be "n.doubleValue()"...

src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
pjfanning: ok, I've removed the validation

src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
pjfanning: fixed

cowtowncoder (approved):
LGTM, will merge

cowtowncoder:
This works well for the issue (passing!), for `jackson-databind`, and almost all format backends.

But frustratingly there is ONE new test failure for CBOR (in https://github.com/FasterXML/jackson-dataformats-binary/) for "parser.nextTextValue()". I don't think this is necessarily a problem with change here, but might expose some problem CBORParser state keeping. Interestingly enough seems to be related to `float` type handling...

cowtowncoder:
Ah-ha. I think it's StringRef changes from:

https://github.com/FasterXML/jackson-dataformats-binary/pull/347

that are the root cause; and maybe changes here simply caused different encoding of CBOR when copying.

pjfanning:
The JsonGenerator change means the number in the broken test will be output differently. The CBOR code writes BigDecimals very differently from how it writes doubles/floats. 

If needs be, we could hack CBORGenerator to override the new behaviour and to work more like it did before.

cowtowncoder:
@pjfanning While true that things change, bug I see is almost certainly not due to that: it's more a combination of test code and bug (I think) in CBOR for new (in 2.15) StringRef stuff -- and/or handling of `BigDecimal`.

So: basically change here means that test code in CBORTestBase creates slightly different CBOR Doc -- instead of `0.5` as `double` (64-bit fixed) it is written as `BigDecimal` equivalent (since copy method now plays it safe).
Test itself would pass except that somehow decoder gets confused OR exposes START_ARRAY instead of `BigDecimal`.

here-abarany:
Just so there's a record in this conversation, the cause was the optimized `nextTextValue()` function wasn't set up to handle extension tags for multiple types, most notably arrays for this case. This was a bug in previous versions as well, but wasn't triggered until now due to the specifics of how the test was set up.

A separate question is whether it's beneficial to write a `BigDecimal` representation, at least without a way to disable this behavior, in cases such as passing a `JsonParser` to a `CBORGenerator`. Since this is an extension for the file format, it makes the files less portable when used with other parsers. Other binary formats (such as Smile) may not have the portability issue, though performance and/or file size may be a concern if you are storing files that contain many, many doubles.

cowtowncoder:
I am leaning towards reverting this change, due to problems @here-abarany is pointing out.
I just need to check that `TokenBuffer` from `jackson-databind` deals with things already (which I think it does); and I may want to add a new method (`copyCurrentEventExact(...)`) that would handle accuracy in improved way, for users that want that.
(whether to add `copyCurrentStructureExact()` is another question).

cowtowncoder:
Ah ok: `TokenBuffer` overrides both methods so it does not rely on default implementations.
It also relies more on `JsonParser.getNumberValueDeferred()` that we added to defer decoding for textual values (until true result type we want is needed).

So I think reversion is fine.

pjfanning:
This change fixes an issue. If we revert it, what are the plans for the future revisiting of this issue. I prefer correctness over performance. Do we have any test scenarios that can be looked at? Proof of serious perf issues caused by this change.

cowtowncoder:
@pjfanning There are two parts to this: the original report is quite far removed from the test(s) I added -- so my test is tied to implementation, but does not necessarily prove original issue could not be resolved: we have added more functionality to allow accurate retaining of content. We also have not released the fix yet so it's not a regression in that sense, even if test did recreate problem to fix.

The other part is that like I said we can (and should I think) add a new method -- `copyCurrentEventExact()` -- to be used when maximum accuracy retaining is needed. Existing method can provide its existing implementation.

So while I agree that accuracy is, in general, preferred, I think that at this point change of actual functionality is risky and it is better to add new functionality over changes that risk breaking (in some sense) existing usage.

PR #987, opened 2023-04-10
remove problematic bigint optimisation

pjfanning:


cowtowncoder:
@pjfanning Looks like some unit tests need removal of (now) illegal input too.

pjfanning:
@cowtowncoder the failing test was a new one that I had intended to use to prove we could support the '1e10' format but now, I've changed tack and am no longer allowing it. TLDR I've fixed the test so the PR should build ok.

cowtowncoder (approved):

pjfanning:
Users who need faster Biginteger parsing in jackson 21.5, you should consider https://github.com/FasterXML/jackson-core/pull/851

Issue #1002, opened 2023-04-28
Clarify license implications for those who make custom jackson-core builds


vlsi:
Jackson seems to redistribute fastdoubleparser in the shaded form.
fastdoubleparser's license requires that every redistribution should include the copyright notice and the permissions must be included.

See https://github.com/wrandelshofer/FastDoubleParser/blob/aeeab26365235cc2fbfb68fea2145a4b86a800fd/LICENSE#L12-L13

Unfortunately, jackson fails to comply with the license requirements.

I would suggest either stop using shaded dependency, or include the relevant information into the jackson's license that is already included into the release jar.

Is it really required to shade the parser? If you use a regular dependency, then the users can upgrade it in case of CVEs.

See a corresponding licensing issue at fastdoubleparser side: https://github.com/wrandelshofer/FastDoubleParser/issues/38

vlsi:
@cowtowncoder , it would be nice if you could incorporate it into 2.15.1 as well (e.g. either un-shade fastdoubleparser or include the proper licensing information for all the shaded dependencies).

vlsi:
Looks like jackson uses a custom-licensed implementation: https://github.com/FasterXML/jackson-core/issues/577#issuecomment-1086600093

vlsi:
Even though there was a custom agreement between fastdoubleparser and jackson-core developers, it is not clear what are the exact terms of fastdoubleparser code in jackson-core.

For instance, it is not clear if the agreement was for a single version only or if all fastdoubleparser versions.
It is not clear if others can build jackson-core and still get the same licensing preferences for fastdoubleparser.

From the consumer perspective, it would be so much easier if fastdoubleparser shadowing followed the regular procedure, so jackson-core would mention which software is shaded, mention the incense, and so on.

pjfanning:
@vlsi there is a PR that does just that - https://github.com/FasterXML/jackson-core/pull/1003

wrandelshofer:
@pjfanning Unfortunately PR #1003 is not sufficient. 

You have to add 2 additional sections to the NOTICE file.

Like so:

```
fastfloat, Copyright (c) 2021 The fast_float authors. MIT License.
https://github.com/fastfloat/fast_float
https://github.com/fastfloat/fast_float/blob/dc88f6f882ac7eb8ec3765f633835cb76afa0ac2/LICENSE-MIT

bigint, Copyright 2020 Tim Buktu. 2-clause BSD License.
[...]

PR #1003, opened 2023-04-28
add FastDoubleParser section to NOTICE

pjfanning:
A potential solution to https://github.com/FasterXML/jackson-core/issues/1002

### DIFF
--- a/src/main/resources/META-INF/NOTICE
+++ b/src/main/resources/META-INF/NOTICE
@@ -19,3 +19,11 @@ To find the details that apply to this artifact see the accompanying LICENSE fil
 A list of contributors may be found from CREDITS(-2.x) file, which is included
 in some artifacts (usually source distributions); but is always available
 from the source code management (SCM) system project uses.
+
+## FastDoubleParser
+
+jackson-core bundles a shaded copy of FastDoubleParser <https://github.com/wrandelshofer/FastDoubleParser>.
+That code is available under an MIT license <https://github.com/wrandelshofer/FastDoubleParser/blob/main/LICENSE>
+under the following copyright.
+
+Copyright © 2023 Werner Randelshofer, Switzerland. MIT License.
--- a/src/main/resources/META-INF/FastDoubleParser-LICENSE
+++ b/src/main/resources/META-INF/FastDoubleParser-LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright <YEAR> <COPYRIGHT HOLDER>
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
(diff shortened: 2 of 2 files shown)


src/main/resources/META-INF/NOTICE
```diff
@@ -19,3 +19,11 @@ To find the details that apply to this artifact see the accompanying LICENSE fil
...
 in some artifacts (usually source distributions); but is always available
 from the source code management (SCM) system project uses.
+
+## FastDoubleParser
+
+jackson-core bundles a shaded copy of FastDoubleParser <https://github.com/wrandelshofer/FastDoubleParser>.
+That code is available under an MIT license <https://github.com/wrandelshofer/FastDoubleParser/blob/main/LICENSE>
+under the following copyright.
+
+Copyright © 2023 Werner Randelshofer, Switzerland. MIT License.
```
vlsi, line 29: Techincally speaking, the copyright is `<YEAR> <COPYRIGHT HOLDER>`. The current license does not mention `2023 Werner Randelshofer, Switzerland`

pjfanning: his source headers have this - so I'd prefer to err on the side of caution

Hopefully, Werner will update the LICENSE text to include this copyright too and I will then adjust the FastDoubleParser-LICENSE file that I have created here to match any changes to the one in FastDoubleParser project.

vlsi: In practice, if Werner includes the proper license file, then you'll get it automatically when shading FastDoubleParser. Then it would be easier for you to maintain since you could count on FDP to ship the up-to-date license always.

That is the reason why I ask projects to ship their license texts under `META-INF/`.

vlsi: If you hard-code 2023 here, you'll have a hard time maintaining the proper copyright header 🤷

cowtowncoder: It could have form of 2023+ (or such), does not need to be updated every year.
If adding license straight as part of shading that'd be nice of course but I don't think we have to get super clever here. The intent of license matters, not just specific form.

pjfanning: I intend to update this once we can get changes made in Fastdoubleparser license file. This PR means Jackson is not blocked from making releases. We can put a more permanent solution in place when next fastdoubleparser release is done.

pjfanning:
@cowtowncoder rebased to v2.15 - I can provide a separate one for the v2.14 - the text will need to be a little different there - eg the copyright year

cowtowncoder:
Thank you for doing this @pjfanning. Separate PR for 2.14 would be great as I am planning on releasing 2.14.3 relatively soon (accumulated patches, now that 2.15.x is out). But I can make sure this license clarification gets in as well.

PR #1006, opened 2023-04-30
Fastdoubleparser 0.9.0

pjfanning:
* uptakes the license and notice files in FastDoubleParser 0.9.0 jar
* please don't merge this yet - there is a slightly messy issue because jackson-core has a META-INF/NOTICE and so now has  FastDoubleParser
* this PR works around it by renaming the jackson-core file but using a maven-shade-plugin relocation to rename it back to META-INF/NOTICE after relocating the FastDoubleParser file to META-INF/FastDoubleParser-NOTICE


dependency-reduced-pom.xml
```diff
@@ -0,0 +1,240 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
cowtowncoder, line 1: Is this meant to be included? Usually I see this (or similar) file generated?

cowtowncoder:
Is there specific reason for considering this only for 2.16? Safety wrt build process etc?

dependency-reduced-pom.xml
pjfanning: now removed - should we add this file to '.gitignore' because it gets generated by the build and can easily be added to commits by mistake?

pjfanning:
> Is there specific reason for considering this only for 2.16? Safety wrt build process etc?

fixed - PR now based on 2.15 branch

dependency-reduced-pom.xml
cowtowncoder: @pjfanning Yes, makes sense to add to `.gitignore`

cowtowncoder:
> > Is there specific reason for considering this only for 2.16? Safety wrt build process etc?
> 
> fixed - PR now based on 2.15 branch

Ah yeah I forgot the most obvious reason -- I changed default branch to 2.16.

cowtowncoder:
Will not merge until converted to regular PR. Will also likely need help (or separate PR) for `master` / 3.0 since `pom.xml` changes won't merge due to version conflicts (and merge settings for "keep our" basically, `pom.xml merge=ours`).

pjfanning:
#1010 is for master

cowtowncoder (approved):

cowtowncoder:
I assume this is ready now, merging in 2.15 for 2.15.1

PR #1010, opened 2023-05-03
fastdoubleparser 0.9.0 (master)

pjfanning:
see https://github.com/FasterXML/jackson-core/pull/1006


cowtowncoder (approved):

cowtowncoder:
Thank you for taking care of these @pjfanning !

One question on state of things: is 2.15 now ok (wrt licensing part, new version) for 2.15.1 release? I don't intend to do that over this week/weekend, but just want to know in case something else comes up to speed up planning.

Similarly, does 2.14 need further changes before 2.14.3 release?

pjfanning:
I think these PRs allow 2.14.3 and 2.15.1 releases when they are ready to go. It's possible we'll get more info about extra changes that are needed but we can address those in future releases.

cowtowncoder:
Perfect.

Issue #1014, opened 2023-05-05
Increase default max allowed String value length from 5 megs to 20 megs


cowtowncoder:
(note: possible follow-up to #959)

Based on user/dev feedback wrt 2.15.0 release, there seem to be use cases where JSON String values (note: NOT Document size) exceed 5 million characters. While this is configurable limit it seems best to try to change default to be yet more lenient, given that this is not the likeliest way to try to cause DoS. It is also the case that those who do wish to use more modest sizes will anyway need to lower limits, regardless of defaults we choose.

My initial thinking is to increase this further to 20 million (giving max memory usage of about 40 megs in JVM -- although temporarily up to twice that due to segmented storage), for Jackson 2.15.1.

gdimitrov7:
We hit the same issue in our product. 20M should be much better and less intrusive and I fully support it.

**Q1**: Why there is a limit for the max String value? I mean I understand the other limits - `maxDepth`, `maxNumLen` - there are performance penalties for them. In the same I do not see exponential increase for the String values length. And yes, there can be OOM but most apps limit the input to be processed - for example on HTTP request body length. Example: Our server has 2GiB heap with max request body 20M - it does not matter if we have JSON with: 

- 2000 fields by 10k chars each (~20M)
- or 1 field by 20M chars (~20M)

in both cases it should take 2x40M bytes to process the request and parse the body, meaning that we can process ~25 request in the same (if there is no other mem consumed).

I do not say we have to remove the limit, but can it be much bigger by default? 

**Q2**: Is it possible to read the default max from system property?

>        public static final int DEFAULT_MAX_STRING_LEN =  
>                   Integer.getInteger("com.fasterxml.jackson.core.StreamReadConstraints.maxStringLen", 20_000_000);

It will be much easy to be overwritten by apps that need bigger limit by some valid reasons.

cowtowncoder:
Yeah, String value length limit is sort of inspired by number value limit; but practical effects are quite different. In theory I think that if limits were included from beginning they wouldn't have been problematic; going from unlimited to limit is the issue.

As to system properties: for libraries like Jackson, which may be used by multiple other things, frameworks, concurrently, system-wide settings do not make sense in my opinion (system properties essentially being mutable singletons).

Be that as it may, Jackson does not use system properties for anything and my view is that while I do not object to their use, it should occur via extension modules (although in this particular case that'd be difficult to apply).

gdimitrov7:
[...]

PR #1018, opened 2023-05-13
increase DEFAULT_MAX_STRING_LEN to 20 million

pjfanning:
* relates to https://github.com/FasterXML/jackson-core/issues/1014


cowtowncoder:
Ah, concurrent evolution. :)

Thanks @pjfanning I think we were working at same time; I assume it's ok to close this one.

PR #1019, opened 2023-05-13
Allow override of `StreamReadContraints` default with `overrideDefaultStreamReadConstraints()`

pjfanning:
if this approach is acceptable, I can extend this PR with some tests

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/StreamReadConstraints.java
+++ b/src/main/java/com/fasterxml/jackson/core/StreamReadConstraints.java
@@ -54,8 +54,31 @@ public class StreamReadConstraints
     protected final int _maxNumLen;
     protected final int _maxStringLen;
 
-    private static final StreamReadConstraints DEFAULT =
+    private static StreamReadConstraints DEFAULT =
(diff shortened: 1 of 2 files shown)


cowtowncoder:
As per my other note, I don't want to make API changes in 2.15. Will release 2.15.1 soon; this PR may be left as fallback in case there really is enough push back to require addition (i.e. I can still consider change, 2.15.2 or 2.15.1 makes little difference semantically).

darrenoc3:
@cowtowncoder I am just one person, but consider this my pushback: it's extremely inconvenient for legacy projects that are already handling documents >5MB or greater in production to have no mechanism to configure the default globally. You can see another example in Github where Palantir have had to jump through hoops and use Reflection to override the default in their codebase.

cowtowncoder (approved):

cowtowncoder:
I am coming around to thinking this would make sense for 2.15.2. I'll think about it bit more but will probably merge it.

cowtowncoder:
@pjfanning Is this ready to be merged? Only asking as it's still in Draft stage.

pjfanning:
@cowtowncoder this is ready to merge

cowtowncoder (approved):

PR #1021, opened 2023-05-14
Prevent misuse of `StreamReadConstraints.builder` by handling maximum int value `0` appropriately

JooHyukKim:
### Motivation

Prevent potential API misuse of `StreamReadConstraints.builder` configuration regarding setting the maximum value to zero (`0`) by making the handling of the maximum value more explicit and secure, we eliminate confusion and promote safer usage of the builder methods.

Current behavior to setting max value to 0 should be a whole new separate feature like `RETURN_EMPTY_FOR_ANY_STRING_VALUE`.

### Changes

- Include `maxNestingDepth()`, `maxNumberLength()`, and `maxStringLength()` methods.
- Setting the maximum value to 0 will now apply the **default value** instead.

### Further Note

These changes align with conventions. For example, in the [Apache Commons HTTP Server](https://httpd.apache.org/docs/2.4/mod/mpm_common.html#:~:text=Setting%20this%20value%20to%20zero%20means%20that%20the%20server%20will%20wait%20indefinitely%20until%20all%20remaining%20requests%20have%20been%20fully%20served.), setting the maximum value to zero means the server will wait indefinitely until all remaining requests are served.

### DIFF
--- a/src/test/java/com/fasterxml/jackson/core/StreamReadConstraintsTest.java
+++ b/src/test/java/com/fasterxml/jackson/core/StreamReadConstraintsTest.java
@@ -8,10 +8,8 @@ public class StreamReadConstraintsTest extends BaseTest {
         _verifyDefaultValues(_builderWith(0, 0, 0));
 
         // custom
-        _verifyValues(_builderWith(500, 10, 50),
-                500, 10, 50);
-        _verifyValues(_builderWith(0, 20, 100),
-                StreamReadConstraints.DEFAULT_MAX_STRING_LEN, 20, 100);
+        _verifyValues(_builderWith(500, 10, 50), 500, 10, 50);
+        _verifyValues(_builderWith(0, 20, 100), StreamReadConstraints.DEFAULT_MAX_STRING_LEN, 20, 100);
     }
 
     public void testFailingBuild() {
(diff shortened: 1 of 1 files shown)


src/main/java/com/fasterxml/jackson/core/StreamReadConstraints.java
```diff
@@ -77,8 +77,11 @@ public static final class Builder {
         public Builder maxNestingDepth(final int maxNestingDepth) {
             if (maxNestingDepth < 0) {
                 throw new IllegalArgumentException("Cannot set maxNestingDepth to a negative value");
+            } else if (maxNestingDepth == 0) {
+                this.maxNestingDepth = DEFAULT_MAX_DEPTH;
+            } else {
```
pjfanning, line 82: -1 from me - behaviour change in a patch release

I think this change is unnecessary

If the javadoc is wrong about the existing code, we can the javadoc.

cowtowncoder:
Ok, I agree with @pjfanning: I don't want `0` to be an alias for "use default". Either it should be invalid (but if so, needed to go in 2.16), or it just means that only non-structured values can be output (so START_ARRAY / START_OBJECT would fail).

pjfanning:
In the end, a user could set the max string len to 1. That would be nearly as bad as 0. So I don't really see the need to add extra checks for max = 0.

At some stage, you have to trust that users will do some basic testing and that we can't be overly concerned if users try to set bad configs.

PR #1026, opened 2023-05-17
Feature Request - allow all array elements in JsonPointerBasedFilter - updated

DemonicTutor:
Currently `JsonPointerBasedFilter` only processes Json Arrays by index. eg `/0/id`.

With this change it is possible to extract all elements using a pointer in the form of `//id`: 

source:
```
[
   {
      "id":1,
      "stuff":[
         {
            "name":"first",
            "type":"a"
         },
         {
            "name":"second",
            "type":"b"
         }
      ]
   },
   {
      "id":2,
      "stuff":[
         {
            "name":"third",
            "type":"c"
         },
         {
            "name":"fourth",
            "type":"d"
         }
      ]
   }
]
```

pointers:
```
"//id"
"//stuff//name"
```

result:
```
[
   {
      "id":1,
      "stuff":[
         {
            "name":"first"
         },
         {
            "name":"second"
         }
      ]
   },
   {
      "id":2,
      "stuff":[
         {
            "name":"third"
         },
         {
            "name":"fourth"
         }
      ]
   }
]
```

### DIFF
--- a/src/main/java/tools/jackson/core/filter/JsonPointerBasedFilter.java
+++ b/src/main/java/tools/jackson/core/filter/JsonPointerBasedFilter.java
@@ -13,25 +13,47 @@ import tools.jackson.core.JsonPointer;
 public class JsonPointerBasedFilter extends TokenFilter
 {
     protected final JsonPointer _pathToMatch;
+    /**
+     * if true include all array elements by ignoring the array index match and advancing the JsonPointer to the next level
+     */
+    protected final boolean _includeAllElements;
 
     public JsonPointerBasedFilter(String ptrExpr) {
-        this(JsonPointer.compile(ptrExpr));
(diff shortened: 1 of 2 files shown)


pjfanning:
Any chance you could target at branch 2.16? master branch is for a 3.0 release that is probably a long way away? Jackson 2 uses com.fasterxml.jackson instead of tools.jackson package?

cowtowncoder:
Right, +1 for targeting 2.16 for reasons @pjfanning mentioned. Would be happy to merge (assuming CLA handled)

cowtowncoder:
Checked it: yes, we do have CLA. So that's not a problem.

cowtowncoder:
@DemonicTutor Would be happy to merge but I think we'd really like this against 2.16 branch.

cowtowncoder:
Np @DemonicTutor !

PR #1029, opened 2023-05-18
chore(deps): Update dependencies and plugins using properties

reneleonhardt:
Introduce properties for dependency and plugin versions, and update them to the latest versions.

- animal-sniffer-maven-plugin 1.2 to 1.23
- junit 5.9.2 to 5.9.3
- jacoco-maven-plugin 0.8.8 to 0.8.10
- maven-wrapper 3.1.1 to 3.2.0
- maven 3.9.0 to 3.9.2

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -53,6 +53,14 @@ tools.jackson.core.*;version=${project.version}
 
     <!-- for Reproducible Builds -->
     <project.build.outputTimestamp>2022-11-27T00:00:00Z</project.build.outputTimestamp>
+
+    <!-- dependency and plugin versions -->
+    <animal-sniffer-maven-plugin.version>1.23</animal-sniffer-maven-plugin.version>
+    <fastdoubleparser.version>0.9.0</fastdoubleparser.version>
+    <find-and-replace-maven-plugin.version>1.1.0</find-and-replace-maven-plugin.version>
+    <jacoco-maven-plugin.version>0.8.10</jacoco-maven-plugin.version>
+    <junit-bom.version>5.9.3</junit-bom.version>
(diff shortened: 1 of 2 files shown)


pjfanning:
I'm afraid we have dependabot and central version management in https://github.com/FasterXML/oss-parent

cowtowncoder:
That, plus plug-in versions come from combination of parent poms (oss-parent, jackson-parent, jackson-base, jackson-bom). They are not to be specified locally except temporarily.

pom.xml
```diff
@@ -53,6 +53,14 @@ tools.jackson.core.*;version=${project.version}
 
     <!-- for Reproducible Builds -->
     <project.build.outputTimestamp>2022-11-27T00:00:00Z</project.build.outputTimestamp>
+
+    <!-- dependency and plugin versions -->
```
cowtowncoder, line 57: These are to be upgraded in parent pom:

https://github.com/FasterXML/oss-parent/

so could you please remove?

pom.xml
```diff
@@ -275,7 +285,7 @@ tools.jackson.core.*;version=${project.version}
     <dependency>
       <groupId>ch.randelshofer</groupId>
       <artifactId>fastdoubleparser</artifactId>
-      <version>0.9.0</version>
+      <version>${fastdoubleparser.version}</version>
```
cowtowncoder, line 288: No point in add property just for sake of adding one; will never be overridden or reused.

cowtowncoder:
Thank you for submitting this; as per my notes, closing, will apply mvnw changes separately however & upgrade animal sniffer.

PR #1043, opened 2023-06-11
Make MAX_ERROR_TOKEN_LENGTH configurable via `ErrorReportConfiguration`

JooHyukKim:
as discussed in [google groups page](https://groups.google.com/g/jackson-user/c/4rjcZ3s1pDE/m/rNfUomagAQAJ).

### Motivation

Currently, when Jackson throws an exception, it quotes the JSON that triggered the error. However, there's a limitation: the quoted message is truncated after MAX_ERROR_TOKEN_LENGTH as specified in... https://github.com/FasterXML/jackson-core/blob/832b52068c295b5e57fbec7a67d2934487fec88a/src/main/java/com/fasterxml/jackson/core/base/ParserMinimalBase.java#L132 

.... which makes it difficult to debug issues when the relevant section is further down in the JSON.

This PR elliviates that by introducing ErrorTokenConfiguration configurable via `JsonFactory.Builder` and `JsonFactory.setErrorTokenConfiguration`

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/ErrorReportConfiguration.java
+++ b/src/main/java/com/fasterxml/jackson/core/ErrorReportConfiguration.java
@@ -8,8 +8,8 @@ import java.io.Serializable;
  * <p>
  * Currently default settings are
  * <ul>
- *     <li>Maximum length of token to include in error messages : default 256 (see {@link #_maxErrorTokenLength})
- *     <li>Maximum length of raw content to include in error messages : default 256 (see {@link #_maxRawContentLength})
+ *     <li>Maximum length of token to include in error messages (see {@link #_maxErrorTokenLength})
+ *     <li>Maximum length of raw content to include in error messages (see {@link #_maxRawContentLength})
  * </ul>
  *
  * @since 2.16


src/main/java/com/fasterxml/jackson/core/JsonFactory.java
```diff
@@ -1056,6 +1071,21 @@ public JsonFactory setCodec(ObjectCodec oc) {
...
+    /**
+     * @param maxErrorTokenLength Constraints
+     * @return This factory instance (to allow call chaining)
+     * @throws IllegalArgumentException if {@code maxErrorTokenLength} is less than 0
+     *
+     * @since 2.16
+     */
+    public JsonFactory setMaxErrorTokenLength(int maxErrorTokenLength) {
+        if (maxErrorTokenLength < 0) {
+            throw new IllegalArgumentException("maxErrorTokenLength (" + maxErrorTokenLength + ") must be greater than 0");
```
cowwoc, line 1083: Either the condition or the error message is incorrect. I believe the input value must be non-negative, meaning that zero is also acceptable.

src/main/java/com/fasterxml/jackson/core/io/IOContext.java
```diff
@@ -129,6 +137,30 @@ public IOContext(StreamReadConstraints src, BufferRecycler br,
...
+    public IOContext(StreamReadConstraints src, BufferRecycler br,
+                    ContentReference contentRef, boolean managedResource, int maxErrorTokenLength)
+    {
+        _streamReadConstraints = (src == null) ?
+                StreamReadConstraints.defaults() : src;
+        _bufferRecycler = br;
+        _contentReference = contentRef;
+        _sourceRef = contentRef.getRawContent();
+        _managedResource = managedResource;
+        _maxErrorTokenLength = maxErrorTokenLength < 0 ? DEFAULT_MAX_ERROR_TOKEN_LENGTH :
```
cowwoc, line 162: This behavior should be documented in the Javadoc (i.e. "a negative number to use the the default value")

src/main/java/com/fasterxml/jackson/core/JsonFactory.java
JooHyukKim: @cowwoc Right, thank you for finding out! 🙏🏼 Applied.

src/main/java/com/fasterxml/jackson/core/io/IOContext.java
JooHyukKim: True, true 😄. Applied. Thank you!

cowwoc: Looks good to me

src/main/java/com/fasterxml/jackson/core/JsonFactory.java
cowwoc: Looks good to me

JooHyukKim: Cheers

src/main/java/com/fasterxml/jackson/core/io/IOContext.java
JooHyukKim: Cheers

cowtowncoder:
Quick note: due to number of changes needed for configuration, I think I really want a configuration object to be created to pass this setting, along with others. Similar to `StreamReadConstraints`; immutable. Otherwise next time a size settings is needed same number of changes will be needed.

I don't know what a good name would be, nor if there are other settings that could already be added. But I think adding individual scalars in `JsonFactory` is not very maintainable.

JooHyukKim:
> a configuration object to be created to pass this setting, along with others. Similar to `StreamReadConstraints`; immutable...
> But I think adding individual scalars in JsonFactory is not very maintainable.... 

Totally agree on this 👍🏻👍🏻 So I will come back with an object version(`StreamReadConstraints` as reference) and hopefully some reasonable draft of names 😆

cowwoc:
I would ask @cowtowncoder for review. I'm just an end-user :)

cowtowncoder:
@JooHyukKim Sorry, I should have explained bit further: although the encapsulation as `ErrorTokenConfiguration` is fine, I think its scope is too small as in we won't really have too many aspects of error token configuration. What I'd be looking for is bit more generic; something where we would be adding more settings.
But I don't have a good idea immediately what that would be... something along lines of "DiagnosticsConfiguration" or maybe "ErrorReportingConfiguration". But it is difficult to predict what will be needed.

So I think that at this point I'd probably leave this PR as work-in-progress to see if we can use configuration for other aspects.

cowtowncoder:
@JooHyukKim I like that name -- and yeah, let's let this wait for a bit -- I don't necessarily mind having just one setting as long as it seems plausible we could have more.

And now that you mention this, one thing we could perhaps use it for would be configuring inclusion (or not) of source.
We have added `StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` which probably should remain.
But we could make max length of snippet configurable, perhaps.
OTOH, is this "error" information? I guess it kind of is.

cowtowncoder:
@JooHyukKim Sorry, what I meant to say is that we'd leave the on/off feature as-is (too much hassle moving thing just added IMO, unless for maybe 3.0).
But we could specify length of source to include (if enabled) -- max 500 characters, 1000, etc as a new configuration setting in this config value object.

src/main/java/com/fasterxml/jackson/core/JsonFactory.java
```diff
@@ -374,6 +384,8 @@ protected JsonFactory(JsonFactory src, ObjectCodec codec)
         _generatorDecorators = _copy(src._generatorDecorators);
         _streamReadConstraints = src._streamReadConstraints == null ?
             StreamReadConstraints.defaults() : src._streamReadConstraints;
+        _errorReportConfiguration = src._errorReportConfiguration == null ?
```
cowtowncoder, line 387: Let's not allow `null`s, so remove check, just assign.

Actually should probably remove null checks from `_streamReadConstraints` case as well. We should never have those as `null` anywhere.

src/main/java/com/fasterxml/jackson/core/base/ParserMinimalBase.java
```diff
@@ -117,20 +117,6 @@ public abstract class ParserMinimalBase extends JsonParser
...
-     */
-
-    /**
-     * Maximum number of characters to include in token reported
-     * as part of error messages.
-     *
-     * @since 2.9
-     */
-    protected final static int MAX_ERROR_TOKEN_LENGTH = 256;
-
```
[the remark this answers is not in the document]
cowtowncoder: yeah let's leave, mark as deprecated for now, just in case.

src/main/java/com/fasterxml/jackson/core/io/ContentReference.java
```diff
@@ -98,6 +110,28 @@ protected ContentReference(boolean isContentTextual, Object rawContent,
         _rawContent = rawContent;
         _offset = offset;
         _length = length;
+        ErrorReportConfiguration errorReportConfiguration = ErrorReportConfiguration.defaults();
```
cowtowncoder, line 113: Instead of this assignment, I think constructor should just call new one, with extra argument specified as `ErrorReportConfiguration.defaults()`.

src/main/java/com/fasterxml/jackson/core/io/ContentReference.java
```diff
@@ -131,7 +165,23 @@ public static ContentReference construct(boolean isContentTextual, Object rawCon
 
     public static ContentReference construct(boolean isContentTextual, Object rawContent,
             int offset, int length) {
-        return new ContentReference(isContentTextual, rawContent, offset, length);
+        return new ContentReference(isContentTextual, rawContent, offset, length, null);
```
cowtowncoder, line 168: Pass `ErrorReportConfiguration.defaults()` here instead of `null`

cowtowncoder:
Ok yes this is better. I wish passing new configuration was not so awkward though, it is unfortunate to need to modify both `ContentReference` and (especially) `IOContext`. But perhaps that cannot be helped.

One thing that would help merging would, however, be splitting one set of change to be merged first:

1. Add new `ErrorReportConfiguration` type with builder, unit test(s)
2. Change `TSFBuilder` / `JsonFactory` to allow configuring and holding on to that config object

and then in a follow-up PR do the rest. This will be bit easier to merge cleanly; one problem are being `JsonFactory` that is somehow unlinked (`git merge` thinking it is deleted from `master`).

It would also be bit easier to maybe see if there's any way to minimize changes to `ContentReference` or `IOContext`.

I think easiest way might be to leave this PR as-is, create new "minimal" one from some of changes.

PR #1051, opened 2023-06-14
Support for JsonGeneratorDecorator

digulla:
Support to register decorators for JsonGenerator in the builder and factory.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/JsonFactory.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonFactory.java
@@ -7,6 +7,8 @@ package com.fasterxml.jackson.core;
 import java.io.*;
 import java.lang.ref.SoftReference;
 import java.net.URL;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Objects;
 
 import com.fasterxml.jackson.core.format.InputAccessor;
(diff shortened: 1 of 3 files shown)


cowtowncoder:
Cool, I like this. Let me think a bit about it, just to be sure, but I think this could work about as-is.

cowtowncoder:
Forgot to ask: any chance for a simple test or two, to verify decoration actually works...

src/main/java/com/fasterxml/jackson/core/JsonFactory.java
```diff
@@ -1892,7 +1898,14 @@ protected JsonGenerator _createGenerator(Writer out, IOContext ctxt) throws IOEx
         if (rootSep != DEFAULT_ROOT_VALUE_SEPARATOR) {
             gen.setRootValueSeparator(rootSep);
         }
-        return gen;
+        return _decorate(gen);
+    }
+
+    private JsonGenerator _decorate(JsonGenerator result) {
```
cowtowncoder, line 1904: Should be `protected` to allow overriding (I can change)

cowtowncoder (approved):
Looks good overall; added notes but will address them post-merge.

cowtowncoder:
> One thing that comes to my mind: Is there anything else in the factory which could use this?
> 
> Also, is the `_decorate()` method in the right class? It could go into TSFBuilder instead.

I think it's in the right class -- it could not really go in factory builder (as that is only part of constructing factory itself, not generator).

One bigger thing we should tackle is to change all format backends to allow decoration. In 3.0 (master) this is relatively easy due to full refactoring of TokenStreamFactory hierarchy. But in 2.x there's much more duplication so pretty much every `XxxFactory` needs to call `_decorate()`.

I did at least add tests for `JsonFactory`, fwtw. Changing other backends is not difficult, just lots of work.

PR #1056, opened 2023-06-19
Add StreamWriteConstraints with nested depth check (master branch)

pjfanning:
* relates to #1055 
* the code in master is so different from 2.16, it's virtually impossible to merge changes from 2.16
* I had to write at least half of this again

### DIFF
--- a/src/main/java/tools/jackson/core/JsonGenerator.java
+++ b/src/main/java/tools/jackson/core/JsonGenerator.java
@@ -52,6 +52,19 @@ public abstract class JsonGenerator
 
     /*
     /**********************************************************************
+    /* Constraints violation checking
+    /**********************************************************************
+     */
+
+    /**
+     * Get the constraints to apply when performing streaming writes.
+     */
+    public StreamWriteConstraints streamWriteConstraints() {
+        return StreamWriteConstraints.defaults();
+    }
+
+    /*
+    /**********************************************************************
     /* Public API, output configuration, state access
     /**********************************************************************
      */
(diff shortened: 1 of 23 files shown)


pjfanning:
@cowtowncoder I think the tests are fixed now

src/main/java/tools/jackson/core/JsonGenerator.java
```diff
@@ -50,6 +50,19 @@ protected JsonGenerator() { }
...
+    /**********************************************************************
+    /* Constraints violation checking
+    /**********************************************************************
+     */
+
+    /**
+     * Get the constraints to apply when performing streaming writes.
+     */
+    public StreamWriteConstraints streamWriteConstraints() {
+        return StreamWriteConstraints.defaults();
```
cowtowncoder, line 63: We can probably leave this abstract, similar to how `JsonParser` leaves matching method, to avoid possibility of something not overriding it.
(can't do in 2.x for compatibility reasons etc)

I can do this post-merge, just noting it here for posterity.

src/main/java/tools/jackson/core/io/IOContext.java
```diff
@@ -104,16 +107,21 @@ public class IOContext
...
      */
-    public IOContext(StreamReadConstraints streamReadConstraints, BufferRecycler br,
-            ContentReference contentRef, boolean managedResource,
+    public IOContext(StreamReadConstraints streamReadConstraints,
+            StreamWriteConstraints streamWriteConstraints,
+            BufferRecycler br, ContentReference contentRef, boolean managedResource,
             JsonEncoding enc)
     {
-        _streamReadConstraints = streamReadConstraints;
+        _streamReadConstraints = streamReadConstraints == null ?
```
cowtowncoder, line 121: For 3.0 can also not allow `null`s I think so I'll change to `Objects.nonNull(...)` or such.

cowtowncoder (approved):

Issue #1058, opened 2023-07-11
Private constructor of the ByteArrayBuilder doesn't use the BufferRecycler parameter


eneim:
This is only a question:

I notice the following [(private) constructor](https://github.com/FasterXML/jackson-core/blob/76d0749ff072acb7fdf4a924b80e7e347b192451/src/main/java/com/fasterxml/jackson/core/util/ByteArrayBuilder.java#L65C27-L65C27) of the `ByteArrayBuilder`:

```
private ByteArrayBuilder(BufferRecycler br, byte[] initialBlock, int initialLen) {
    _bufferRecycler = null;
    _currBlock = initialBlock;
    _currBlockPtr = initialLen;
}
```

I want to ask why `_bufferRecycler` is initialized to `null` given the constructor provides a `BufferRecycler`? Is that by intent or a mistake? If it is by intent, can you explain why?

pjfanning:
@eneim thanks for raising this - in practise, the only code that uses this constructor does not have a bufferRecycler anyway but I created #1059 anyway

cowtowncoder:
I'd venture a guess that this is accidental oversight. Thank you @eneim for reporting it.

cowtowncoder:
Fixed via #1059

PR #1059, opened 2023-07-11
constructor param ignored

pjfanning:
* relates to #1058 
* the only usage is from this method which hardcodes bufferRecycler=null
```
    public static ByteArrayBuilder fromInitial(byte[] initialBlock, int length) {
        return new ByteArrayBuilder(null, initialBlock, length);
    }
```
* seems tidier to not ignore to constructor param in case something else starts using the constructor


cowtowncoder (approved):

cowtowncoder:
Yeah I'll merge this.

Issue #919, opened 2023-02-08
Be ready for Virtual Thread: avoid ThreadLocal pooling


franz1981:
Hi team.

this has been raised because, as Jackson users i.e https://github.com/quarkusio/quarkus with existing (experimental) support for https://quarkus.io/guides/virtual-threads, we would like to help fixing any integration issue with such technology.

Based on https://openjdk.org/jeps/425 and https://openjdk.org/jeps/429, re thread local pooling:
> There are a few scenarios that favor thread-local variables. An example is caching objects that are expensive to create and use, such as instances of java.text.DateFormat. Notoriously, a DateFormat object is mutable, so it cannot be shared between threads without synchronization. Giving each thread its own DateFormat object, via a thread-local variable that persists for the lifetime of the thread, is often a practical approach.

In short, https://openjdk.org/jeps/429 warn against using `ThreadLocal`s as pooling pattern, and that `Scoped Values` mechanism won't save the day nor will be able to replace such mechanism.

said that, **where Jackson exibit this problem?** 
Let me attach some data:
![image](https://user-images.githubusercontent.com/13125299/217560540-922bdc5c-8f2b-4535-8d05-2457c44ac49b.png)

In the above flamegraph, a lot of cycles are spent in https://github.com/FasterXML/jackson-core/blob/390472bbdf4722fe058f48bb0eff5865c8d20f73/src/main/java/com/fasterxml/jackson/core/util/BufferRecyclers.java#L67 while called from a virtual thread forces both `ThreadLocal` and subsequent pooled buffers allocations.

The common usage scenario for virtual threads is to NOT cache them (as the Loom team advertise against it) and to be "a lot" and short lived; meaning that Jackson will end up creating tons of useless garbage that won't be reused, defeating the purpose and benefit of pooling.

**What solution exists here?**
1. in case of virtual thread encoding, use a proper pool that's decent enough to deal with high contention, instead of the usual thread local one
2. JDK makes uses of an "hidden" feature for cases like these i.e. [Per-carrier-thread cache](https://github.com/openjdk/jdk/blob/861cc671e2e4904d94f50710be99a511e2f9bb68/src/java.base/share/classes/sun/nio/ch/Util.java#L56)

[...]

PR #1061, opened 2023-07-18
introducing BufferRecyclerPool

mariofusco:
This pull request is a work in progress and at the moment it is **not** intended to be merged as it is. This is mostly to start a discussion and proposing a first tentative fix for the problem reported [here](https://github.com/FasterXML/jackson-core/issues/919). 

In particular this is trying to implement the solution 1. suggested by @franz1981. The solution 2. would have the advantage of being the less invasive for jackson (and would be probably guarantee better performances), but it is totally impossible now and even in a near or far future it is very unlikely that the JDK team will agree to make the per-carrier-thread cache generally available since it would break some of the virtual threads encapsulations.

Also following what suggested in [this comment](https://github.com/FasterXML/jackson-core/issues/919#issuecomment-1425028482) by @cowtowncoder, I'm trying to better define the lifecycle of the `JsonParser` and `JsonGenerator` making their `close` methods to even close the underlying `IOContext`. The `IOContext` is now the only owner of its `BufferRecycler`: instead of being injected with a `BufferRecycler` instance as an argument of its constructor, it borrows the `BufferRecycler` directly from the new `ObjectPool`, and put it back into the pool when it is closed.

This solution of course relies on the fact that the `close` methods of the `JsonParser` and `JsonGenerator` are always called correctly. I haven't found a place when this doesn't happen and in reality I conversely found at least [one situation when this happens twice](https://github.com/FasterXML/jackson-jr/blob/master/jr-objects/src/main/java/tools/jackson/jr/ob/JSON.java#L1466), without any apparent valid reason.

The `ObjectPool` implementation is extremely naive and the main reason why this pull request has to be considered just a draft. However it only has the `borrow` and `offer` methods, so I believe that it could be easily replaced by any real-world and battle tested high performance implementation. To this purpose I'd prefer to avoid reinventing the wheel and reusing one of the object pool implementations already available like the one in JCTools suggested by @franz1981 or one of the many alternatives listed [here](https://medium.com/@chrishantha/benchmarking-object-pools-6df007a31ada), even though not all of them are really an option because some also internally relies on `ThreadLocal`s in some way. 

So here the first questions: is it ok for `jackson-core` to directly depend on one of those libraries providing a production ready object pool or do we need/want to have an implementation directly inside jackson codebase? Do you know any of those libraries/implementations or have a specific preference for any of them? /cc @cowtowncoder @pjfanning 

Finally I also made a first attempt to check the performance of this solution, despite and regardless the poor object pool implementation, using a benchmark that I'm pasting at the end of this comment. Running this benchmark against the current master branch I obtained the following results:

```
Benchmark                                                                           (parallelTasks)  (useVirtualThreads)   Mode  Cnt         Score        Error   Units
JacksonMultithreadWriteVanilla.writePojoMediaItem                                              1000                 true  thrpt    5       509.060 ±      6.795   ops/s
JacksonMultithreadWriteVanilla.writePojoMediaItem                                              1000                false  thrpt    5      1402.935 ±     56.546   ops/s

```

while rerunning again using this pull request I got:

```
Benchmark                                                                           (parallelTasks)  (useVirtualThreads)   Mode  Cnt        Score       Error   Units
JacksonMultithreadWriteVanilla.writePojoMediaItem                                              1000                 true  thrpt    5      680.704 ±    14.860   ops/s
JacksonMultithreadWriteVanilla.writePojoMediaItem                                              1000                false  thrpt    5     1279.282 ±     4.998   ops/s
```

As expected the native threads execution is suffering of the worse performing object pool implementation, compared with the current `ThreadLocal` based one, while the one running on virtual threads is already faster in my version, so I'm confident of being on the right path for a good solution once the object pooling problem will be properly addressed.

Any remark or suggestion to improve this (including any concern on why this solution couldn't work) is warmly welcome. The code of the benchmark I implemented and used follows:

```
package com.fasterxml.jackson.perf.json;

import tools.jackson.jr.ob.JSON;
import com.fasterxml.jackson.perf.model.MediaItem;
import com.fasterxml.jackson.perf.model.MediaItems;
import com.fasterxml.jackson.perf.util.NopOutputStream;
import org.openjdk.jmh.annotations.*;
import org.openjdk.jmh.infra.Blackhole;

import java.util.concurrent.CountDownLatch;
import java.util.concurrent.Executors;
import java.util.concurrent.TimeUnit;
import java.util.function.Consumer;

@State(value = Scope.Benchmark)
@Fork(1)
public class JacksonMultithreadWriteVanilla {
    private static final JSON json = JSON.std;

    private static final MediaItem item = MediaItems.stdMediaItem();

    @Param({"true", "false"})
    private boolean useVirtualThreads;

    @Param({"1000"})
    private int parallelTasks;

    private Consumer<Runnable> runner;

    @Setup
    public void setup() {
        this.runner = createRunner(useVirtualThreads);
    }

    @Benchmark
    public void writePojoMediaItem(Blackhole bh) throws Exception {
        CountDownLatch countDown = new CountDownLatch(parallelTasks);

        for (int i = 0; i < parallelTasks; i++) {
            runner.accept(() -> {
                bh.consume(write(item, json));
                countDown.countDown();
            });
        }

        try {
            countDown.await();
        } catch (InterruptedException e) {
            throw new RuntimeException(e);
        }
    }

    protected final int write(Object value, JSON writer) {
        NopOutputStream out = new NopOutputStream();
        writer.write(value, out);
        return out.size();
    }

    private Consumer<Runnable> createRunner(boolean useVirtualThreads) {
        if (useVirtualThreads) {
            return Thread::startVirtualThread;
        } else {
            return Executors.newWorkStealingPool()::execute;
        }
    }
}
```

### DIFF
--- a/src/main/java/tools/jackson/core/TokenStreamFactory.java
+++ b/src/main/java/tools/jackson/core/TokenStreamFactory.java
@@ -20,7 +20,6 @@ import tools.jackson.core.json.JsonFactory;
 import tools.jackson.core.sym.PropertyNameMatcher;
 import tools.jackson.core.sym.SimpleNameMatcher;
 import tools.jackson.core.util.BufferRecycler;
-import tools.jackson.core.util.BufferRecyclers;
 import tools.jackson.core.util.JacksonFeature;
 import tools.jackson.core.util.Named;
 import tools.jackson.core.util.Snapshottable;
@@ -1183,25 +1182,6 @@ public abstract class TokenStreamFactory
      */
 
     /**
-     * Method used by factory to create buffer recycler instances
-     * for parsers and generators.
-     *<p>
-     * Note: only public to give access for {@code ObjectMapper}
-     *
-     * @return BufferRecycler instance to use
-     */
-    public BufferRecycler _getBufferRecycler()
-    {
-        // 23-Apr-2015, tatu: Let's allow disabling of buffer recycling
-        //   scheme, for cases where it is considered harmful (possibly
-        //   on Android, for example)
-        if (Feature.USE_THREAD_LOCAL_FOR_BUFFER_RECYCLING.enabledIn(_factoryFeatures)) {
-            return BufferRecyclers.getBufferRecycler();
-        }
-        return new BufferRecycler();
-    }
-
-    /**
      * Overridable factory method that actually instantiates desired
      * context object.
      *
@@ -1212,7 +1192,7 @@ public abstract class TokenStreamFactory
      */
     protected IOContext _createContext(ContentReference contentRef, boolean resourceManaged) {
         return new IOContext(_streamReadConstraints, _streamWriteConstraints,
-                _getBufferRecycler(), contentRef,
+                contentRef,
                 resourceManaged, null);
     }
 
@@ -1229,7 +1209,7 @@ public abstract class TokenStreamFactory
     protected IOContext _createContext(ContentReference contentRef, boolean resourceManaged,
             JsonEncoding enc) {
         return new IOContext(_streamReadConstraints, _streamWriteConstraints,
-                _getBufferRecycler(), contentRef,
+                contentRef,
                 resourceManaged, enc);
     }
 
--- a/src/main/java/tools/jackson/core/io/IOContext.java
+++ b/src/main/java/tools/jackson/core/io/IOContext.java
@@ -1,11 +1,13 @@
 package tools.jackson.core.io;
 
 import java.util.Objects;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import tools.jackson.core.JsonEncoding;
 import tools.jackson.core.StreamReadConstraints;
 import tools.jackson.core.StreamWriteConstraints;
 import tools.jackson.core.util.BufferRecycler;
+import tools.jackson.core.util.BufferRecyclerPool;
 import tools.jackson.core.util.TextBuffer;
 import tools.jackson.core.util.ReadConstrainedTextBuffer;
 
@@ -15,7 +17,7 @@ import tools.jackson.core.util.ReadConstrainedTextBuffer;
  * readers and writers are combined under this object. One instance
  * is created for each reader and writer.
  */
-public class IOContext
+public class IOContext implements AutoCloseable
 {
     /*
     /**********************************************************************
@@ -99,6 +101,8 @@ public class IOContext
      */
     protected char[] _nameCopyBuffer;
 
+    private volatile AtomicBoolean closed = new AtomicBoolean(false);
+
     /*
     /**********************************************************************
     /* Life-cycle
@@ -110,19 +114,18 @@ public class IOContext
      *
      * @param streamReadConstraints constraints for streaming reads
      * @param streamWriteConstraints constraints for streaming writes
-     * @param br BufferRecycler to use, if any ({@code null} if none)
      * @param contentRef Input source reference for location reporting
      * @param managedResource Whether input source is managed (owned) by Jackson library
      * @param enc Encoding in use
      */
     public IOContext(StreamReadConstraints streamReadConstraints,
             StreamWriteConstraints streamWriteConstraints,
-            BufferRecycler br, ContentReference contentRef, boolean managedResource,
+            ContentReference contentRef, boolean managedResource,
             JsonEncoding enc)
     {
         _streamReadConstraints = Objects.requireNonNull(streamReadConstraints);
         _streamWriteConstraints = Objects.requireNonNull(streamWriteConstraints);
-        _bufferRecycler = br;
+        _bufferRecycler = BufferRecyclerPool.borrowBufferRecycler();
         _contentReference = contentRef;
         _managedResource = managedResource;
         _encoding = enc;
@@ -361,5 +364,12 @@ public class IOContext
     private IllegalArgumentException wrongBuf() {
         // sanity check failed; trying to return different, smaller buffer.
         return new IllegalArgumentException("Trying to release buffer smaller than original");
+    }
+
+    @Override
+    public void close() {
+        if (closed.compareAndSet(false, true)) {
+            BufferRecyclerPool.offerBufferRecycler(_bufferRecycler);
+        }
     }
 }
--- a/src/main/java/tools/jackson/core/util/BufferRecyclerPool.java
+++ b/src/main/java/tools/jackson/core/util/BufferRecyclerPool.java
@@ -1,0 +1,13 @@
+package tools.jackson.core.util;
+
+public class BufferRecyclerPool {
+
+    private static final ObjectPool<BufferRecycler> pool = ObjectPool.newLockFreePool(BufferRecycler::new);
+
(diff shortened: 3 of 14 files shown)


src/main/java/tools/jackson/core/io/IOContext.java
```diff
@@ -362,4 +365,11 @@ private IllegalArgumentException wrongBuf() {
         // sanity check failed; trying to return different, smaller buffer.
         return new IllegalArgumentException("Trying to release buffer smaller than original");
     }
+
+    @Override
+    public void close() {
+        if (closed.compareAndSet(false, true)) {
```
franz1981, line 371: This is supposed to be concurrent? If not, and just "ordered" you can use a plain bool

mariofusco: For now I implemented it in a concurrent way, but I agree that this constraint can be relaxed if not necessary.

src/main/java/tools/jackson/core/TokenStreamFactory.java
```diff
@@ -1182,25 +1181,6 @@ public JsonGenerator createGenerator(Writer w) throws JacksonException {
...
 
-    /**
-     * Method used by factory to create buffer recycler instances
-     * for parsers and generators.
-     *<p>
-     * Note: only public to give access for {@code ObjectMapper}
-     *
-     * @return BufferRecycler instance to use
-     */
-    public BufferRecycler _getBufferRecycler()
```
pjfanning, line 1193: It's unlikely that we will accept having any public APIs removed - you will ideally need to keep them and possibly deprecate them.

Otherwise, this PR will need to be retargeted and master and aimed at the 3.0 release which has no release date set for it.

pjfanning: I see this is targeted at master - but be aware, it could be years before 3.0 is released.

mariofusco: I indeed sent this PR against master, as I expect this to be ready and merged for the 3.0 release. I don't think that we could keep alive both this method and the solution based on the new object pool at the same time, because using one will break the assumptions made by the other. That said my understanding was that the methods and fields starting with an underscore were intended only for internal use and not considered part of the public API, am I missing something?

mariofusco: > I see this is targeted at master - but be aware, it could be years before 3.0 is released.

Ok, if so I guess this won't be acceptable for our usage purposed in Quarkus. Should I retarget this against 2.16 branch? Also can you clarify the naming convention around method names starting with an underscore? Is it safe to assume that they are not part of the public API?

pjfanning: public means public. @cowtowncoder may consider allowing an exception - he's the BDFL.

Since it's following the underscore naming convention, you might say it is public but not guaranteed not to be removed at short notice (not up to me to make this call though).

src/main/java/tools/jackson/core/io/IOContext.java
```diff
@@ -99,6 +101,8 @@ public class IOContext
      */
     protected char[] _nameCopyBuffer;
 
+    private volatile AtomicBoolean closed = new AtomicBoolean(false);
```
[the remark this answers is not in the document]
cowtowncoder: Not quite sure how one could reduce footprint as `IOContext`s are independent and independently handled?

src/main/java/tools/jackson/core/io/IOContext.java
```diff
@@ -362,4 +365,11 @@ private IllegalArgumentException wrongBuf() {
         // sanity check failed; trying to return different, smaller buffer.
         return new IllegalArgumentException("Trying to release buffer smaller than original");
     }
+
+    @Override
+    public void close() {
+        if (closed.compareAndSet(false, true)) {
+            BufferRecyclerPool.offerBufferRecycler(_bufferRecycler);
```
cowtowncoder, line 372: If `final` definition was removed, we could perhaps use null-ness of `_bufferRecycler` itself (i.e. only offer if not null)

src/main/java/tools/jackson/core/util/BufferRecyclerPool.java
```diff
@@ -0,0 +1,13 @@
+package tools.jackson.core.util;
+
+public class BufferRecyclerPool {
+
+    private static final ObjectPool<BufferRecycler> pool = ObjectPool.newLockFreePool(BufferRecycler::new);
```
cowtowncoder, line 5: Does this mean that pooling would be global, across all `TokenStreamFactory` instances? If so, that would be against Jackson design in which nothing should be shared between unrelated factory instances.

cowtowncoder, line 5: Ok; since this is now global pool (across all factories), there definitely should be maximum size. But then again, having such setting would lead to need for configuration.
And static singletons are generally a bad idea and not suitable for use by (low-level) libraries.

In fact, now that I think about this, I think pooling (or default implementation at any rate) really needs to be segmented by factory instance -- no buffers should be shared (by default impl) across `TokenStreamFactory` instances.

src/main/java/tools/jackson/core/util/ObjectPool.java
```diff
@@ -0,0 +1,65 @@
...
+
+    static <T> ObjectPool<T> newLockFreePool(Supplier<T> factory) {
+        return new LockFreeObjectPool<>(factory);
+    }
+
+    static <T> ObjectPool<T> newLockFreePool(Supplier<T> factory, Consumer<T> destroyer) {
+        return new LockFreeObjectPool<>(factory, destroyer);
+    }
+
+    class LockFreeObjectPool<T> implements ObjectPool<T> {
```
cowtowncoder, line 30: I think that any pool to use should have maximum size, as a general principle.

cowtowncoder:
Ok so... first, apologies if I misread code. I tried to go over it with limited amount of time so I may well have misunderstood something.

But with that, I think my main concern is that it seems that buffer recycling would now become global across all `TokenStreamFactory` instances.
Ideally (IMO) recycling would -- by default, at least -- be limited to all parsers/generators constructed by a single factory. While there are pros and cons for single global pool vs smaller per-factory ones, I feel that scoping by Jackson should always be by factories (`JsonFactory`, `ObjectMapper`) to keep allocations and contestion bounded.
(I do not mind ability to provide other implementations, with different segmentation).

With per-factory recycling/pooling, configurability could then also be exposed on per-factory basis -- otherwise any configurability would have to be global (static), affecting all use by all frameworks.

To do this, I think factory would need to hold on to factory pools `BufferRecycler`s; create one on construction, pass it on to `IOContext`.

Does this make sense?

cowtowncoder:
I suspect this should be closed for now; we should be able to merge #1061 up to `master` once it gets resolved.

mariofusco:
> I suspect this should be closed for now; we should be able to merge #1061 up to `master` once it gets resolved.

Yes, I will close this pull request. I will keep working on `2.16` branch and will forward port that work to `master` when it will be finalized and merged.

PR #1062, opened 2023-07-20
ReadConstrainedTextBuffer bug reproducer

mariofusco:
This is a reproducer for a bug I found in the `ReadConstrainedTextBuffer`. When reusing a `BufferRecycler` the constraint is calculated on the size of that buffer instead of the limit provided with `maxStringLength`.

### DIFF
--- a/src/test/java/com/fasterxml/jackson/core/util/ReadConstrainedTextBufferTest.java
+++ b/src/test/java/com/fasterxml/jackson/core/util/ReadConstrainedTextBufferTest.java
@@ -1,21 +1,34 @@
 package com.fasterxml.jackson.core.util;
 
+import com.fasterxml.jackson.core.JsonEncoding;
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonToken;
 import com.fasterxml.jackson.core.StreamReadConstraints;
 import com.fasterxml.jackson.core.StreamWriteConstraints;
 import com.fasterxml.jackson.core.io.ContentReference;
 import com.fasterxml.jackson.core.io.IOContext;
+import com.fasterxml.jackson.core.testsupport.AsyncReaderWrapper;
+import com.fasterxml.jackson.core.testsupport.AsyncReaderWrapperForByteArray;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.io.OutputStream;
+import java.io.Writer;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


src/test/java/com/fasterxml/jackson/core/util/ReadConstrainedTextBufferTest.java
```diff
@@ -33,22 +46,64 @@ public void appendString() throws Exception {
...
+
+        // Since it is running on the same thread the ThreadLocal based pooling will make this to reuse the same BufferRecycler
+        TextBuffer constrained = makeConstrainedBuffer(SEGMENT_SIZE, true);
+
+        char[] chars = new char[SEGMENT_SIZE];
+        Arrays.fill(chars, 'A');
+        constrained.append(new String(chars), 0, SEGMENT_SIZE);
+
+        // This should throw an exception but doesn't since the constraint is calculated on the size of the formerly released buffer
+        Assertions.assertThrows(IOException.class, () -> constrained.append(new String(chars), 0, SEGMENT_SIZE));
```
pjfanning, line 86: I'm not worried about this yet. The `append` check is currently not guaranteed to catch when the contents are too long. We know this already. The code was written this way to try to avoid doing an expensive exact size check on all appends.

We have a check in the `contentsAsString` so the StreamConstraintsException will be thrown at that point

Try changing your test to see this in action.

```
        // This should throw an exception but doesn't since the constraint is calculated on the size of the formerly released buffer
        //Assertions.assertThrows(IOException.class, () -> constrained.append(new String(chars), 0, SEGMENT_SIZE));

        constrained.append(new String(chars), 0, SEGMENT_SIZE);
        constrained.contentsAsString(); // fails here
```

To reiterate, this is a deliberate decision to try to maintain performance for the usual scenario where the appends are safe while trying to in some or most cases catch the fact the string is too large at an early stage.

Since buffers are being reused - we are not using much extra memory here.

pjfanning: this appears in our javadocs:

This limit is not exact, the limit is applied when we increase internal buffer sizes and an exception will happen at sizes greater than this limit.

mariofusco: @pjfanning ok, thanks for explanation, I also confirm that calling `contentsAsString` throws an IOException as expected. I will close this pull request and the corresponding bug report and modify the other tests in `ReadConstrainedTextBufferTest` so that they will succeed regardless if its using a brand new `BufferRecycler` or one taken from the pool.

pjfanning: * please don't modify the existing tests - if they can't use buffer-recyler, so be it
* if we have to add separate tests that use buffer-recyclers but that behave a little bit differently, that is fine

Issue #1063, opened 2023-07-20
Wrong limit calculation in the ReadConstrainedTextBuffer when reusing a BufferRecycler.


mariofusco:
When reusing a `BufferRecycler` the constraint of a `ReadConstrainedTextBuffer` is calculated on the size of that buffer instead of the limit provided with `maxStringLength`. A reproducer for this problem is provided [here](https://github.com/FasterXML/jackson-core/pull/1062).

PR #1064, opened 2023-07-20
Introducing BufferRecyclerPool

mariofusco:
This is an adaptation for branch 2.16 of [this pull request for master](https://github.com/FasterXML/jackson-core/pull/1061).

As requested I didn't remove any public method, but added a few new ones, deprecating the old methods that they are intended to replace. I also improved many tests trying to make sure that all Closeable resources are properly closed after their usage. This is necessary to guarantee the correct reuse of the `BufferRecycler` taken from the pool and in general to adhere with the lifecycle of those resources.

The last outstanding part is having an efficient implementation of the object pool itself. If you agree on the general idea of this pull request I will work on it. /cc @cowtowncoder @pjfanning @franz1981 

Note that at the moment this pull request is the indirect cause of a couple of test failures that are actually caused by the bug I reported [here](https://github.com/FasterXML/jackson-core/issues/1063).


src/main/java/com/fasterxml/jackson/core/JsonFactory.java
```diff
@@ -1809,6 +1809,7 @@ protected JsonParser _createParser(InputStream in, IOContext ctxt) throws IOExce
                     e.addSuppressed(e2);
                 }
             }
+            ctxt.close();
```
pjfanning, line 1812: If this is needed, could you move it to its own PR? Seems like a general bug (if it is an issue at all).

mariofusco: No this is not a bug in the current implementation because `IOContext` is not `Autocloseable`. I made it so only with this pull request.

pjfanning: Ok. That makes rolling this out harder. We will need to track down everywhere that an IOContext is created and ensure that it is closed. This could have a big impact.

mariofusco: > Is there any way that we could consider a v2 solution where the IOContexts are registered in a static collection? Meaning we can have a centralised way of closing unclosed IOContexts.
> 
> We can still have the close on the IOContext too and this would include removing the context from the static collection.

This is possible even though will add some (hopefully minimal) performance overhead. However I see 2 more problems with this solution:

1. How can I know if an `IOContext` is a leftover that can be closed or it is still in use? 
2. Who will call the centralized cleanup and when?

pjfanning: sure - if the IOContext only needs to be closed when the new BufferRecyclerPool feature is enabled, then maybe that is enough. If we have a bug where the close is not being called when it should then users can go back to the legacy behaviour until we fix the bug.

mariofusco: The other alternative would be that the `IOContext` doesn't have a reference to a `BufferRecycler` at all. For each and every operation that needs a `BufferRecycler` it will take it from the pool and put it back into the pool immediately after, when it's done. This will increase the contention on the pool (maybe not a big issue with a well implemented pool), but will have the twofold advantage of not requiring any `close` method and also increasing the reuse of the `BufferRecycler`s themselves.

@pjfanning If you think this is more acceptable (and safe) I could try to implement it with a separated pull request. 
@cowtowncoder @franz1981 Feedback is welcome :)

mariofusco: Nope, I'm realizing that what I suggested in my last comment cannot work (or I think so) because doing so the same `BufferRecycler` will be used by different contexts at the same time :( If so please feel free to ignore it.

src/main/java/com/fasterxml/jackson/core/io/IOContext.java
```diff
@@ -122,14 +125,37 @@ public class IOContext
     /**
      * Main constructor to use.
      *
+     * @param src constraints for streaming reads
+     * @param swc constraints for streaming writes
+     * @param contentRef Input source reference for location reporting
+     * @param managedResource Whether input source is managed (owned) by Jackson library
+     *
+     * @since 2.17
```
pjfanning, line 133: 2.16 is not released - assume that this will go into 2.16

src/main/java/tools/jackson/core/util/BufferRecyclerPool.java
```diff
@@ -0,0 +1,15 @@
+package tools.jackson.core.util;
```
pjfanning, line 1: for jackson 2.x, use package com.fasterxml.jackson.core.uti

mariofusco:
With [this commit](https://github.com/FasterXML/jackson-core/pull/1064/commits/fc965499c4aef18f9e3716a54c8749f2f576069a) I tested different object pools implementations in order to benchmark them and pick the one that fit in the best possible way the jackson use case. I also temporarily add a `dependency to JCTools` because I wanted to also test a pool specifically designed to be very good with contention and use it at least as a baseline. Moreover the JCTools' one is the only pool implementation that I found not using `ThreadLocal`s internally.

mariofusco:
I made an extensive performance comparison of all the different pool implementations that I introduced with [this commit](https://github.com/FasterXML/jackson-core/pull/1064/commits/fc965499c4aef18f9e3716a54c8749f2f576069a) in order to choose the best fit for jackson needs.

For this comparison I used the following benchmark that serialize in json a very simple pojo made of only 3 fields (a Person with firstName, lastName and age), performing this operation in parallel on 10, 100 or 1000 (native or virtual) threads. 

```
    @Benchmark
    public void writePojoMediaItem(Blackhole bh) throws Exception {
        CountDownLatch countDown = new CountDownLatch(parallelTasks);

        for (int i = 0; i < parallelTasks; i++) {
            runner.accept(() -> {
                bh.consume(write(item, json));
                countDown.countDown();
            });
        }

        try {
            countDown.await();
        } catch (InterruptedException e) {
            throw new RuntimeException(e);
        }
    }

    protected final int write(Object value, JSON writer) {
        NopOutputStream out = new NopOutputStream();
        try {
            writer.write(value, out);
        } catch (IOException e) {
            throw new RuntimeException(e);
        }
        return out.size();
    }
```

I used this smaller `Person` object instead of the usual `MediaItems.stdMediaItem()` used in other benchmarks because I wanted to maximize the portion of time used by the benchmark to access and use the pool, thus making as evident as possible the performance gaps among the different implementations.

In the remaining part of this analysis I will consider only 5 different possibilities: 

1. using no pool at all.
2. the existing `BufferRecyclers`.
3. the pool using `JCTools` that as written before has been introduced only to have a comparison with a pool specifically designed to be very good with contention and use it as a baseline.

plus the 2 best performing implementations not requiring the introduction of any external dependency

4. the [one based on a `LinkedTransferQueue`](https://github.com/FasterXML/jackson-core/blob/fc965499c4aef18f9e3716a54c8749f2f576069a/src/main/java/com/fasterxml/jackson/core/util/ObjectPool.java#L66) 
5. the [one using a very simple lock free algorithm](https://github.com/FasterXML/jackson-core/blob/fc965499c4aef18f9e3716a54c8749f2f576069a/src/main/java/com/fasterxml/jackson/core/util/ObjectPool.java#L272) based on the CAS on an `AtomicReference`. 

As expected when running with virtual threads the `ThreadLocal` based `BufferRecyclers` performs particularly bad, even worse than not having any pool at all. Conversely the lock free pool is the one with best performances regardless of the number of parallel virtual threads used to run the benchmark.

image

On the other side, when running with the traditional native threads, the `BufferRecyclers` still outperforms all other solutions, with the lock free algorithm being a good second and not far at all from the `BufferRecyclers` especially with a relatively low number of threads. When the number of parallel threads reaches 100 or above the performances of this lock free implementation (while remaining acceptable) tend to worsen a bit and become even lower than the `LinkedTransferQueue` solution, very likely because the highest contention increases the probability of a failure in the CAS.

image

Given all these considerations, and keeping in mind that we are introducing this new pool mostly to deal with virtual threads, I decided to keep the lock free implementation. As suggested by @pjfanning, at least for jackson 2.x, the actual use of this pool will be a feature opt-in that in this way could be used by projects and frameworks already leveraging virtual threads, while the existing `BufferRecyclers` will remain the default implementation in use.

src/main/java/com/fasterxml/jackson/core/util/ObjectPool.java
```diff
@@ -0,0 +1,330 @@
...
+
+        @Override
+        public T borrow() {
+            for (int i = 0; i < 3; i++) {
+                Node<T> currentHead = head.get();
+                if (currentHead == null) {
+                    return factory.get();
+                }
+                if (head.compareAndSet(currentHead, currentHead.next)) {
+                    return currentHead.value;
```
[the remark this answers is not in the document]
franz1981: More info on the GC issue here -> http://psy-lob-saw.blogspot.com/2016/03/gc-nepotism-and-linked-queues.html

NOTE: the most of Java concurrent linked queues/stacks within Java collection lib are already fixed for this

src/main/java/com/fasterxml/jackson/core/util/ObjectPool.java
```diff
@@ -0,0 +1,330 @@
...
+
+        public AtomicReferenceArrayPool(Supplier<T> factory, int size) {
+            this.factory = factory;
+            this.size = size;
+            this.pool = new AtomicReferenceArray<>(size);
+        }
+
+        @Override
+        public T borrow() {
+            int pos = counter.getAndUpdate(i -> i > 0 ? i-1 : i);
```
franz1981, line 117: Same performance problem of the Lock free variant pool: `counter` is highly contended (they are both stacks, actually - linked or array-based) hence would limit scalability. 
I would instead use a striped variant of this

eg

add a parameter "concurrency" which represent the number of "slots" of your pool.
each slot contains a fixed size partition of the total size (ie total size/concurrency).
If you make both total size and concurrency power of 2 would be better (more info later why).

You can then use the thread id to decide which slot to hit first (using power of 2 size allow to perform modulus with `threadId & (size -1)`) and you have 2 strategies here:
- performing cas on the counter owning that slot's index, moving tot he next counter, if failing till trying with all `concurrency` counters (that means that we need to keep track "where" we have borrowed it, or we risk to unbalance the pool on release - searching for some available slot)
- just keep on trying (via a cas loop) on the same slot's index: thanks to the thread id distributions (with v threads in particular) you'll probably be lucky to spread the contention the same

To better deal with the contention, too, the counters per partition could be all stored into an AtomicLongArray at 8/16 long distance to each others, to keep each counter separated by 1 or 2 cache lines to avoid false-sharing.

Another interesting way to deal with this, could be by make the algorithm a bit unbalanced
ie in order to make progress we need to first to borrow right?
Then we could just perform `getAndIncrement` there....
- if we overflow (more then slot capacity) we just switch to a mix of get and compareAndSet (because no available stuff is in) or just allocate
- the release side instead can check what's the status of the counter and use get + compareAndSet: if we overflow it means we have exhausted the capacity, and we just need to decrease it by 1 from the max capacity

src/main/java/com/fasterxml/jackson/core/util/ObjectPool.java
```diff
@@ -0,0 +1,330 @@
...
+        private final Supplier<T> factory;
+
+        public JCToolsPool(Supplier<T> factory, int size) {
+            this.factory = factory;
+            this.queue = new MpmcArrayQueue<>(size);
+        }
+
+        @Override
+        public T borrow() {
+            T t = queue.poll();
```
franz1981, line 253: you can use https://github.com/JCTools/JCTools/blob/master/jctools-core/src/main/java/org/jctools/queues/MpscArrayQueue.java#L330 and https://github.com/JCTools/JCTools/blob/master/jctools-core/src/main/java/org/jctools/queues/MpscArrayQueue.java#L460 to deal with contention in a similar way (and still symmetric) then the lock free pool; if you want to be fair among impls that have this option

pjfanning:
For the feature, could I suggest adding something alongside `JsonFactory.Feature.USE_THREAD_LOCAL_FOR_BUFFER_RECYCLING`.

@mariofusco could you load test disabling the USE_THREAD_LOCAL_FOR_BUFFER_RECYCLING feature when you are testing the BufferRecyclers case? It would be useful to know the performance of BufferRecyclers with this feature enabled (the default) and with it disabled.

mariofusco:
@cowtowncoder @pjfanning @franz1981 I now consider this pull request completed and ready to be reviewed and merged. 

I made the new `ObjectPool` optional and disabled by default, so this shouldn't break backward compatibility in any way. Also there is still probably room to improve the performances of the pool implementation, I left in only the 2 more promising ones for now. As I wrote before the main goal of this pull request is clearly define the life cycle of all objects involved and make them compatible with the use of a pool that is different from the old `ThreadLocal` based one. The use of the pool is well encapsulated and it defines an interface with only 2 methods, so if we will find a better performing implementation it will be straightforward to replace it. This could be eventually investigated with a subsequent pull request.

src/main/java/com/fasterxml/jackson/core/util/BufferRecycler.java
```diff
@@ -189,4 +191,16 @@ protected int charBufferLength(int ix) {
 
     protected byte[] balloc(int size) { return new byte[size]; }
     protected char[] calloc(int size) { return new char[size]; }
+
+    BufferRecycler withPool(ObjectPool<BufferRecycler> pool) {
+        this._pool = pool;
```
franz1981, line 196: do we want to throw an exception in case is set twice? or will replace anything within?
or adding an assert to make evident the impl invariant here?

src/main/java/com/fasterxml/jackson/core/util/BufferRecycler.java
```diff
@@ -189,4 +191,16 @@ protected int charBufferLength(int ix) {
...
+        return this;
+    }
+
+    @Override
+    public void close() {
+        if (_pool != null) {
+            _pool.release(this);
```
franz1981, line 203: if we release we want to make the field null again?

src/main/java/com/fasterxml/jackson/core/util/ObjectPool.java
```diff
@@ -0,0 +1,126 @@
+package com.fasterxml.jackson.core.util;
+
+import java.util.Deque;
+import java.util.concurrent.ConcurrentLinkedDeque;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
+public interface ObjectPool<T> extends AutoCloseable {
```
pjfanning, line 9: Can we add javadoc? - `@since 2.16` and specified to be for internal use only

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclerPool.java
```diff
@@ -0,0 +1,14 @@
+package com.fasterxml.jackson.core.util;
+
+public class BufferRecyclerPool {
+
+    private static final ObjectPool<BufferRecycler> pool = ObjectPool.newObjectPool(new BufferRecycler()::withPool);
+
+    public static BufferRecycler acquireBufferRecycler() {
+        return pool.acquire();
+    }
```
pjfanning, line 9: Any chance that you could keep a count of acquired and released recyclers?

Could be useful in debugging issues where pools are not released.

mariofusco: I added a `DebugPoolDecorator` that can be optionally enabled to this purpose.

src/main/java/com/fasterxml/jackson/core/util/ObjectPool.java
mariofusco: Done.

src/main/java/com/fasterxml/jackson/core/util/ObjectPool.java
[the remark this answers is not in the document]
mariofusco: Fixed.

mariofusco:
@cowtowncoder did you give a look at this? any comment? If you think it's ok I'd appreciate if it could be merged in a reasonable time: I'm keeping having conflicts with other commits (I will fix the last ones asap) and it's becoming increasingly hard to resolve them.

mariofusco:
@cowtowncoder sorry if I insist and ask again, but I'm literally resolving conflicts on this pull request on a daily basis and it's almost becoming a full time job. I'd appreciate it if you could at least provide some feedback. /cc @pjfanning @franz1981

cowtowncoder:
@mariofusco Sorry, I haven't had any time to look into this. I am going on a vacation, so on a plus side there shouldn't be anything to merge for that time (2 weeks). Reading through this PR is high on my TODO list but it requires quite a bit of focus.

I'll add one more (separate) note on a general approach I think makes sense, for 2.16 timeline -- apologies for not trying to reconcile it with your work so far.

cowtowncoder:
So, to me what makes sense is (and once again, apologies for not figuring out how close this PR is from these ideas) as follows:

1. Starting point like `BufferRecyclers` is associated with a single `JsonFactory` (or in 3.0 `TokenStreamFactory`). This is overridable
2. Each new `JsonParser`/`JsonGenerator` is allocated a `BufferRecycler` by a call to `BufferRecyclers` -- and then char/byte arrays are taken from recycler by parser/generator
3. Each `BufferRecycler` needs to be returned to `BufferRecyclers` on `close()` of parser/generator; parser/generator should return actual array to `BufferRecycler` right before this
4. It may be that access to arrays via `BufferRecycler` still needs to use atomic operations? I am not 100% sure about this, but this is probably not a performance hot spot since these should always be uncontested


Once this life-cycle works in 2.16; and with one backing implementation (existing `ThreadLocal` based one), I think there's need to get experience with release, and target alternate implementations for later releases.
This could be 2.17, or, if there are blockers, then master/3.0.

Approach above is based on my strong preference for allowing per-factory pooling as one of the options -- but also allowing global: latter case is achieved by just using single `BufferRecyclers` (with appropriate pooling), former by separate recyclers instance.

mariofusco:
The main intent of this pull request is clearly define a lifecycle for `JsonParser`/`JsonGenerator`. This is necessary to make sure that a resource, in this case the `BufferRecycler`, is properly released into the pool from where it has been acquired, without relying on a more forgiving mechanism like the one provided by the `ThreadLocal`-based pool. 

I also enforced the whole test suite to follow this lifecycle. In particular I ran all tests using the `DebugPoolDecorator`, suggested by @pjfanning, printing a line every time a  `BufferRecycler` is acquired from the pool or released into it and making sure that the 2 methods are invoked exactly the same amount of times.

Regarding your points: 
 
>     1. Starting point like `BufferRecyclers` is associated with a single `JsonFactory` (or in 3.0 `TokenStreamFactory`). This is overridable

At the moment the `BufferRecyclers` isn't associated with anything, but it just provides a few `static` methods and in practice behaves like a singleton. What you wrote is (incidentally) true only because the `JsonFactory` is the only class accessing and using it in its `_getBufferRecycler()` method. I followed exactly the same pattern introducing my `BufferRecyclerPool` that has the same intent and behavior of the `BufferRecyclers`, just without using any `ThreadLocal`. Eventually both the old `BufferRecyclers` and the new `BufferRecyclerPool` could easily become proper instances of the `JsonFactory`, but I believe that this change is out of scope for this pull request and could be implemented with a subsequent commit (I could also take care of this if required).

>     2. Each new `JsonParser`/`JsonGenerator` is allocated a `BufferRecycler` by a call to `BufferRecyclers` -- and then char/byte arrays are taken from recycler by parser/generator

This is already what happens, or more precisely this is what the `JsonParser`/`JsonGenerator` do via their `IOContext` which is the only class holding a reference to the `BufferRecycler`.

>     3. Each `BufferRecycler` needs to be returned to `BufferRecyclers` on `close()` of parser/generator; parser/generator should return actual array to `BufferRecycler` right before this

This is indeed the main point of this pull request: now `IOContext` also implements `AutoCloseable`, so when the `JsonParser`/`JsonGenerator` are closed they also close the underlying `IOContext` which in turn releases the `BufferRecycler` it owns to the pool.
 
>     4. It may be that access to arrays via `BufferRecycler` still needs to use atomic operations? I am not 100% sure about this, but this is probably not a performance hot spot since these should always be uncontested

This has not been changed in any way and it's also out of scope for this pull request.

> Once this life-cycle works in 2.16; and with one backing implementation (existing `ThreadLocal` based one), I think there's need to get experience with release, and target alternate implementations for later releases. This could be 2.17, or, if there are blockers, then master/3.0.

The existing implementation based on `ThreadLocal` is still there and in use by default. My new `ThreadLocal`-free pool is an opt-in option that has to be explicitly enabled with a new `JsonFactory` feature. This should ensure the best and safest backward-compatibility while allowing to frameworks and tools leveraging virtual threads to do so correctly by simply enabling that feature.

src/main/java/com/fasterxml/jackson/core/JsonFactory.java
```diff
@@ -2084,7 +2085,9 @@ protected JsonGenerator _decorate(JsonGenerator g) {
      * Note: only public to give access for {@code ObjectMapper}
      *
      * @return Buffer recycler instance to use
+     * @deprecated use <code>BufferRecyclerPool</code> to get a pooled instance of a <code>BufferRecycler</code>
      */
+    @Deprecated
     public BufferRecycler _getBufferRecycler()
```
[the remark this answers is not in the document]
cowtowncoder: I would much prefer configurable approach instead of hard-coding 2 choices and adding Feature(s) to switch between choices.

src/main/java/com/fasterxml/jackson/core/io/IOContext.java
```diff
@@ -130,25 +133,43 @@ public class IOContext
...
      * @since 2.16
      */
+    @Deprecated
     public IOContext(StreamReadConstraints src, StreamWriteConstraints swc, ErrorReportConfiguration erc,
-            BufferRecycler br, ContentReference contentRef, boolean managedResource)
+                     BufferRecycler br, ContentReference contentRef, boolean managedResource)
     {
-        _streamReadConstraints = src;
-        _streamWriteConstraints = swc;
+        _streamReadConstraints = (src == null) ?
```
cowtowncoder, line 162: Minor nit, probably related to changes in 2.16 in the meantime -- no null checks should be made, caller should pass default values, never null (probably just need to merge/rebase from 2.16)

pjfanning:
Would this work?

* BufferRecycler is still updated to be AutoCloseable. The changes in this PR to close the BufferRecycler will be kept.
* A new BufferRecyclerProvider interface is created. We can have 2 built-in implementations. One based on the existing ThreadLocal approach and one based on this PR's approach.
* JsonFactory has a setter that let's you override the BufferRecyclerProvider.
* No need for a new Feature to control the behaviour.

cowtowncoder:
> Would this work?
> 
>     * BufferRecycler is still updated to be AutoCloseable. The changes in this PR to close the BufferRecycler will be kept.
> 
>     * A new BufferRecyclerProvider interface is created. We can have 2 built-in implementations. One based on the existing ThreadLocal approach and one based on this PR's approach.
> 
>     * JsonFactory has a setter that let's you override the BufferRecyclerProvider.
> 
>     * No need for a new Feature to control the behaviour.

I started thinking along similar lines, although working from `BufferRecyclers` -- however, your suggestion of new provider would be cleaner, basically deprecating `BufferRecyclers` (but using it for legacy support).

I think I can on PR for doing integration pieces separate from this PR (to be merged here when ready).

Aside from mechanics of gettting to `BufferRecycler` (which above would tackle) I think this PR looks reasonable.

cowtowncoder:
@mariofusco @pjfanning  Ok, so, created PR #1083 for suggested integration. I can merge that in 2.16 and master, but wanted to get your feedback first.

mariofusco:
@cowtowncoder @pjfanning I rebased and reworked this pull request against the latest commits and now this is ready to be reviewed again. 

Even though the main purpose of this pull request is not having the fastest possible pool implementations (this is also less relevant now that it is fully pluggable) I also rerun a very quick benchmarking session against it, obtaining the following, quite expected, results:

```
Benchmark                                          (parallelTasks)      (poolStrategy)  (useVirtualThreads)   Mode  Cnt      Score      Error  Units
JacksonMultithreadWriteVanilla.writePojoMediaItem              100               NO_OP                 true  thrpt   10   6222.229 ±   37.120  ops/s
JacksonMultithreadWriteVanilla.writePojoMediaItem              100               NO_OP                false  thrpt   10   7082.762 ±   26.198  ops/s
JacksonMultithreadWriteVanilla.writePojoMediaItem              100        THREAD_LOCAL                 true  thrpt   10   6139.130 ±   35.366  ops/s
JacksonMultithreadWriteVanilla.writePojoMediaItem              100        THREAD_LOCAL                false  thrpt   10  28455.487 ±   78.597  ops/s
JacksonMultithreadWriteVanilla.writePojoMediaItem              100           LOCK_FREE                 true  thrpt   10  13003.136 ±   67.382  ops/s
JacksonMultithreadWriteVanilla.writePojoMediaItem              100           LOCK_FREE                false  thrpt   10  21244.851 ±   99.046  ops/s
JacksonMultithreadWriteVanilla.writePojoMediaItem              100  CONCURRENT_DEQUEUE                 true  thrpt   10   9248.817 ±  134.244  ops/s
JacksonMultithreadWriteVanilla.writePojoMediaItem              100  CONCURRENT_DEQUEUE                false  thrpt   10  22029.189 ±   90.243  ops/s
```

I'd appreciate it if you could give to this the highest possible priority. Let me know if you have any doubts, questions or suggestions to further improve it.

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclers.java
```diff
@@ -185,66 +185,4 @@ public static void quoteAsJsonText(CharSequence input, StringBuilder output) {
...
-        return NonRecyclingRecyclerPool.INSTANCE;
-    }
-
-    /**
-     * Default {@link BufferRecyclerPool} implementation that uses
-     * {@link ThreadLocal} for recycling instances.
-     *
-     * @since 2.16
-     */
-    public static class ThreadLocalRecyclerPool
```
pjfanning, line 209: Is this change agreed with @cowtowncoder ? Also, is there agreement to take the TokenStreamFactory forFactory out of the acquireBufferRecycler params?

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclerPool.java
```diff
@@ -1,17 +1,157 @@
...
+     */
+    static BufferRecyclerPool defaultRecyclerPool() {
+        return PoolStrategy.THREAD_LOCAL.getPool();
+    }
+
+    enum PoolStrategy {
+        NO_OP(new NonRecyclingRecyclerPool()),
+        THREAD_LOCAL(new ThreadLocalRecyclerPool()), 
+        CONCURRENT_DEQUEUE(new ConcurrentDequePool()),
+        LOCK_FREE(new LockFreePool());
```
pjfanning, line 31: do we need both CONCURRENT_DEQUEUE and LOCK_FREE? Since users can provide their BufferRecyclerPools, can't we just have a small set of built-in ones?

src/main/java/com/fasterxml/jackson/core/io/IOContext.java
```diff
@@ -458,4 +479,12 @@ private IllegalArgumentException wrongBuf() {
         // sanity check failed; trying to return different, smaller buffer.
         return new IllegalArgumentException("Trying to release buffer smaller than original");
     }
+
+    @Override
+    public void close() {
+        if (!_closed) {
+            _bufferRecycler.release();
```
pjfanning, line 486: Can we agree on a lifecycle? There is an argument to only release the recycler when the JsonFactory is closed. Meaning IOContext would not need a close method.

mariofusco: Sorry, but I'm not understanding your comment on this. The `JsonFactory` is never closed afaik, maybe you mean the `JsonGenerator`?

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclerPool.java
mariofusco: We can keep only one of them if you prefer, but I would keep at least one to have a built-it virtual threads friendly implementation.

pjfanning: So, the close could be on the JsonFactory though.

src/main/java/com/fasterxml/jackson/core/io/IOContext.java
pjfanning: So, the close could be on the JsonFactory though. Or on the JsonGenerator. With IOContext, that makes 3 possible homes for the code that releases the buffer recycler. We should agree on which of the 3 homes makes the most sense.

mariofusco: The only owner of the `BufferRecycler` (I mean the only holding a reference to it) and then the only one being responsible to release it back to the pool is the `IOContext`.

src/main/java/com/fasterxml/jackson/core/JsonFactory.java
```diff
@@ -2138,7 +2160,12 @@ protected JsonGenerator _decorate(JsonGenerator g) {
      */
     public BufferRecycler _getBufferRecycler()
     {
-        return _getBufferRecyclerPool().acquireBufferRecycler(this);
+        return _getBufferRecyclerPool().acquireBufferRecycler();
+    }
+
+    public static BufferRecycler _getDefaultBufferRecycler()
```
pjfanning, line 2166: why is this needed? For me, the method above should suffice.

pjfanning:
@cowtowncoder I don't like the changes here to make IOContext closeable. I really think that it would be simpler if the JsonFactory acquired one buffer recycler instance (or any TokenStreamFactory but let's just talk about JsonFactory for simplicity). This buffer recycler instance would be returned to the pool after the JsonFactory was closed (a new close method on TokenStreamFactory). All generators and parsers created with a JsonFactory instance would use the same buffer recycler instance.

With the ThreadLocal buffer recycler, the lifecycle doesn't matter. With the new BufferRecyclerPool in this PR, do we really need to acquire and release the buffer recyclers after each generate or parse job?

If we do need to acquire and release the buffer recyclers then I think it is better to manage this in the JsonGenerator and JsonParser code. When each JsonGenerator/JsonParser is created and the buffer recycler instance is released back to the pool when the generator/parser is closed (the close methods already exist).

I think either of these approaches will have far less impact on the code. We won't need to rewrite all the tests to close IOContexts.

cowtowncoder:
Um, no, JsonFactory/TokenStreamFactory is never closed; it does not have lifecycle at this point, nor is there proposal to do that.
I guess adding that would allow closing of (some) `BufferRecyclerPool`s but I don't think that was the intent.
And even if we did that, it would not really help with the part we do want to solve, which is returning `BufferRecycler`s into pool; for that we either need to directly release recyclers, or (which seemed preferable to me) make `IOContext` as owner do that upon getting closed/released.

I'll have to read through PR and discussion again, seeing how things are progressing.

cowtowncoder:
Ok, so @pjfanning, aside from question of whether `JsonParser`/`JsonGenerator` or `IOContext` should own and close `BufferRecycler` instances, I thought there may be misunderstanding on what `BufferRecycler` does -- these are NOT shared instances, ideally, but containers of a set of buffers needed for a single read (JsonParser) or write (JsonGenerator) operation (reading/writing a JSON (etc) document).
So `JsonFactory` cannot own just one instance: each parser/generator needs its own instance. Pool is then the thing that manages access to a set of `BufferRecycler` instances needed. Hence factories have a reference to a pool instance (which may be per-Factory, global singleton, or something in between)

I hope this helps in understanding logic here.

mariofusco:
> Ok, so @pjfanning, aside from question of whether `JsonParser`/`JsonGenerator` or `IOContext` should own and close `BufferRecycler` instances, I thought there may be misunderstanding on what `BufferRecycler` does -- these are NOT shared instances, ideally, but containers of a set of buffers needed for a single read (JsonParser) or write (JsonGenerator) operation (reading/writing a JSON (etc) document). So `JsonFactory` cannot own just one instance: each parser/generator needs its own instance. Pool is then the thing that manages access to a set of `BufferRecycler` instances needed. Hence factories have a reference to a pool instance (which may be per-Factory, global singleton, or something in between)
> 
> I hope this helps in understanding logic here.

I had a similar intuition and was trying to explain the same, but thanks for this more in depth clarification. I believe that my pull request is already coherent with what you described, but if you want to change something or have any suggestion to improve please let me know.

pjfanning:
But can we get rid of IOContext close and move the release of the buffer recyclers to the JsonParser and JsonGenerator close? That would reduce the impact of this PR. Most of the test code changes would vanish.

mariofusco:
> But can we get rid of IOContext close and move the release of the buffer recyclers to the JsonParser and JsonGenerator close? That would reduce the impact of this PR. Most of the test code changes would vanish.

Most of the test code changes are there exactly because in many cases the `JsonParser`s and `JsonGenerator`s weren't properly closed. When they're closed now they also close the underlying `IOContext`, which is the only one holding a reference to the `BufferRecycler` and then in my opinion should be responsible of releasing it to the pool.

src/main/java/com/fasterxml/jackson/core/JsonFactory.java
cowtowncoder: Yeah I don't know why we'd want that.

cowtowncoder: Looks like it is used from IOContext legacy constructor. I'll think of an alternative as this seems bit off.

src/main/java/com/fasterxml/jackson/core/util/BufferRecycler.java
```diff
@@ -82,6 +82,8 @@ public class BufferRecycler
     // Note: changed from simple array in 2.10:
     protected final AtomicReferenceArray<char[]> _charBuffers;
 
+    private BufferRecyclerPool _pool;
```
cowtowncoder, line 85: I am thinking that better place for pool reference would be in `IOContext`; could be made final, not referenced from `BufferRecycler`. Although maybe keeping non-final wherever makes sense to avoid double-release?

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclerPool.java
```diff
@@ -1,17 +1,157 @@
...
+    void releaseBufferRecycler(BufferRecycler recycler);
+
+    /**
+     * Returns the default BufferRecyclerPool implementation which is the thread local based one.
+     */
+    static BufferRecyclerPool defaultRecyclerPool() {
+        return PoolStrategy.THREAD_LOCAL.getPool();
+    }
+
+    enum PoolStrategy {
```
cowtowncoder, line 27: I don't understand the point of `enum` here, to be honest. Maybe it becomes more clear after reading these a few times over.

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclerPool.java
cowtowncoder: I'll have to think about it: I agree about at least one alternative to ThreadLocal (not counting "no-op" one).
I guess due to compact size of implementations I am not super worried about having 2 alternate impls, I guess.

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclerPool.java
```diff
@@ -1,17 +1,157 @@
...
+        }
+    }
+
+    /**
+     * {@link BufferRecyclerPool} implementation that uses
+     * {@link ConcurrentLinkedDeque} for recycling instances.
+     *
+     * @since 2.16
+     */
+    class ConcurrentDequePool implements BufferRecyclerPool {
```
cowtowncoder, line 90: I think it would make sense to require an upper bound for size? Require caller to specify it, even if "unbounded" (Integer.MAX_VALUE).
Would prevent use of `enum` Strategy which I am not big fan of anyway.

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclerPool.java
```diff
@@ -1,17 +1,157 @@
...
+        private final Deque<BufferRecycler> pool = new ConcurrentLinkedDeque<>();
+
+        @Override
+        public BufferRecycler acquireBufferRecycler() {
+            return getBufferRecycler().withPool(this);
```
cowtowncoder, line 95: Instead of chaining like this, could add a new constructor in `BufferRecycler`. Although if ref is to be moved that won't be necessary.

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclerPool.java
```diff
@@ -1,17 +1,157 @@
...
+        }
+    }
+
+    /**
+     * {@link BufferRecyclerPool} implementation that uses
+     * a lock free linked list for recycling instances.
+     *
+     * @since 2.16
+     */
+    class LockFreePool implements BufferRecyclerPool {
```
cowtowncoder, line 115: I'd ask for maximum size, but that might not be possible with lock-free implementation?

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclerPool.java
```diff
@@ -1,17 +1,157 @@
...
+        private final AtomicReference<LockFreePool.Node<BufferRecycler>> head = new AtomicReference<>();
+
+        @Override
+        public BufferRecycler acquireBufferRecycler() {
+            return getBufferRecycler().withPool(this);
+        }
+
+        private BufferRecycler getBufferRecycler() {
+            for (int i = 0; i < 3; i++) {
```
cowtowncoder, line 124: Could use some commentary on why looping for 3 times etc.

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclerPool.java
```diff
@@ -1,17 +1,157 @@
...
+            LockFreePool.Node<BufferRecycler> newHead = new LockFreePool.Node<>(bufferRecycler);
+            for (int i = 0; i < 3; i++) {
+                newHead.next = head.get();
+                if (head.compareAndSet(newHead.next, newHead)) {
+                    return;
+                }
+            }
+        }
+
+        static class Node<T> {
```
cowtowncoder, line 148: Is there point to making this generic, if only used with `Node<BufferRecycler>` parameter?

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclers.java
cowtowncoder: Let me think about this a bit. I don't mind moving implementations out per se, but not sure I like `Strategy` approach.

src/test/java/com/fasterxml/jackson/core/TestVersions.java
```diff
@@ -50,6 +50,6 @@ private IOContext getIOContext() {
         return new IOContext(StreamReadConstraints.defaults(),
```
cowtowncoder, line 50: I think I'd like having `createTestContext()` method in `IOContext` that uses defaults for StreamRead/WriteConstraints, ErrorReportConfiguration(), `ContentReference.unknown()`.

I think I should probably add it concurrently and merge here; not just for this PR but more generally testing.

src/test/java/com/fasterxml/jackson/core/json/async/AsyncConcurrencyTest.java
```diff
@@ -123,24 +130,20 @@ void _testConcurrentAsyncOnce(final int round, final int maxRounds) throws Excep
...
-                        if (w.process()) {
-                            completedCount.incrementAndGet();
-                        }
-                    } catch (Throwable t) {
-                        if (errorCount.getAndIncrement() == 0) {
-                            errorRef.set(t.toString());
-                        }
-                    } finally {
-                        q.add(w);
+            Callable<Void> c = () -> {
```
cowtowncoder, line 133: It'd be great to avoid changes like this to reduce size of diffs.... I assume this may have been introduced by IDE so hard to avoid.

src/test/java/com/fasterxml/jackson/core/json/async/AsyncMissingValuesInObjectTest.java
```diff
@@ -42,94 +42,94 @@ public static Collection<EnumSet<JsonReadFeature>> getTestCases()
...
+            assertEnd(p);
+        }
   }
 
   @Test
   public void testObjectInnerComma() throws Exception {
     String json = "{\"a\": true,, \"b\": false}";
 
-    AsyncReaderWrapper p = createParser(factory, json);
+    try (AsyncReaderWrapper p = createParser(factory, json)) {
```
cowtowncoder, line 66: Similarly would be nice to do this improvement via different PR.

cowtowncoder:
@mariofusco I went ahead and merged latest from 2.16, including some additional refactoring to reduce direct construction of `IOContext` by tests of `jackson-core` (and I hope to reduce calls from other format package tests too).

But I also noticed PR has lots of test refactoring to use auto-close for JsonParser/JsonGenerator -- these are nice and good to have, but increase size of diff a lot. So I was wondering if they'd be easy enough to recreate for separate PR that I could merge into 2.16, and this PR, to make "real changes" here bit easier to see.

I think we are converging on good solution here. Apologies for taking my time and asking for all kinds of things. But I think we are getting nice design and implementation here, all in all.
Big thank you for your patience so far.

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclerPool.java
mariofusco: I believe that the different pool implementations should be singletons (something you already did in your former commit by using that `static final INSTANCE`) and imo the most idiomatic way to have singletons in Java is using `enum`. Anyway I'm open to change this if you want to suggest a different way of doing this.

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclerPool.java
mariofusco: Good point, I just copied a reference implementation. I will remove the generic here.

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclerPool.java
mariofusco: I will need this method anyway because the pool from where this `BufferRecycler` has been taken will have to be set not only when the instance is created for the first time, but also when an existing one is taken again from the pool and acquired by a different  `IOContext`. That because when the `BufferRecycler` is released to the pool I nullify the reference to the pool itself as also suggested by @franz1981.

mariofusco:
> @mariofusco I went ahead and merged latest from 2.16, including some additional refactoring to reduce direct construction of `IOContext` by tests of `jackson-core` (and I hope to reduce calls from other format package tests too).

Ok, I will bring in again your latest commits before continuing this work.

> But I also noticed PR has lots of test refactoring to use auto-close for JsonParser/JsonGenerator -- these are nice and good to have, but increase size of diff a lot. So I was wondering if they'd be easy enough to recreate for separate PR that I could merge into 2.16, and this PR, to make "real changes" here bit easier to see.

Sure, I will split this pull request into 2, one with only tests refactor and the other with the actual pool implementation. I will need to change something also in the main code to compile what I did on the tests side, but I will try to keep these changes minimal. 

> I think we are converging on good solution here. Apologies for taking my time and asking for all kinds of things. But I think we are getting nice design and implementation here, all in all. Big thank you for your patience so far.

No problem, I understand this is quite a huge change and needs to be evaluated and discussed carefully. I also appreciate the work done so far.

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclerPool.java
cowtowncoder: I think ability to create instances can be useful and important for "real" pools, at least ones with bounded size.
This allows use of one pool per factory, or some other splitting. Basically, it seems useful to not mandate use of only global pools.

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclerPool.java
cowtowncoder: Ah. Yes, right you are.

cowtowncoder:
@mariofusco I went ahead and merged test changes related to (auto-)closing of parsers & generators so diff is much smaller now.

mariofusco:
> @mariofusco I went ahead and merged test changes related to (auto-)closing of parsers & generators so diff is much smaller now.

@cowtowncoder That's great thanks a lot! But what's left now? Do you want me to remove that `enum` of pool implementations? Is there anything else that I should do?

src/main/java/com/fasterxml/jackson/core/util/BufferRecycler.java
mariofusco: Releasing the same `BufferRecycler` twice to the same pool will compromise the pool integrity and I implemented it in this way exactly to avoid that risk. I suggest to leave it as it is, but of course I'm open to discuss it. For now I will add a comment in the code to better clarify this.

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclerPool.java
mariofusco: The [`ConcurrentLinkedDeque`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/ConcurrentLinkedDeque.html) on which I based this implementation is "an unbounded concurrent deque based on linked nodes". 

Moreover it worth to note that the javadoc of that class also says: "Beware that, unlike in most collections, the size method is NOT a constant-time operation. Because of the asynchronous nature of these deques, determining the current number of elements requires a traversal of the elements, and so may report inaccurate results if this collection is modified during traversal."

For these reasons I'm afraid that it won't be possible to give an upper bound size to this specific pool implementation. 

More in general I'm skeptical that such a bounded implementation would be even necessary for this kind of pool. To have its size growing in an unmanageable way it should happen that hundreds of threads try to acquire a `BufferRecycler` at the very same time and then also try to release them all together. In my opinion this would be an extreme and pathological situation and in such unlikely scenario not having a pool at all (using the no-op pool) would be the best possible solution. 

Finally what is not clear to me is how a pool should work when its limit is reached and one more thread tries to release one more `BufferRecycler` into the pool. Should we just throw away that `BufferRecycler` instance? Would this be more efficient or convenient than keeping it into the pool anyway?

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclerPool.java
mariofusco: Done.

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclerPool.java
mariofusco: The same things that I wrote in [this other comment](https://github.com/FasterXML/jackson-core/pull/1064#discussion_r1307382381) are also valid here. That said, please let me know if you also want a bounded implementation of the pool (that shouldn't be based on a linked list/deque anyway) and I will add it.

mariofusco:
@cowtowncoder I replied to your questions, added a few comments and removed that `enum` of pool implementations. Please review this again and let me know if you think that it is ok now or want other changes.

cowtowncoder:
@mariofusco Thank you! I think I have grown to like the idea of `BufferRecycler` holding a backref to pool so that's fine.
I'll have to merge one other PR (for property name max length), apologies for once again sluggish follow-up. But I want to get this right which needs focus (... due to distraction day job brings etc).

mariofusco:
No problem @cowtowncoder, please let me know if there is anything else that you want to change here as soon as you will have time to review it.

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclerPool.java
cowtowncoder: Ok so the reason I think a limit makes sense is that it reduces unintentional retention of highest number of buffers ever needed concurrently. With thread-bound pooling there's fixed per-thread overhead; but here we can have somewhat poorly bounded number of instances (each thread, or rather read-or-write operation, can request and release any number of recyclers).

So if I am not mistaken, unbounded pool will retain its maximum peak memory usage, ever, as further operation cannot really reduce number pooled (unless they leak BufferRecyclers :) )

And yes, when pool is full, just throwing away `BufferRecycler` would make sense to me.
The goal is to limit amount of memory retained by underlying byte/char buffers, in the worst case scenario.

Finally, if a pool implementation cannot support max size then we wouldn't try to do impossible but rather just document such limitation.

Does this make more sense?

franz1981: In case the limits are in the number of buffers and not the number of bytes I can provide a mpmc array queue from JCTools, which should scale devently under heavy load.

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclerPool.java
cowtowncoder: @mariofusco I think I would like a bounded variant too, if it's not overly complicated (tons of code). Having unbounded alternatives is fine then, too, with brief note in comments.

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclerPool.java
cowtowncoder: Right, since detecting amount of actual memory (via buffers) is difficult to track, I was thinking of simple number of `BufferRecycler`s to retain.

And when doing that, could have access for approximate size. And maybe even `clear()` method if easy enough to do safely (user would need to keep a reference of pool etc; Jackson would not expose any control methods via `BufferRecyclerPool`  (nor size since many implementations cannot provide that efficiently).

pjfanning: Can we just keep ConcurrentDequePool and delete the LockFreePool? We only need one of these two pools. Since the new code is extensible, users can provide their own pools anyway.

cowtowncoder: I am ok with that: I think minimally we need:

1. No-recycling "pool"
2. ThreadLocal-based (current impl)
3. New non-thread based pool with maximum size limit

but may also choose to add 

4. New non-thread based pool, unbounded

At the same time I am not super worried about having more than minimal number if implementations are small like they are here.

src/main/java/com/fasterxml/jackson/core/io/IOContext.java
cowtowncoder: I think close makes sense here; closing discussion.

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclers.java
cowtowncoder: Ok, so removal of Strategy made sense to me.

And now that I think it through, removal of `TokenStreamFactory` does make sense: it would be useful for constructing or initializing pool, but not for every recycler access.

I am tempted to add some sort of "init(TokenStreamFactory)" method in `BufferRecyclerPool`, but right now don't really have compelling reason for that -- theoretically this would be useful for letting Pool implementations access `TokenStreamFactory` configuration, but none of implementations do that.
And I think if an implementation wants that, they can require use of factory method, and/or just expose configurability explicitly.

So I think this is fine.

cowtowncoder:
@mariofusco Ok, apologies for making tons of tweaking, but I think results are now something I could merge.

Changes mostly concern following things:

1. JDK serializability of `ThreadLocalPool`s: this is only needed in 2.x if `ObjectMapper` (or in theory `JsonFactory` instances) are JDK serialized -- I know some frameworks do that (whether sensibly or not), like Hadoop and Spark (when distributing jobs). In Jackson 3.x this is actually much more efficient thing to do. At any rate, `ThreadLocalPool` is part of serializable configuration of `JsonFactory` (and hence `ObjectMapper`) so it needs to be supported
2. Factory methods for pools: I think distinction between "shared" (global) and "non-shared" (private per-factory instances) needs to be made clear

On (1) there is one thing I didn't yet figure out, and that is keeping identity of global LockFree / DeQue -based pools.
For now JDK ser/deser will not re-create Shared instance. It is doable but for now I didn't have time to make that work.
(for No-op and ThreadLocal one it'd be easy to use `readResolve()` to return shared instance but I didn't bother since they are stateless).

Remaining questions or (minor) open issues from my end are:

1. Can we get one size-bound, stateful instance to complete the set
2. Would it be possible to have unit tests that actually exercise pooling aspect (just verify, without even multi-threading, that `BufferRecycler` instances are recycled as expected. 

Neither of these is really blocker; but if we can resolve them, great. Either way I hope to actually finally get this merged tomorrow if all goes well.
And then have fun merging it into `master` (3.0.0)!

src/test/java/com/fasterxml/jackson/core/write/BufferRecyclerPoolTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+import com.fasterxml.jackson.core.BaseTest;
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.util.BufferRecyclerPool;
+
+import java.io.IOException;
+import java.io.OutputStream;
+
+public class BufferRecyclerPoolTest extends BaseTest
```
cowtowncoder, line 11: It would be great to actually test recycling aspects: no need to use multi-threading, just create sequence of operations, starting with a single `JsonFactory` and single parser/generator, to allocate buffers, check that underlying `BufferRecycler` is properly acquired/released.

If an accessor needs to be added in parser/generator implementation for testing, that's fine.

This is a nice to have tho and just for basic sanity checking. No need to create complicated schemes.

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclerPool.java
```diff
@@ -1,17 +1,288 @@
...
+        }
+
+        // // // JDK serialization support
+
+        /**
+         * To avoid serializing pool contents we made {@code pool} {@code transient};
+         * to compensate, need to re-create proper instance using constructor.
+         */
+        protected Object readResolve() {
+            return new ConcurrentDequePool();
```
cowtowncoder, line 170: This is where we'd need to choose between returning `SHARED` vs creating new one, if we serialized some sort of marker.

cowtowncoder (approved):

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclerPool.java
mariofusco: The `readResolve()` method is typically used to implement the Singleton pattern, where the same object needs to be returned after deserialization. I'd rather return the `SHARED` instance here, otherwise you're silently duplicating the instances available after deserialization, or in other words you will have `JsonFactory` created before and after deserialization that should have the same `SHARED` instance but in reality have different pool instances.

More in general I don't understand the need of a "non-shared" pool instance (at least for these "default" pool implementations provided by us). Can you please clarify in which situation an user may want to have such "non-shared" instance? If there isn't any evident need, I'd rather keep the pools constructors private and only expose the shared instance.

src/test/java/com/fasterxml/jackson/core/write/BufferRecyclerPoolTest.java
mariofusco: Done https://github.com/FasterXML/jackson-core/pull/1064/commits/53941560446ffddec4ac17b245884de7a557e474

mariofusco:
> Remaining questions or (minor) open issues from my end are:
> 
>     1. Can we get one size-bound, stateful instance to complete the set
> 
>     2. Would it be possible to have unit tests that actually exercise pooling aspect (just verify, without even multi-threading, that `BufferRecycler` instances are recycled as expected.

I did both things with my 2 latest commits. 

My last outstanding concern is on the non-shared pools versions. At the moment I see them useful only to test the pools themselves, but nothing else, and they are also problematic with serialization. Also, even if we want to keep them, I also have some doubts about the naming: `shared`/`non-shared` doesn't entirely make sense to me (or at least I don't think it would make sense for somebody not knowing much of this internal implementation detail). I'd keep only the shared version, maybe simply calling it `INSTANCE`, leaving to users the possibility to directly invoke one of the pools' constructors if they really want/need to do so.

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclerPool.java
cowtowncoder: On `readResolve()`: it (or `writeReplace()`) can also be used to prevent serializing part of state, such as actual cached contents. Since JDK serialization does not call any constructors on deserialization (nor, I think, initialization statements? Or does it?), it is necessary to either;

* Clear out cache contents Map before write (writeReplace), to read back empty Map (etc)
* Mark Map field `transient`; use `readResolve()` to create properly initialized empty instance (with empty Map)

cowtowncoder:
> > Remaining questions or (minor) open issues from my end are:
> > ```
> > 1. Can we get one size-bound, stateful instance to complete the set
> > 
> > 2. Would it be possible to have unit tests that actually exercise pooling aspect (just verify, without even multi-threading, that `BufferRecycler` instances are recycled as expected.
> > ```
> 
> I did both things with my 2 latest commits.
> 
> My last outstanding concern is on the non-shared pools versions. At the moment I see them useful only to test the pools themselves, but nothing else, and they are also problematic with serialization. Also, even if we want to keep them, I also have some doubts about the naming: `shared`/`non-shared` doesn't entirely make sense to me (or at least I don't think it would make sense for somebody not knowing much of this internal implementation detail). I'd keep only the shared version, maybe simply calling it `INSTANCE`, leaving to users the possibility to directly invoke one of the pools' constructors if they really want/need to do so.

I feel strongly in this case that plain "instance" is not adequate -- I was thinking of `global` instead of `shared`, as an alternative. 

I agree that convenient access to the globally shared instances, along with allowing construction of non-shared, makes sense.

For the last part, then, the only (?) question is that of JDK serialization properly linking back shared/global to that, but creating new instance for non-shared.
It is doable by f.ex writing an `int` or `boolean` for type (or even `String`). Just some work.

src/main/java/com/fasterxml/jackson/core/json/JsonGeneratorImpl.java
```diff
@@ -245,6 +245,10 @@ public JacksonFeatureSet<StreamWriteCapability> getWriteCapabilities() {
         return JSON_WRITE_CAPABILITIES;
     }
 
+    public IOContext _getIoContext() {
```
cowtowncoder, line 248: Nit:

* Capitalization `getIOContext()`
* If it needs to be public, no leading underscore; only internal methods to use that

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclerPool.java
```diff
@@ -285,4 +287,70 @@ static class Node {
...
+
+    /**
+     * {@link BufferRecyclerPool} implementation that uses
+     * a bounded queue for recycling instances.
+     */
+    class BoundedPool implements BufferRecyclerPool
+    {
+        private static final long serialVersionUID = 1L;
+
+        private static final BufferRecyclerPool SHARED = new BoundedPool(16);
```
cowtowncoder, line 299: 16? If this is number of instances to share globally, should probably be higher, maybe 100?

Either way, should probably be defined as constant, and mentioned in Javadocs.

cowtowncoder: Changed to 100 for now.

src/main/java/com/fasterxml/jackson/core/json/JsonGeneratorImpl.java
cowtowncoder: Changed. Ideally would avoid altogether but... oh well.

cowtowncoder:
Ok: to expedite things I will go ahead and merge -- this does not mean that aspects, naming etc could not be changed based on discussions; I absolutely expect some minor tweaking.
But I figured that since I now have time I can get merging done.

cowtowncoder:
Ok: I am looking for comments to #1117 -- changing the default pool used by Jackson 2.17 and later.

My strawman argument is that we should use:

1. per-factory
2. lock-free
3. That is: one returned by `JsonRecyclerPool.newLockFreePool()`

pool as the default. This is mostly to get discussion going; I don't have strong objection over alternatives.

Issue #1066, opened 2023-07-27
Add configurable error report behavior via `ErrorReportConfiguration`


JooHyukKim:
(note: started from in [google groups discussion](https://groups.google.com/g/jackson-user/c/4rjcZ3s1pDE/m/rNfUomagAQAJ) and follow up of #1043 )

### Motivation

Currently there is not a user-friendly way to configure how error report is made. So when the issue is resolved, user would be able to perform following configurations.

- Configure maximum length of token to include in error messages
- Configure maximum length of raw content to include in error messages

### Modifications

This issue will be used to contain the sub-PRs that are broken down from the very original PR #1043. Currently planned ones are

- First, Add new `ErrorReportConfiguration` type with builder, unit test(s)
    - #1067 
- Retrofit `IOContext` with new `ErrorReportConfiguration`
   - #1068 
- Retrofit `ContentReferece`
    - #1070 
- Change `TSFBuilder` / `JsonFactory` to allow configuring and holding on to that config object
- Minor clean-up, JavaDoc etc...

cowtowncoder:
Completed!

PR #1067, opened 2023-07-27
Add new `ErrorReportConfiguration`

JooHyukKim:
part of #1066

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/ErrorReportConfiguration.java
+++ b/src/main/java/com/fasterxml/jackson/core/ErrorReportConfiguration.java
@@ -1,0 +1,191 @@
+package com.fasterxml.jackson.core;
+
+import java.io.Serializable;
+
+/**
+ * Container for configuration values used when handling errorneous token inputs. 
+ * For example, unquoted text segments.
+ * <p>
+ * Currently default settings are
+ * <ul>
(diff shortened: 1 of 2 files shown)


src/main/java/com/fasterxml/jackson/core/ErrorReportConfiguration.java
```diff
@@ -0,0 +1,191 @@
...
+     * Maximum length of raw content to include in error messages
+     * 
+     * @see Builder#maxRawContentLength(int) 
+     */
+    protected final int _maxRawContentLength;
+
+    private static ErrorReportConfiguration DEFAULT =
+            new ErrorReportConfiguration(DEFAULT_MAX_ERROR_TOKEN_LENGTH, DEFAULT_MAX_RAW_CONTENT_LENGTH);
+
+    public static void overrideDefaultErrorReportConfiguration(final ErrorReportConfiguration errorReportConfiguration) {
```
cowtowncoder, line 49: Could you copy Javadoc from matching method in `StreamWriteConstaints`, warning against use of this method?
In general it should not be used and specifically libraries should not change global defaults, ever.

cowtowncoder: Or actually I can do this.

cowtowncoder (approved):

cowtowncoder:
Thank you @JooHyukKim !

PR #1068, opened 2023-07-27
Retrofit `IOContext` with `ErrorReportConfiguration`

JooHyukKim:
part of #1066 

~~(blocked by #1067)~~

<del>
### Notes

1. As suggested by [original PR 1043](https://github.com/FasterXML/jackson-core/pull/1043#issuecomment-1650737034), trying to minimize changes by

- Introduce "another" new constructor in 2.16(This will keep tests unchanged)
- Use `defaults()` method and not use new constructor in old constructors

2. After this PR merges, we can do clean ups by

- retrofitting old constructors with new constructor 
- Remove the first new constructor in Jackson 2.16 (one without `ErrorReportConfiguration`</del>

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/io/IOContext.java
+++ b/src/main/java/com/fasterxml/jackson/core/io/IOContext.java
@@ -154,6 +154,7 @@ public class IOContext
      * @param br BufferRecycler to use, if any ({@code null} if none)
      * @param contentRef Input source reference for location reporting
      * @param managedResource Whether input source is managed (owned) by Jackson library
+     * @param erc Error report configuration to use
      *
      * @since 2.16
      */
@@ -221,7 +222,7 @@ public class IOContext
     }
 
     /**
-     * Returns : {@link ErrorReportConfiguration}, container for configuration values used when 
+     * @return Configured {@link ErrorReportConfiguration}, container for configuration values used when 
      * handling errorneous token inputs. 
      *
      * @since 2.16
--- a/src/main/java/com/fasterxml/jackson/core/io/IOContext.java
+++ b/src/main/java/com/fasterxml/jackson/core/io/IOContext.java
@@ -239,7 +239,7 @@ public class IOContext
      *
      * @since 2.16
      */
-    public ErrorReportConfiguration errorReportConfiguration() {
+    public ErrorReportConfiguration getErrorReportConfiguration() {
         return _errorReportConfiguration;
     }
 
--- a/src/main/java/com/fasterxml/jackson/core/io/IOContext.java
+++ b/src/main/java/com/fasterxml/jackson/core/io/IOContext.java
@@ -1,5 +1,6 @@
 package com.fasterxml.jackson.core.io;
 
+import com.fasterxml.jackson.core.ErrorReportConfiguration;
 import com.fasterxml.jackson.core.JsonEncoding;
 import com.fasterxml.jackson.core.StreamReadConstraints;
 import com.fasterxml.jackson.core.StreamWriteConstraints;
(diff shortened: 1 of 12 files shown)


src/main/java/com/fasterxml/jackson/core/io/IOContext.java
```diff
@@ -199,6 +233,16 @@ public StreamWriteConstraints streamWriteConstraints() {
...
     }
 
+    /**
+     * Returns : {@link ErrorReportConfiguration}, container for configuration values used when 
+     * handling errorneous token inputs. 
+     *
+     * @since 2.16
+     */
+    public ErrorReportConfiguration errorReportConfiguration() {
+        return _errorReportConfiguration;
```
pjfanning, line 243: can this be called `getErrorReportConfiguration`? Jackson typically uses set/get method naming.

src/main/java/com/fasterxml/jackson/core/io/IOContext.java
```diff
@@ -129,6 +136,8 @@ public class IOContext
      * @param managedResource Whether input source is managed (owned) by Jackson library
      *
      * @since 2.16
+     * @deprecated Since 2.16, use {@link #IOContext(StreamReadConstraints, StreamWriteConstraints, BufferRecycler, 
+     * ContentReference, boolean, ErrorReportConfiguration)} instead.
      */
     public IOContext(StreamReadConstraints src, StreamWriteConstraints swc, BufferRecycler br,
```
cowtowncoder, line 142: Could you also add `@Deprecated` here (for some reason Eclipse warns if that's missing when there's Javadoc `@deprecated`).

src/main/java/com/fasterxml/jackson/core/io/IOContext.java
```diff
@@ -141,6 +150,30 @@ public IOContext(StreamReadConstraints src, StreamWriteConstraints swc, BufferRe
         _contentReference = contentRef;
         _sourceRef = contentRef.getRawContent();
         _managedResource = managedResource;
+        _errorReportConfiguration = ErrorReportConfiguration.defaults();
```
cowtowncoder, line 153: Maybe instead call:

    this(src, swc, br, contentRef, managedResource, ErrorReportConfiguration.defaults();

src/main/java/com/fasterxml/jackson/core/io/IOContext.java
```diff
@@ -199,6 +233,16 @@ public StreamWriteConstraints streamWriteConstraints() {
         return _streamWriteConstraints;
     }
 
+    /**
+     * Returns : {@link ErrorReportConfiguration}, container for configuration values used when 
+     * handling errorneous token inputs. 
+     *
+     * @since 2.16
+     */
+    public ErrorReportConfiguration getErrorReportConfiguration() {
```
cowtowncoder, line 242: Let's follow the New Jackson Naming convention :)

and drop "get", so `errorReportConfiguration()` (like `streamWriteConstraints()`)

cowtowncoder:
Looks good, added couple of suggestions. You need to merge/rebase from 2.16, I made some changes hopefully easy enough to merge.

src/main/java/com/fasterxml/jackson/core/io/IOContext.java
cowtowncoder: Ok, I notice @pjfanning requested `getXxx()` accessor, so it was "right way" at first. We are moving to reduce use of "get" in cases where there is no existing convention (for particular class) -- `get` and `set` method are to be used for general purpose properties/attributes, but not for configuration.
For time being there is obviously some inconsistency: for 3.0 more of get/set prefixes are being/have been removed. But there's no urgency for 2.x.

cowtowncoder (approved):

cowtowncoder:
Ok, and then one important next step is to wire `JsonFactory.Builder` to allow configuration... right now we just get `ErrorReportConfiguration` defaults.

PR #1071, opened 2023-08-01
Change order of arguments to IOContext constructor (group config objects together)

cowtowncoder:


cowtowncoder:
@JooHyukKim I should have noticed it earlier, but I think new config objects should be grouped together, so I moved `ErrorReportConfiguration` to be 3rd arg, not last.

src/main/java/com/fasterxml/jackson/core/io/IOContext.java
```diff
@@ -125,26 +125,6 @@ public class IOContext
...
-     * @param swc constraints for streaming writes
-     * @param br BufferRecycler to use, if any ({@code null} if none)
-     * @param contentRef Input source reference for location reporting
-     * @param managedResource Whether input source is managed (owned) by Jackson library
-     *
-     * @since 2.16
-     * @deprecated Since 2.16, use {@link #IOContext(StreamReadConstraints, StreamWriteConstraints, BufferRecycler, 
-     * ContentReference, boolean, ErrorReportConfiguration)} instead.
-     */
-    @Deprecated
```
cowtowncoder, line 141: Not removed, moved further down (without Javadoc comment).

src/main/java/com/fasterxml/jackson/core/io/IOContext.java
```diff
@@ -157,46 +137,62 @@ public IOContext(StreamReadConstraints src, StreamWriteConstraints swc, BufferRe
...
     @Deprecated
     public IOContext(StreamReadConstraints src, BufferRecycler br,
                      ContentReference contentRef, boolean managedResource)
     {
-        this(src, StreamWriteConstraints.defaults(), br, contentRef, managedResource, ErrorReportConfiguration.defaults());
+        this(src, StreamWriteConstraints.defaults(), ErrorReportConfiguration.defaults(),
+                br, contentRef, managedResource);
+    }
+
+    @Deprecated // In 2.16: remove ASAP
```
cowtowncoder, line 172: Moved from further up; only needed temporarily since it'd have been added in 2.16 so once factories/builders call proper main constructor, this can be safely removed.

Issue #1047, opened 2023-06-12
Add configurable limit for the maximum length of Object property names to parse before failing


cowtowncoder:
(note: part of #637) 

Similar to token-length limits for Numbers (see #815) and Strings (see #863), we need an option to limit maximum Object Property name length to something reasonable, as otherwise name tokenization can become performance issue for very long names.

As with other limits, units would related to underlying input units -- `byte`s or `char`s depending on input source.

We need to decide on reasonable defaults to use: my initial suggestion is to limit to 64k but this probably needs to correlated to:

1. Performance-characteristics (can we find a pivot point where length increase has super-linear negative effect?)
2. Longest legit (non-malicious) property names used in the wild.

That is: figure out highest limit that works for DoS aspect, balanced with lowest limit that would be unlikely to break existing legit usage.

AlessandroPerucchi:
I am also in favor of this ticket, as I wrote in this comment: https://github.com/FasterXML/jackson-core/issues/1014#issuecomment-1623045770

cowtowncoder:
@AlessandroPerucchi note that there is already ability to change existing limits: this would introduce one more limit.

cowtowncoder:
Quick note: based on simple local testing with longer names, it looks like performance of, say, 1_000_000 (ascii) bytes/chars decoding is not superbly bad -- actually, unit test with cold start seems to process doc with 10 meg name in about 150 msecs. And although other factors (escaping, UTF-8 chars) no doubt change things (quick testing with escapes and 2-byte chars actually increases time to almost 500 msecs), I think we can -- if we want to -- allow larger names than 50k, by default, if we want to.

I'll see if I can find limits other processors impose.

cowtowncoder:
Implemented via #1078.

PR #1075, opened 2023-08-02
throw checked exception if growArray has a size that is too big

pjfanning:
* useful as a check in its own right
* also improves behaviour in https://github.com/FasterXML/jackson-core/issues/1047 case
  * I'm not saying that there shouldn't be a max element name length
  * just that this change sorts out the case where you have a crazy long element name - you get a checked exception (StreamConstraintsException) instead of a java.lang.NegativeArraySizeException

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/base/ParserBase.java
+++ b/src/main/java/com/fasterxml/jackson/core/base/ParserBase.java
@@ -6,6 +6,7 @@ import java.math.BigInteger;
 import java.util.Arrays;
 
 import com.fasterxml.jackson.core.*;
+import com.fasterxml.jackson.core.exc.StreamConstraintsException;
 import com.fasterxml.jackson.core.io.IOContext;
 import com.fasterxml.jackson.core.io.ContentReference;
 import com.fasterxml.jackson.core.io.NumberInput;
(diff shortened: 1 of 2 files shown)


src/main/java/com/fasterxml/jackson/core/base/ParserBase.java
```diff
@@ -1553,12 +1554,16 @@ protected ContentReference _contentReferenceRedacted() {
         return ContentReference.redacted();
     }
 
-    protected static int[] growArrayBy(int[] arr, int more)
+    protected static int[] growArrayBy(int[] arr, int more) throws StreamConstraintsException
```
cowtowncoder, line 1557: I don't like use of API level `StreamConstraintsException` at low-level helper methods; just make it `IllegalArgumentException`

cowtowncoder:
Hmmmh. As per my comment, I don't like the idea of API exception from general-purpose helper.
I am bit meh on checking in general, actually, since it seems unlikely this would be hit. But I guess it's fine.
I'll merge & change exception type.

cowtowncoder (approved):

pjfanning:
@cowtowncoder The point is to use a checked exception (ie not an exception that extends RuntimeException). Unfortunately, the people who are obsessed with getting CVEs assigned to everything are obsessed about unchecked exceptions. They go bananas when an exception or error occurs that is not part of the API signature. That spring-boot user was upset that jackson failed with java.lang.NegativeArraySizeException.

If StreamConstraintsException is a problem, could we use IOException?

cowtowncoder:
@pjfanning Let's add proper name size constraints checks and not rely on array expansion exhausting max array size. This is not the way to fix problems -- decoding will become problem way way before int32 exhausts.

src/main/java/com/fasterxml/jackson/core/base/ParserBase.java
cowtowncoder: As per my comment on issue, we can change it back to `StreamConstraintsException` if we absolutely must (run out of time for other checks). Just not yet.

PR #1078, opened 2023-08-08
validate the length of names

pjfanning:
* so far just a POC that only works with the UTF8StreamJsonParser and possibly NonBlockingUtf8JsonParsers
* start with 50,000 char limit on names (as a default, that can be adjusted by users)
* if this approach is ok, the other JsonParser implementations can be given the equivalent checks
  * I have added some checks to the ReaderBasedJsonParser but we need more checks (while the name is being streamed)
* see https://github.com/FasterXML/jackson-core/issues/1047

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/json/ReaderBasedJsonParser.java
+++ b/src/main/java/com/fasterxml/jackson/core/json/ReaderBasedJsonParser.java
@@ -1848,9 +1848,11 @@ public class ReaderBasedJsonParser
             int ch = _inputBuffer[ptr];
             if (ch < codes.length && codes[ch] != 0) {
                 if (ch == '"') {
-                    int start = _inputPtr;
+                    final int start = _inputPtr;
                     _inputPtr = ptr+1; // to skip the quote
-                    return _symbols.findSymbol(_inputBuffer, start, ptr - start, hash);
+                    final int len = ptr - start;
+                    _streamReadConstraints.validateNameLength(len);
+                    return _symbols.findSymbol(_inputBuffer, start, len, hash);
                 }
                 break;
             }
@@ -1864,7 +1866,9 @@ public class ReaderBasedJsonParser
 
     private String _parseName2(int startPtr, int hash, int endChar) throws IOException
     {
-        _textBuffer.resetWithShared(_inputBuffer, startPtr, (_inputPtr - startPtr));
+        final int initLen = _inputPtr - startPtr;
+        _streamReadConstraints.validateNameLength(initLen);
+        _textBuffer.resetWithShared(_inputBuffer, startPtr, initLen);
 
         /* Output pointers; calls will also ensure that the buffer is
          * not shared and has room for at least one more char.
@@ -1908,10 +1912,11 @@ public class ReaderBasedJsonParser
         }
         _textBuffer.setCurrentLength(outPtr);
         {
-            TextBuffer tb = _textBuffer;
-            char[] buf = tb.getTextBuffer();
-            int start = tb.getTextOffset();
-            int len = tb.size();
+            final TextBuffer tb = _textBuffer;
+            final char[] buf = tb.getTextBuffer();
+            final int start = tb.getTextOffset();
+            final int len = tb.size();
+            _streamReadConstraints.validateNameLength(len);
             return _symbols.findSymbol(buf, start, len, hash);
         }
     }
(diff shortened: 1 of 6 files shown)


pjfanning:
@cowtowncoder are these changes on the right track?

src/main/java/com/fasterxml/jackson/core/StreamReadConstraints.java
```diff
@@ -45,6 +45,12 @@ public class StreamReadConstraints
      */
     public static final int DEFAULT_MAX_STRING_LEN = 20_000_000;
 
+    /**
+     * Default setting for maximum name length: see {@link Builder#maxNameLength(int)}
+     * for details.
+     */
+    public static final int DEFAULT_MAX_NAME_LEN = 50_000;
```
cowtowncoder, line 52: This seems like a reasonable value. I am curious as to how timing works on unit tests wrt longer names; quad-based handling is likely rather inefficient as size grows (esp. for escaped names and/or multi-byte UTF-8).

But we definitely need to ask on mailing lists to see if someone out there knows of use cases with unusually long names (I expect someone somewhere does).

src/main/java/com/fasterxml/jackson/core/json/ReaderBasedJsonParser.java
```diff
@@ -1848,9 +1848,11 @@ protected final String _parseName() throws IOException
             int ch = _inputBuffer[ptr];
             if (ch < codes.length && codes[ch] != 0) {
                 if (ch == '"') {
-                    int start = _inputPtr;
+                    final int start = _inputPtr;
                     _inputPtr = ptr+1; // to skip the quote
-                    return _symbols.findSymbol(_inputBuffer, start, ptr - start, hash);
+                    final int len = ptr - start;
+                    _streamReadConstraints.validateNameLength(len);
```
cowtowncoder, line 1854: Ideally we should only (have to) check length when adding new name in `CharsToNameCanonicalizer` (and byte-based counterpart); not for every time we decode a name.

But to do that would need to pass constraints... that might be doable in `makeChild()` method. Maybe adding new variant that takes `JsonFactory`, instead of flags.

src/main/java/com/fasterxml/jackson/core/json/UTF8StreamJsonParser.java
```diff
@@ -2014,6 +2014,12 @@ private final String parseName(int q1, int ch, int lastQuadBytes) throws IOExcep
         return parseEscapedName(_quadBuffer, 0, q1, ch, lastQuadBytes);
     }
 
+    private int[] growArrayWithNameLenCheck(int[] arr, int more) throws StreamConstraintsException {
+        // the following check will fail if the array is already bigger than is allowed for names
+        _streamReadConstraints.validateNameLength(arr.length);
```
cowtowncoder, line 2019: Need to multiply by 4 (`arr.length << 2`) to get to name length in bytes.

cowtowncoder:
@pjfanning yes, this is pretty much along the lines I was thinking. The main concern would be trying to avoid checks for char-based names on every decoding -- seems like it'd make sense to move check into `CharsToNameCanonicalizer` but that requires some plumbing to pass `StreamReadConstraints`.

For byte/quad-based decoding I think checks on array expansion make sense; just need to pass byte-size, not quad size.

src/main/java/com/fasterxml/jackson/core/StreamReadConstraints.java
pjfanning: I created a thread in the Google Group - https://groups.google.com/g/jackson-dev

cowtowncoder:
Should have asked this before but... any chance to get `StreamReadConstraints` additions in a separate PR first, to get merged to `master` -- and then the rest separately?

src/main/java/com/fasterxml/jackson/core/json/async/NonBlockingUtf8JsonParserBase.java
```diff
@@ -2169,12 +2170,13 @@ private final JsonToken _parseEscapedName(int qlen, int currQuad, int currQuadBy
...
         if (currQuadBytes > 0) {
             if (qlen >= quads.length) {
-                _quadBuffer = quads = growArrayBy(quads, quads.length);
+                _quadBuffer = quads = growArrayWithNameLenCheck(quads, quads.length);
             }
             quads[qlen++] = _padLastQuad(currQuad, currQuadBytes);
         } else if (qlen == 0) { // rare, but may happen
             return _fieldComplete("");
         }
+        _streamReadConstraints.validateNameLength(qlen << 2);
```
cowtowncoder, line 2179: Should be moved later, only check if `_symbols.findName()` can't find it.

cowtowncoder:
@pjfanning Actually, I think this is good -- I can make minor changes after merge as I see fit. So if you think this is ready, please change to regular PR from draft?

src/main/java/com/fasterxml/jackson/core/json/async/NonBlockingUtf8JsonParserBase.java
pjfanning: moved to _addName

pjfanning:
@cowtowncoder should I move some of the UTF8StreamJsonParser parser checks to addName? The code there is a bit different because the ByteQuadsCanonicalizer doesn't validate the name length.

src/main/java/com/fasterxml/jackson/core/json/UTF8StreamJsonParser.java
```diff
@@ -2188,10 +2195,11 @@ protected String _handleOddName(int ch) throws IOException
 
         if (currQuadBytes > 0) {
             if (qlen >= quads.length) {
-                _quadBuffer = quads = growArrayBy(quads, quads.length);
+                _quadBuffer = quads = growArrayWithNameLenCheck(quads, quads.length);
             }
             quads[qlen++] = currQuad;
         }
+        _streamReadConstraints.validateNameLength(qlen << 2);
```
cowtowncoder, line 2202: Yes, this could be instead moved to inside `addName()` which is only called if `findName()` does not find already canonicalized name.

cowtowncoder (approved):
LGTM, will merge tomorrow (merging to master/3.0 will take some time)

cowtowncoder:
> @cowtowncoder should I move some of the UTF8StreamJsonParser parser checks to addName? The code there is a bit different because the ByteQuadsCanonicalizer doesn't validate the name length.

Yes, I think moving those couple of cases to only validate in `addName()` and not earlier would make sense.
Theoretically could refactor `ByteQuadsCanonicalizer` to take in validation similar to char-based one, but that's more work and not necessarily better.

cowtowncoder (approved):

cowtowncoder:
Thank you again @pjfanning ! This time merging to `master` went VERY smoothly, somehow. Good job!
I made some minor tweaks to naming but essentially implementation looks straight-forward.
Over time should add similar verification to Smile, CBOR, YAML, XML, Properties. But for now great to have it for JSON.

PR #1079, opened 2023-08-10
ByteSourceJsonBootstrapper uses CharBufferReader for byte[] inputs

schlosna:
Initial draft to address https://github.com/FasterXML/jackson-core/issues/593 (and https://github.com/FasterXML/jackson-core/pull/995#issuecomment-1668292432 ) when deserializing from a `byte[]` as the `InputStreamReader` code path triggers an 8KiB `HeapByteBuffer` allocation for `StreamDecoder` regardless of input byte array length. This allocation significantly penalizes smaller `byte[]` sources.

image

I would appreciate thoughts on the approach here implementing a `Reader` via a `CharBuffer` that is decoded via wrapping the source `byte[]`. This should avoid unnecessary 8KiB heap byte buffer allocation and leverage OpenJDK's continued charset decoding improvements (e.g. https://cl4es.github.io/2021/02/23/Faster-Charset-Decoding.html ).

Initial benchmarks from https://github.com/FasterXML/jackson-benchmarks/pull/9 show `CharBufferReader` providing performance equivalent to `ByteArrayInputStream` source in worst case, and anywhere from ~2x to ~10x speedup in best case.

@carterkozak for thoughts as well.

```
# JMH version: 1.27
# VM version: JDK 20.0.2, OpenJDK 64-Bit Server VM, 20.0.2+9-FR
# 2021 Apple M1 Pro (aarch64)
Benchmark                                       (mode)                   (shape)        (type)  Mode  Cnt   Score   Error  Units
JsonArbitraryFieldNameBenchmark.parse          DEFAULT                   KEY_MAP  INPUT_STREAM  avgt    4   0.107 ± 0.001  us/op
JsonArbitraryFieldNameBenchmark.parse          DEFAULT                   KEY_MAP        READER  avgt    4   0.509 ± 0.030  us/op
JsonArbitraryFieldNameBenchmark.parse          DEFAULT                   KEY_MAP   CHAR_READER  avgt    4   0.127 ± 0.003  us/op
JsonArbitraryFieldNameBenchmark.parse          DEFAULT            RANDOM_KEY_MAP  INPUT_STREAM  avgt    4  10.175 ± 0.354  us/op
JsonArbitraryFieldNameBenchmark.parse          DEFAULT            RANDOM_KEY_MAP        READER  avgt    4   1.835 ± 0.040  us/op
JsonArbitraryFieldNameBenchmark.parse          DEFAULT            RANDOM_KEY_MAP   CHAR_READER  avgt    4   1.539 ± 0.105  us/op
JsonArbitraryFieldNameBenchmark.parse          DEFAULT  BEAN_WITH_RANDOM_KEY_MAP  INPUT_STREAM  avgt    4  10.474 ± 0.178  us/op
JsonArbitraryFieldNameBenchmark.parse          DEFAULT  BEAN_WITH_RANDOM_KEY_MAP        READER  avgt    4   2.346 ± 0.066  us/op
JsonArbitraryFieldNameBenchmark.parse          DEFAULT  BEAN_WITH_RANDOM_KEY_MAP   CHAR_READER  avgt    4   1.894 ± 0.061  us/op
JsonArbitraryFieldNameBenchmark.parse        NO_INTERN                   KEY_MAP  INPUT_STREAM  avgt    4   0.107 ± 0.003  us/op
JsonArbitraryFieldNameBenchmark.parse        NO_INTERN                   KEY_MAP        READER  avgt    4   0.523 ± 0.044  us/op
JsonArbitraryFieldNameBenchmark.parse        NO_INTERN                   KEY_MAP   CHAR_READER  avgt    4   0.127 ± 0.004  us/op
JsonArbitraryFieldNameBenchmark.parse        NO_INTERN            RANDOM_KEY_MAP  INPUT_STREAM  avgt    4   8.891 ± 0.142  us/op
JsonArbitraryFieldNameBenchmark.parse        NO_INTERN            RANDOM_KEY_MAP        READER  avgt    4   1.284 ± 0.027  us/op
JsonArbitraryFieldNameBenchmark.parse        NO_INTERN            RANDOM_KEY_MAP   CHAR_READER  avgt    4   0.947 ± 0.022  us/op
JsonArbitraryFieldNameBenchmark.parse        NO_INTERN  BEAN_WITH_RANDOM_KEY_MAP  INPUT_STREAM  avgt    4   9.178 ± 0.113  us/op
JsonArbitraryFieldNameBenchmark.parse        NO_INTERN  BEAN_WITH_RANDOM_KEY_MAP        READER  avgt    4   1.742 ± 0.110  us/op
JsonArbitraryFieldNameBenchmark.parse        NO_INTERN  BEAN_WITH_RANDOM_KEY_MAP   CHAR_READER  avgt    4   1.365 ± 0.022  us/op
JsonArbitraryFieldNameBenchmark.parse  NO_CANONICALIZE                   KEY_MAP  INPUT_STREAM  avgt    4   0.533 ± 0.027  us/op
JsonArbitraryFieldNameBenchmark.parse  NO_CANONICALIZE                   KEY_MAP        READER  avgt    4   0.484 ± 0.010  us/op
JsonArbitraryFieldNameBenchmark.parse  NO_CANONICALIZE                   KEY_MAP   CHAR_READER  avgt    4   0.130 ± 0.001  us/op
JsonArbitraryFieldNameBenchmark.parse  NO_CANONICALIZE            RANDOM_KEY_MAP  INPUT_STREAM  avgt    4   0.535 ± 0.037  us/op
JsonArbitraryFieldNameBenchmark.parse  NO_CANONICALIZE            RANDOM_KEY_MAP        READER  avgt    4   0.502 ± 0.024  us/op
JsonArbitraryFieldNameBenchmark.parse  NO_CANONICALIZE            RANDOM_KEY_MAP   CHAR_READER  avgt    4   0.163 ± 0.001  us/op
JsonArbitraryFieldNameBenchmark.parse  NO_CANONICALIZE  BEAN_WITH_RANDOM_KEY_MAP  INPUT_STREAM  avgt    4   0.842 ± 0.057  us/op
JsonArbitraryFieldNameBenchmark.parse  NO_CANONICALIZE  BEAN_WITH_RANDOM_KEY_MAP        READER  avgt    4   0.741 ± 0.032  us/op
JsonArbitraryFieldNameBenchmark.parse  NO_CANONICALIZE  BEAN_WITH_RANDOM_KEY_MAP   CHAR_READER  avgt    4   0.416 ± 0.008  us/op
```

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/io/CharBufferReader.java
+++ b/src/main/java/com/fasterxml/jackson/core/io/CharBufferReader.java
@@ -1,0 +1,66 @@
+package com.fasterxml.jackson.core.io;
+
+import java.io.Reader;
+import java.nio.CharBuffer;
+
+public class CharBufferReader extends Reader {
+    private final CharBuffer charBuffer;
+
+    public CharBufferReader(CharBuffer buffer) {
+        this.charBuffer = buffer.duplicate();
+    }
+
+    @Override
+    public int read(char[] chars, int off, int len) {
+        int remaining = this.charBuffer.remaining();
+        if (remaining <= 0) {
+            return -1;
+        }
+        int length = Math.min(len, remaining);
+        this.charBuffer.get(chars, off, length);
+        return length;
+    }
+
+    @Override
+    public int read() {
+        if (this.charBuffer.hasRemaining()) {
+            return this.charBuffer.get();
+        }
+        return -1;
+    }
+
+    @Override
+    public long skip(long n) {
+        if (n < 0L) {
+            throw new IllegalArgumentException("number of characters to skip cannot be negative");
+        }
+        int skipped = Math.min((int) n, this.charBuffer.remaining());
+        this.charBuffer.position(this.charBuffer.position() + skipped);
+        return skipped;
+    }
+
+    @Override
+    public boolean ready() {
+        return true;
+    }
+
+    @Override
+    public boolean markSupported() {
+        return true;
+    }
+
+    @Override
+    public void mark(int readAheadLimit) {
+        this.charBuffer.mark();
+    }
+
+    @Override
+    public void reset() {
+        this.charBuffer.reset();
+    }
+
+    @Override
+    public void close() {
+        this.charBuffer.position(this.charBuffer.limit());
+    }
+}
--- a/src/main/java/com/fasterxml/jackson/core/json/ByteSourceJsonBootstrapper.java
+++ b/src/main/java/com/fasterxml/jackson/core/json/ByteSourceJsonBootstrapper.java
@@ -1,6 +1,9 @@
 package com.fasterxml.jackson.core.json;
 
 import java.io.*;
+import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
+import java.nio.charset.Charset;
 
 import com.fasterxml.jackson.core.*;
 import com.fasterxml.jackson.core.format.InputAccessor;
@@ -230,7 +233,9 @@ public final class ByteSourceJsonBootstrapper
                 InputStream in = _in;
 
                 if (in == null) {
-                    in = new ByteArrayInputStream(_inputBuffer, _inputPtr, _inputEnd);
+                    // [jackson-core#488] Avoid overhead of InputStreamReader allocating ByteBuffer
+                    ByteBuffer byteBuffer = ByteBuffer.wrap(_inputBuffer, _inputPtr, _inputEnd);
+                    return new CharBufferReader(Charset.forName(enc.getJavaName()).decode(byteBuffer));
                 } else {
                     /* Also, if we have any read but unused input (usually true),
                      * need to merge that input in:
--- a/src/main/java/com/fasterxml/jackson/core/json/ByteSourceJsonBootstrapper.java
+++ b/src/main/java/com/fasterxml/jackson/core/json/ByteSourceJsonBootstrapper.java
@@ -2,7 +2,6 @@ package com.fasterxml.jackson.core.json;
 
 import java.io.*;
 import java.nio.ByteBuffer;
-import java.nio.CharBuffer;
 import java.nio.charset.Charset;
 
 import com.fasterxml.jackson.core.*;
@@ -233,9 +232,14 @@ public final class ByteSourceJsonBootstrapper
                 InputStream in = _in;
 
                 if (in == null) {
-                    // [jackson-core#488] Avoid overhead of InputStreamReader allocating ByteBuffer
-                    ByteBuffer byteBuffer = ByteBuffer.wrap(_inputBuffer, _inputPtr, _inputEnd);
-                    return new CharBufferReader(Charset.forName(enc.getJavaName()).decode(byteBuffer));
+                    int size = _inputEnd - _inputPtr;
+                    if (size >= 0 && size <= 8192) {
+                        // [jackson-core#488] Avoid overhead of heap ByteBuffer allocated by InputStreamReader
+                        // when processing small inputs up to 8KiB.
+                        Charset charset = Charset.forName(enc.getJavaName());
+                        return new CharBufferReader(charset.decode(ByteBuffer.wrap(_inputBuffer, _inputPtr, _inputEnd)));
+                    }
+                    in = new ByteArrayInputStream(_inputBuffer, _inputPtr, _inputEnd);
                 } else {
                     /* Also, if we have any read but unused input (usually true),
                      * need to merge that input in:
--- a/src/test/java/com/fasterxml/jackson/core/io/CharBufferReaderTest.java
+++ b/src/test/java/com/fasterxml/jackson/core/io/CharBufferReaderTest.java
@@ -1,0 +1,103 @@
+package com.fasterxml.jackson.core.io;
+
+import java.io.IOException;
+import java.io.Reader;
+import java.nio.CharBuffer;
+import java.util.Arrays;
+
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertThrows;
+
+public class CharBufferReaderTest extends com.fasterxml.jackson.core.BaseTest {
+
+    public void testSingleCharRead() throws IOException {
+        CharBuffer buffer = CharBuffer.wrap("aB\u00A0\u1AE9\uFFFC");
+        Reader charBufferReader = new CharBufferReader(buffer);
+        try (Reader reader = charBufferReader) {
+            assertEquals('a', reader.read());
+            assertEquals('B', reader.read());
+            assertEquals('\u00A0', reader.read());
+            assertEquals('\u1AE9', reader.read());
+            assertEquals('￼', reader.read());
+            assertEquals(-1, reader.read());
+        }
+        assertEquals(-1, charBufferReader.read());
+    }
+
+
+    public void testBulkRead() throws IOException {
+        CharBuffer buffer = CharBuffer.wrap("abcdefghijklmnopqrst\u00A0");
+        char[] chars = new char[12];
+        Reader charBufferReader = new CharBufferReader(buffer);
+        try (Reader reader = charBufferReader) {
+            assertEquals(12, reader.read(chars));
+            assertArrayEquals("abcdefghijkl".toCharArray(), chars);
+            assertEquals(9, reader.read(chars));
+            assertArrayEquals("mnopqrst\u00A0".toCharArray(), Arrays.copyOf(chars, 9));
+            assertEquals(-1, reader.read(chars));
+        }
+        assertEquals(-1, charBufferReader.read(chars));
+    }
+
+
+    public void testSkip() throws IOException {
+        CharBuffer buffer = CharBuffer.wrap("abcdefghijklmnopqrst\u00A0");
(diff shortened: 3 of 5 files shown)


src/main/java/com/fasterxml/jackson/core/io/CharBufferReader.java
```diff
@@ -0,0 +1,66 @@
...
+        }
+        return -1;
+    }
+
+    @Override
+    public long skip(long n) {
+        if (n < 0L) {
+            throw new IllegalArgumentException("number of characters to skip cannot be negative");
+        }
+        int skipped = Math.min((int) n, this.charBuffer.remaining());
```
pjfanning, line 37: please don't cast long to int - there is an edge case where n is too big and the cast value could be nonsense

src/main/java/com/fasterxml/jackson/core/io/CharBufferReader.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @Override
+    public boolean markSupported() {
+        return true;
+    }
+
+    @Override
+    public void mark(int readAheadLimit) {
+        this.charBuffer.mark();
```
pjfanning, line 54: this ignores the input which seems wrong - if we can't work out the right impl, then you should change the `markSupported` to return false

src/main/java/com/fasterxml/jackson/core/io/CharBufferReader.java
```diff
@@ -0,0 +1,66 @@
...
+    @Override
+    public void reset() {
+        this.charBuffer.reset();
+    }
+
+    @Override
+    public void close() {
+        this.charBuffer.position(this.charBuffer.limit());
+    }
+}
```
pjfanning, line 66: can you add a new line at end of file?

src/test/java/com/fasterxml/jackson/core/io/CharBufferReaderTest.java
```diff
@@ -0,0 +1,103 @@
...
+        try (Reader reader = new CharBufferReader(CharBuffer.wrap("test"))) {
+            assertEquals(2, reader.read(chars));
+            assertArrayEquals("te".toCharArray(), chars);
+            reader.close();
+            Arrays.fill(chars, '\0');
+            assertEquals(-1, reader.read(chars));
+            assertArrayEquals("\0\0".toCharArray(), chars);
+        }
+    }
+}
```
pjfanning, line 103: files should end with new lines

src/main/java/com/fasterxml/jackson/core/io/CharBufferReader.java
```diff
@@ -0,0 +1,66 @@
+package com.fasterxml.jackson.core.io;
+
+import java.io.Reader;
+import java.nio.CharBuffer;
+
+public class CharBufferReader extends Reader {
```
pjfanning, line 6: Don't do anything yet but since CharBufferReader is only used in one place - I think it would be a good idea to move it the same package as ByteSourceJsonBootstrapper and to make it package private and final to discourage its use by other users. We should at least add javadoc to the class saying that this class is only for internal use of jackson-core.

src/main/java/com/fasterxml/jackson/core/io/CharBufferReader.java
```diff
@@ -0,0 +1,66 @@
...
+    private final CharBuffer charBuffer;
+
+    public CharBufferReader(CharBuffer buffer) {
+        this.charBuffer = buffer.duplicate();
```
pjfanning, line 10: Is it necessary to duplicate the char buffer? If we make strongly discourage the use of this class and we only have 1 usage of this class - that usage will not modify the input buffer. If we know the input buffer won't be modified, then theoretically, we don't need to duplicate it.

src/main/java/com/fasterxml/jackson/core/io/CharBufferReader.java
schlosna: will change this to:
```suggestion
        int skipped = Math.min(Math.toIntExact(n), this.charBuffer.remaining());
```

pjfanning: My suggestion works for values of n greater than max-int. Your suggestion will raise an exception. If the API defines  n to be long, I think we should respect the API and handle large longs.

pjfanning: It is a funny set of APIs, `CharBuffer.remaining` returns an int so that suggests you can't have more the max-int chars. So maybe, your suggestion works fine. The skip API accepts longs but it appears that it not something we should worry about.

src/main/java/com/fasterxml/jackson/core/io/CharBufferReader.java
schlosna: Will do, wanted to figure out where all this might be needed then encapsulate.

src/main/java/com/fasterxml/jackson/core/io/CharBufferReader.java
schlosna: Will refactor this when encapsulating

src/main/java/com/fasterxml/jackson/core/json/ByteSourceJsonBootstrapper.java
```diff
@@ -230,6 +232,13 @@ public Reader constructReader() throws IOException
...
 
                 if (in == null) {
+                    int size = _inputEnd - _inputPtr;
+                    if (size >= 0 && size <= 8192) {
+                        // [jackson-core#488] Avoid overhead of heap ByteBuffer allocated by InputStreamReader
+                        // when processing small inputs up to 8KiB.
+                        Charset charset = Charset.forName(enc.getJavaName());
+                        return new CharBufferReader(charset.decode(ByteBuffer.wrap(_inputBuffer, _inputPtr, _inputEnd)));
+                    }
                     in = new ByteArrayInputStream(_inputBuffer, _inputPtr, _inputEnd);
```
schlosna, line 242: Open to thoughts here on threshold. 

This new path decoding from `ByteBuffer` to `CharBuffer` will allocate a `char[]` based on the encoding `Charset`, so I chose `8192` as that would be the break even point for `InputStreamReader`'s `StreamDecoder`'s `8192` byte heap `ByteBuffer`.

schlosna (commented):
Thanks for the review @pjfanning ! I have updated to address some of your comments.

src/main/java/com/fasterxml/jackson/core/json/ByteSourceJsonBootstrapper.java
```diff
@@ -230,6 +232,13 @@ public Reader constructReader() throws IOException
                 InputStream in = _in;
 
                 if (in == null) {
+                    int size = _inputEnd - _inputPtr;
+                    if (size >= 0 && size <= 8192) {
+                        // [jackson-core#488] Avoid overhead of heap ByteBuffer allocated by InputStreamReader
+                        // when processing small inputs up to 8KiB.
+                        Charset charset = Charset.forName(enc.getJavaName());
+                        return new CharBufferReader(charset.decode(ByteBuffer.wrap(_inputBuffer, _inputPtr, _inputEnd)));
```
carterkozak, line 240: I wonder how this compares with constructing a string:
```java
return new StringReader(new String(_inputBuffer, _inputPtr, _inputEnd - _inputPtr, charset));
```
The string methods tend to perform very well due to extensive optimization in the jdk, however when we use a reader, the `char[]` used by `CharBufferReader` may perform better since it's not converting between bytes and chars on read.

schlosna: Good call, will benchmark and test this out. For Latin-1/ASCII compressed strings, we'll eat the cost of a byte array copy on string creation and decode should be no-op, while non-Latin-1 case may be a bit more expensive.

pjfanning: The JsonEncoding class only supports Unicode charsets. UTF8, variants of UTF16, variants of UTF32.

schlosna: https://github.com/FasterXML/jackson-benchmarks/pull/9#issuecomment-1673979137 shows `new String` to `StringReader` is better, so I created https://github.com/FasterXML/jackson-core/pull/1081 as an alternative to this PR.

schlosna: > The JsonEncoding class only supports Unicode charsets. UTF8, variants of UTF16, variants of UTF32. 7 bit ASCII is a sunset of UTF8. Latin-1 is not supported.

I should have been more precise. The `Latin-1` was intended to reference [the single byte encoding compact strings JEP 254](https://openjdk.org/jeps/254), but I should have said `non-ASCII UTF-8` will be a bit more expensive in terms of memory overhead.

schlosna:
I'm going to close this out in favor of the simpler, and more efficient https://github.com/FasterXML/jackson-core/pull/1081

PR #1081, opened 2023-08-10
Make `ByteSourceJsonBootstrapper` use `StringReader` for < 8KiB byte[] inputs

schlosna:
From @carterkozak https://github.com/FasterXML/jackson-core/pull/1079#discussion_r1290512114 this is an alternative implementation draft to address https://github.com/FasterXML/jackson-core/issues/593 (and https://github.com/FasterXML/jackson-core/pull/995#issuecomment-1668292432 ) when deserializing from a `byte[]` as the `InputStreamReader` code path triggers an 8KiB `HeapByteBuffer` allocation for `StreamDecoder` regardless of input byte array length. This allocation significantly penalizes smaller `byte[]` sources.

image

The approach here converts `byte[]` inputs smaller thank 8KiB to a `String` and processed via `StringReader`. This should avoid unnecessary 8KiB heap byte buffer allocation and leverage OpenJDK's continued charset decoding improvements (e.g. https://cl4es.github.io/2021/02/23/Faster-Charset-Decoding.html ).

Initial benchmarks from https://github.com/FasterXML/jackson-benchmarks/pull/9 show `StringReader` providing performance equivalent to `ByteArrayInputStream` source in worst case, and anywhere from ~2x to ~10x speedup in best case.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/json/ByteSourceJsonBootstrapper.java
+++ b/src/main/java/com/fasterxml/jackson/core/json/ByteSourceJsonBootstrapper.java
@@ -234,7 +234,7 @@ public final class ByteSourceJsonBootstrapper
 
                 if (in == null) {
                     int length = _inputEnd - _inputPtr;
-                    if (length <= STRING_READER_BYTE_ARRAY_LENGTH_LIMIT && length >= 0) {
+                    if (length <= STRING_READER_BYTE_ARRAY_LENGTH_LIMIT) {
                         // [jackson-core#488] Avoid overhead of heap ByteBuffer allocated by InputStreamReader
                         // when processing small inputs up to 8KiB.
                         return new StringReader(new String(_inputBuffer, _inputPtr, length, enc.getJavaName()));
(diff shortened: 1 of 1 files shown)


src/main/java/com/fasterxml/jackson/core/json/ByteSourceJsonBootstrapper.java
```diff
@@ -230,6 +230,12 @@ public Reader constructReader() throws IOException
                 InputStream in = _in;
 
                 if (in == null) {
+                    int length = _inputEnd - _inputPtr;
+                    if (length >= 0 && length <= 8192) {
```
pjfanning, line 234: minor nit: It may not be worth checking `length >= 0`. It has a small perf impact and the length is very unlikely to negative and will just fail at the ByteArrayInputStream stage anyway. The length == 0 case will probably be more tidily handled with the StringReader than the ByteArrayInputStream route.

schlosna: I flipped the condition to short circuit for large inputs first `if (length <= STRING_READER_BYTE_ARRAY_LENGTH_LIMIT && length >= 0)`.

I'd prefer to keep the `length >= 0` as the array length check here should be negligible, and ensures the `InputStreamReader` continues handling any negative array length path.

schlosna: After thinking it through over night, I removed the `length >= 0` condition.

cowtowncoder:
Ok I assume this has been benchmarked and since it only affects what I assume is minority use case (non-UTF-8 input that comes as byte input), I am ok merging it.
(but obviously is a use case that matters to some otherwise wouldn't be contributed :) ).

Thank you @schlosna !

PR #1085, opened 2023-08-22
Changes to initial BufferRecyclerProvider PR

cowtowncoder:
Addressing issues uncovered in the first PR to add pluggable BufferRecycler pooling.


mariofusco:
@cowtowncoder Thanks a lot. I will rebase https://github.com/FasterXML/jackson-core/pull/1064 against this commit and the former one and rework it accordingly. Please don't change anything in this area in the meanwhile, or at least warn me if you have to do so.

src/main/java/com/fasterxml/jackson/core/util/BufferRecyclerPool.java
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Interface for entity that controls creation and possible reuse of {@link BufferRecycler}
+ * instances used for recycling of underlying input/output buffers.
+ *
+ * @since 2.16
+ */
+public interface BufferRecyclerPool
+    extends java.io.Serializable
+{
+    public BufferRecycler acquireBufferRecycler(TokenStreamFactory forFactory);
```
mariofusco, line 14: @cowtowncoder I don't understand why this method takes a `TokenStreamFactory` as argument, can you please clarify?

cowtowncoder: It allows inspecting configuration of the factory that creates parser/generator that needs recycler: mostly `JsonFactory.Feature` settings.
`TokenStreamFactory` is that abstract super-type of `JsonFactory`; could pass latter if that proves necessary.

cowtowncoder:
> @cowtowncoder Thanks a lot. I will rebase #1064 against this commit and the former one and rework it accordingly. Please don't change anything in this area in the meanwhile, or at least warn me if you have to do so.

Thanks, will try to avoid/limit changes in relevant areas.

PR #1086, opened 2023-08-23
track JsonFactory in CharsToNameCanonicalizer

pjfanning:
* not yet making a change for ByteQuadsCanonicalizer
* based on https://github.com/FasterXML/jackson-core/pull/1078#issuecomment-1688978066
* I don't really understand the subtleties in CharsToNameCanonicalizer. In #1078, I am already validating the length before I call findSymbol on the CharsToNameCanonicalizer so I don't know what benefit there is to also checking in CharsToNameCanonicalizer. The changes in ReaderBasedJsonParser in 1078 are incomplete because waiting until the CharsToNameCanonicalizer findSymbol call is very late. We may already have wasted a lot of processing time by that stage.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/sym/CharsToNameCanonicalizer.java
+++ b/src/main/java/com/fasterxml/jackson/core/sym/CharsToNameCanonicalizer.java
@@ -115,6 +115,11 @@ public final class CharsToNameCanonicalizer
     final protected AtomicReference<TableInfo> _tableInfo;
 
     /**
+     * @since 2.16
+     */
+    final protected JsonFactory _jsonFactory;
+
+    /**
      * Seed value we use as the base to make hash codes non-static between
      * different runs, but still stable for lifetime of a single symbol table
      * instance.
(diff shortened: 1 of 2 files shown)


src/main/java/com/fasterxml/jackson/core/sym/CharsToNameCanonicalizer.java
```diff
@@ -114,6 +114,11 @@ public final class CharsToNameCanonicalizer
      */
     final protected AtomicReference<TableInfo> _tableInfo;
 
+    /**
+     * @since 2.16
+     */
+    final protected JsonFactory _jsonFactory;
```
cowtowncoder, line 120: Ah sorry, I should have been bit clearer here. I think that `JsonFactory` (or ideally just `TokenStreamFactory` if it has all access) should be passed to factory methods, but only used to access configuration that is needed; that configuration (`StreamReadConstraints`, anything else) should be retained but not factory.

cowtowncoder:
Point is not to have check in both places, but only in `CharsToNameCanonicalizer`. For performance reasons check for max length really should not/can not be in tight loop for decoding, unfortunately; things are bit easier for byte-backed case.

But FWTW I am mostly/only worried about byte-based side, decoding of quad-based input into String: not for collecting quads themselves (or `char` buffer). These can be detected right before processing collected raw input.
Some users might prefer more exact failure for exact limit, but that is very difficult to implement for byte-based input (at least wrt limit being defined as `char`s, if that was supported).

cowtowncoder:
@pjfanning I implemented this slightly differently, see #1088. Closing this PR as not needed any more.

PR #1088, opened 2023-08-24
Refactor construction and use of CharsToNameCanonicalizer

cowtowncoder:


cowtowncoder:
@pjfanning This is what I ended up doing instead of #1086 -- it should give access to `StreamReadConstraints` within, to be used for #1078. In the end I think it's slightly better to pass `TokenStreamFactory` anyway; will need to think if I want same for `ByteQuadsCanonicalizer` too for same reason (even if no constraints object needed).

PR #1103, opened 2023-09-07
Hash-pin GitHub Actions

### LINKED ISSUE #1102: Hash-pin workflow GitHub Actions
Hey @cowtowncoder, I'm back with another security suggestion.

When developing with CI workflows, it's common to version-pin dependencies (i.e. `actions/checkout@v3`). Such major-version pins can be useful in that you get immediate access to the latest version of each Action. However, this also makes the project vulnerable should a broken or malicious release be published. However, version tags are mutable, so minor version tags aren't the best solution, since a malicious attacker could overwrite a version tag to point to a malicious or vulnerable commit instead.

Pinning workflow dependencies by hash ensures the dependency is immutable and its behavior is guaranteed.

Since jackson-core already has dependabot set up to monitor GitHub Actions, this won't change much on your end. These hashes will be automatically updated by dependabot.

I'll send a PR pinning the Actions along with this issue.

---

By the way, in case you aren't aware, Dependabot now supports [grouped updates](https://github.blog/2023-08-24-a-faster-way-to-manage-version-updates-with-dependabot/), such that you'll receive all version updates in a single PR, instead of one PR per Action (see [this example](https://github.com/pnacht/libarchive/pull/1)). Let me know if you want me to send a PR to make this change to dependabot as well.

pnacht:
Fixes #1102.

This PR ensures a workflow's behavior isn't disrupted whenever an Action publishes a broken or malicious release by hash-pinning all Actions.

The only exceptions are the cifuzz Actions, which unfortunately have to be kept `@master` due to how the project's infrastructure is set up.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -27,7 +27,7 @@ jobs:
         dry-run: false
         language: jvm
     - name: Upload Crash
-      uses: actions/upload-artifact@v3
+      uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
       if: failure() && steps.build.outcome == 'success'
       with:
         name: artifacts
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,9 +31,9 @@ jobs:
     env:
       JAVA_OPTS: "-XX:+TieredCompilation -XX:TieredStopAtLevel=1"
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4.0.0
     - name: Set up JDK
-      uses: actions/setup-java@v3
+      uses: actions/setup-java@cd89f46ac9d01407894225f350157564c9c7cee2 # v3.12.0
       with:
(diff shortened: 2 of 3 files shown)


cowtowncoder:
Thanks! While I really dislike use of hashes for readability reasons, this is quite ok since:

1. There's comment on which version hash (allegedly :) ) matches, and
2. It's auto-updated by Dependabot (... does it add comment too? Too bad if not)
3. It's for action deps where version doesn't have much significance to me anyway ("just use the latest")

So I think I'll merge this as suggested. Thanks!

cowtowncoder (approved):

pnacht:
> 2. It's auto-updated by Dependabot (... does it add comment too? Too bad if not)

Yes, Dependabot will also keep the version comment up-to-date.

cowtowncoder:
> > 2. It's auto-updated by Dependabot (... does it add comment too? Too bad if not)
> 
> Yes, Dependabot will also keep the version comment up-to-date.

Excellent. That's perfect then.

cowtowncoder:
@pnacht I was wondering if you could provide PR to do the same for `jackson-databind` and `jackson-annotations`? I like how this has worked here.

pnacht:
Done!

PR #1107, opened 2023-09-12
move pool reference from BufferRecycler to IOContext

mariofusco:
This is an alternative solution for the problem reported and tentatively fixed [here](https://github.com/FasterXML/jackson-core/pull/1106). 

This time I moved the pool reference from the `BufferRecycler` to the `IOContext` instance. In this way I got rid of that `withPool` method and now it's `IOContext` responsibility to pass to the `BufferRecycler` the instance of the pool to which it has to be eventually released. 

This solution has the advantage of exposing a much cleaner API, but the drawback of moving the pool reference away from its own `BufferRecycler`. I think that this is fine as long as the `BufferRecycler` is always exclusively used by the `IOContext` and there aren't other classes that reference it either now or in future.

@cowtowncoder @pjfanning Please let me know if this solution is more acceptable than the one that I proposed yesterday (I believe so).

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/JsonFactory.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonFactory.java
@@ -2184,7 +2184,7 @@ public class JsonFactory
             contentRef = ContentReference.unknown();
         }
         return new IOContext(_streamReadConstraints, _streamWriteConstraints, _errorReportConfiguration,
-                _getBufferRecycler(), contentRef, resourceManaged);
+                _getBufferRecycler(), _getBufferRecyclerPool(), contentRef, resourceManaged);
     }
 
     /**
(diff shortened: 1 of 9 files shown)


src/main/java/com/fasterxml/jackson/core/JsonFactory.java
```diff
@@ -2199,8 +2187,9 @@ protected IOContext _createContext(ContentReference contentRef, boolean resource
      */
     @Deprecated // @since 2.13
     protected IOContext _createContext(Object rawContentRef, boolean resourceManaged) {
+        BufferRecyclerPool pool = _getBufferRecyclerPool();
         return new IOContext(_streamReadConstraints, _streamWriteConstraints, _errorReportConfiguration,
-                _getBufferRecycler(),
```
cowtowncoder, line 2203: Alas, I don't like this, due to backwards-(in)compatibility aspect -- this will immediately break usage by those `JsonFactory` sub-classes that override this method (I think CSV one does, probably others).
Or rather, require synchronized changes (technically allowed as per my "internal vs external" compatibility goals.. as long as all 2.16 components work together).

cowtowncoder:
I actually prefer the other approach, even if at some point I was thinking that maybe `IOContext` should handle pool/recycler interaction.

Will add a note on the other PR.

mariofusco:
We preferred and found an agreement on the other solution.

PR #1111, opened 2023-09-21
Call the right filterFinishArray/Object from FilteringParserDelegate

dmikurube:
Hi,

While I was tackling https://groups.google.com/g/jackson-user/c/bLZTjjjXZ28 , I found that `filterFinishObject` is not called properly under some condition. `filterFinishArray` is called against `"}"` instead of `filterFinishObject`.

The first commit is a new test to confirm this behavior (the added test should fail before the second commit), and the second commit fixes the issue.

----

I created this pull request against `2.15` because https://github.com/FasterXML/jackson/blob/master/CONTRIBUTING.md says :

> Most bug-fix Pull Requests should be made against the current stable branch,

(but "the current stable branch" may be now `2.15`, not `2.14`?)

Please tell me if another branch should be the base.

----

I have not signed/sent the CLA yet. I'll do it in a couple of days.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/filter/FilteringParserDelegate.java
+++ b/src/main/java/com/fasterxml/jackson/core/filter/FilteringParserDelegate.java
@@ -403,7 +403,11 @@ public class FilteringParserDelegate extends JsonParserDelegate
                 boolean returnEnd = _headContext.isStartHandled();
                 f = _headContext.getFilter();
                 if ((f != null) && (f != TokenFilter.INCLUDE_ALL)) {
-                    f.filterFinishArray();
+                    if (t.id() == JsonTokenId.ID_END_ARRAY) {
+                        f.filterFinishArray();
+                    } else {
+                        f.filterFinishObject();
+                    }
                 }
                 _headContext = _headContext.getParent();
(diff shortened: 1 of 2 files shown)


cowtowncoder (approved):
LGTM

cowtowncoder:
@dmikurube First of all, thank you for this contribution! Always great to get a bug report AND matching fix.

Second, yes, 2.15 is fine since this looks like very safe fix (low likelihood of regression). I can then merge it forward to 2.16.

Third: one formality before I can merge this in: I need a CLA (from https://github.com/FasterXML/jackson/blob/master/contributor-agreement.pdf) -- it's one-pager and only needs to be sent once before the first contribution, works for all future contributions.
The easiest way is usually to print it, fill & sign, scan/photo, email to `info` at fasterxml dot com.
Once I receive it I'll proceed with merging.

Thank you once again for the fix!

dmikurube:
Thanks, @cowtowncoder!

Yeah, sorry to be late. I sent my CLA just now.

cowtowncoder:
CLA received, will merge.

cowtowncoder (approved):

dmikurube:
Thanks!

PR #1115, opened 2023-09-26
Fix #1114: make BufferRecyclerPool generic

### LINKED ISSUE #1114: Make `BufferRecyclerPool` parameterized with type of instances pooled, to allow reuse
(note: continuation of #1064 and #1106)

As per #1064, #1106 we now have configurability for buffer recycling, usable for JSON parser/generator as well as other data formats.
However: there are some modules (Smile and Avro format modules, JAXB/Jakarta-Bind) that use old-style `ThreadLocal`-based non-configurable pooling.
While I tried to think of ways to perhaps extend `BufferRecycler` to allow for "piggy-backing" other buffers, I couldn't come up with a way that seemed Right. Instead, it seems like we should be able to reuse pool components added so far to 2.16, esp. since we can still change the API before release.

My initial thinking is that parameterizing `BufferRecyclerPool` with type of recycled item (in case of default impl, `BufferRecycler`), coupled with refactoring of the default use, would allow other modules to similarly switch to using the new pluggable mechanism. This is important since same issues wrt use of `ThreadLocal` affect those modules and it seems pointless to copy all (or parts) of code here.

Having said that, addition of type parameters may be cumbersome. But I think it is doable.

And I think it's important to get it in 2.16, if at all possible, so API doesn't need to change in future.

cowtowncoder:
PR to implement #1114 by rewriting `BufferRecyclerPool`, implementations so that there are:

* Abstract base types for recycling any types (not just `BufferRecycler`)
* JsonFactory-specific subtypes, implementation classes that pool `BufferRecycler` values.

Will add new container class `JsonBufferRecyclers` for latter, keep `BufferRecyclerPool` for former.
Will not yet rename `BufferRecyclerPool` but may consider it as follow up.


src/main/java/com/fasterxml/jackson/core/JsonFactory.java
```diff
@@ -262,7 +262,7 @@ public static int collectDefaults() {
     /**
      * @since 2.16
      */
-    protected BufferRecyclerPool _bufferRecyclerPool;
+    protected BufferRecyclerPool<BufferRecycler> _bufferRecyclerPool;
```
cowtowncoder, line 265: Will probably consider renaming of `BufferRecyclerPool` as it'll allow recycling other things, but initially leaving name as-is to simplify merging to `master`.

src/main/java/com/fasterxml/jackson/core/util/JsonBufferRecyclerPools.java
```diff
@@ -0,0 +1,15 @@
...
+import com.fasterxml.jackson.core.JsonFactory;
+
+/**
+ * Set of {@link BufferRecyclerPool} implementations to be used by the default
+ * JSON-backed {@link JsonFactory} for recycling {@link BufferRecycler}
+ * containers.
+ *
+ * @since 2.16
+ */
+public final class JsonBufferRecyclerPools
```
cowtowncoder, line 12: Will move `BufferRecycler`-bound implementations here next.

mariofusco:
@cowtowncoder I saw what you did and I agree with this change.I also originally implemented the pool in a generic way because I "felt" it was the right thing to do, but then I reverted it. 

I also checked that this works fine with the [quarkus integration](https://github.com/quarkusio/quarkus/pull/35816) that I'm attempting. By the way we would need the stable 2.16 asap in order to start using the pool also on our side, do you have any on when it will be released?

cowtowncoder:
@mariofusco I have been about to send an email about 2.16 plans. Things are quite close to getting RC1 out; I just have one feature I really want to get in `jackson-databind` before that. So if all went well, could get RC out in 2-3 weeks. And this time around only doing one -- seems like few devs really use release candidates anyway. RC1 should have recycler pool thing finalized wrt API (unless issues were found from within).

Also: I (or someone else) need to work on using now-generic pools to replace existing `ThreadLocal` based hard-coded reuse on Smile, Avro and... JAX-RS module? Ideally would tackle for 2.16.

cowtowncoder:
Making good progress on "last" issue to resolve ([databind ](https://github.com/FasterXML/jackson-databind/issues/4095), then https://github.com/FasterXML/jackson-databind/issues/4096); email sent.
But maybe I'll get to doing RC1 during next week.

PR #1135, opened 2023-11-07
test that shows that isNaN call is not reliable

pjfanning:
* test that demos that isNan is affected by whether we ask for the value as a double
* the test is useful even without the isNaN check
* we may want to change the impl of isNaN so that the return is not affected by getting the value as double
* see https://github.com/FasterXML/jackson-databind/pull/4191

### DIFF
--- a/src/test/java/com/fasterxml/jackson/core/read/NonStandardNumberParsingTest.java
+++ b/src/test/java/com/fasterxml/jackson/core/read/NonStandardNumberParsingTest.java
@@ -6,6 +6,8 @@ import org.junit.runners.MethodSorters;
 import com.fasterxml.jackson.core.*;
 import com.fasterxml.jackson.core.json.JsonReadFeature;
 
+import java.math.BigDecimal;
+
 @FixMethodOrder(MethodSorters.NAME_ASCENDING) // easier to read on IDE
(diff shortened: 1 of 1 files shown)


cowtowncoder:
I think it would make more sense to create a `failing` test? As is, test verifies what we think is an incorrect behavior, right?

But there is an interesting question here wrt if and how to allow checking of "out of range for Double" case, as opposed to `NaN`.

pjfanning:
The test passes. I think it useful as is. There may be merit changing the behaviour of 'isNan' method or adding a new method that checks if the number if Positive or Negative Infinity or NaN.

cowtowncoder:
I just conceptually dislike tests that verify Wrong behavior... but I can see usefulness.

As to adding more methods, yeah I don't know. Problem is that for Double over-/underflow the answer depends on target: for `BigDecimal` they are finite numbers, for double/Double infinite.

cowtowncoder:
@pjfanning I think I agree with this in general. But since it is very unlikely we'll try to fix this for 2.16 (due to change in behavior someone might rely), could you re-base/re-create this for 2.17?

pjfanning:
@cowtowncoder rebased

cowtowncoder (approved):

PR #1136, opened 2023-11-07
Change parsing error message to mention `-INF`

pjfanning:
-INF is allowed.

The code is unclear about whether 'INF' is allowed. '+INF' seems to be tested and seems to work but 'INF' seems to be documented but I'm not sure if it works.


cowtowncoder:
That sounds reasonable if correct; but let me try to have a look first... definitely needs to include `-INF`

cowtowncoder:
Hmmmh. So, tests in `com.fasterxml.jackson.core.read.NonStandardParserFeaturesTest` check for content:

```
        final String JSON = "[ -INF, +INF, +Infinity, Infinity, -Infinity ]";
```

which does NOT include `INF` but does full versions. `ReaderBasedJsonParser` appears to (maybe?) accept it.
Test mentioned above does fail on `INF` for what that is worth.

So I think I'll go with this.

cowtowncoder (approved):

PR #1150, opened 2023-12-02
Fix `JsonParser.isNaN()` to be stable, only indicate explicit NaN values; not ones due to coercion

cowtowncoder:
First part of fixes related to problems with number Overflow/Underflow for `Double` leading to problems with `BigDecimal` access. Second part of solution will be #1149.


cowtowncoder:
@pjfanning I think that this first part should help a bit: we'll simply keep track of input that is NaN and not consider possible Overflow/Underflow-induced cases actual NaNs (since they would be decodable into `BigDecimal`).
This allows some changes, but more importantly if this goes cleanly, can implement #1149.

/cc @JooHyukKim

Issue #1161, opened 2023-12-10
NPE in "FastDoubleParser", method "JavaBigDecimalParser.parseBigDecimal()"


cowtowncoder:
(note: regression due to #1157)

Looks like there is an NPE in FDP, method `JavaBigDecimalParser#parseBigDecimal` for Strings of length 1000 (and seemingly even less).
I will do a PR with reproduction.

wrandelshofer:
I made a fix and have released it as version 1.0.0 of fastdoubleparser.
https://github.com/wrandelshofer/FastDoubleParser/releases/tag/v1.0.0

Please let me know, if the fix works for you.

pjfanning:
Thanks @wrandelshofer for the prompt fix - I'll try out the 1.0.0 release when it hits https://repo1.maven.org/maven2/ch/randelshofer/fastdoubleparser/ (maven central).

cowtowncoder:
Fixed in 2.16 for 2.16.1 (and `2.17`); may be backported to 2.15 depending.

PR #1162, opened 2023-12-10
Reproduction for #1161: FastDoubleParser NPE

cowtowncoder:
Note: looks like the issue specifically affects parsing from `char[]`, but not (for some reason), `String`.


cowtowncoder:
@pjfanning Looks like FDP has some odd issue with variant of `BigDecimal` parsing method(s) that take `char[]`. I can work around the issue by using `String` taking methods but it'd be better to figure out root cause.

pjfanning:
this will need to reported upstream to the fastdoubleparser team

pjfanning:
created https://github.com/wrandelshofer/FastDoubleParser/pull/77

cowtowncoder:
Thank you @pjfanning.

PR #1163, opened 2023-12-10
fastdoubleparser 1.0.0

pjfanning:
relates to #1161 and https://github.com/wrandelshofer/FastDoubleParser/pull/77

PR #1164, opened 2023-12-11
fastdoubleparser 1.0.0 (#1163)

pjfanning:
backport #1163

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -182,6 +182,7 @@ com.fasterxml.jackson.core.*;version=${project.version}
                   <artifact>ch.randelshofer:fastdoubleparser</artifact>
                   <excludes>
                     <exclude>META-INF/versions/**/module-info.*</exclude>
+                    <exclude>META-INF/versions/22/**/*.*</exclude>
                   </excludes>
                 </filter>
               </filters>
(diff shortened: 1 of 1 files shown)


pjfanning:
@cowtowncoder this would almost certainly backport cleanly to 2.15 if there were any plans for a future 2.15 release

2.15 uses fastdoubleparser 0.9.0 already so same change should work there too

JooHyukKim:
> this would almost certainly backport cleanly to 2.15

If so, we need to create a new PR based on 2.15 branch. That way we can just forward merge.

cowtowncoder:
@JooHyukKim Normally yes, but unfortunately `pom.xml` changes are different due to merge conflict wrt release version. As a result, manual merge (or separate PRs) are typically needed.

@pjfanning Yes, I think 2.15 backport would be welcome too if and when it's easy to do. There will likely be one more patch release from that branch (2.15.4).

cowtowncoder (approved):

cowtowncoder:
@pjfanning whops. Did not notice but this was (accidentally I assume) against `2.17`, not 2.16?

pjfanning:
This PR was targeting the wrong barnch. I have created #1165 for 2.16 branch.

Issue #1173, opened 2023-12-15
`JsonLocation` consistently off by one character for many invalid JSON parsing cases


hal7df:
# Issue
The `JsonLocation` attached to a `JsonProcessingException` thrown when parsing an invalid JSON string is consistently one character to the right of the invalid character, except in cases where the error is due to an unexpected EOF. This affects both `JsonLocation.getCharOffset()` and `JsonLocation.getColumnNr()` (presumably `JsonLocation.getByteOffset()` as well, although I haven't tested that explicitly); because this information is used to construct the exception message, that is affected as well.

I first noticed this on Jackson 2.10.0, but the issue persists on Jackson 2.16.0.

## Minimally reproducible example
```java
package example;

import com.fasterxml.jackson.core.JsonLocation;
import com.fasterxml.jackson.core.JsonProcessingException;
import com.fasterxml.jackson.databind.ObjectMapper;

public class JsonLocationTest {
    public static void main(String[] args) {
        for (String input : INVALID_JSON) {
            try {
                MAPPER.readTree(input);
            } catch (JsonProcessingException jpe) {
                JsonLocation location = jpe.getLocation();

                System.out.printf(
                    "%s (at line %d, column %d):%n  %s%n  ",
                    jpe.getOriginalMessage(),
                    location.getLineNr(),
                    location.getColumnNr(),
                    input
                );

                // annotate the location of the error on the previous line
                for (int i = 0; i < location.getCharOffset(); ++i) {
                  System.out.print(' ');
                }

                System.out.println('^');
            }
        }
    }

    private static final String[] INVALID_JSON = new String[] {
        "{\"invalid\" \"json\"}",
        "{\"also\", \"invalid\"}",
        "{\"still\":[\"invalid\"[]]}",
        "[\"json\":\"doesn't\",\"work\":\"like\",\"this\"]",
        "{\"really\": \"funky💃\" \"unicode\"}",
        "{",
        "}"
    };
    private static final ObjectMapper MAPPER = new ObjectMapper();
}
```

### Expected output
```
Unexpected character ('"' (code 34)): was expecting a colon to separate field name and value (at line 1, column 12):
  {"invalid" "json"}
             ^
Unexpected character (',' (code 44)): was expecting a colon to separate field name and value (at line 1, column 8):
  {"also", "invalid"}
         ^
Unexpected character ('[' (code 91)): was expecting comma to separate Array entries (at line 1, column 20):
  {"still":["invalid"[]]}
                     ^
[...]

PR #1175, opened 2023-12-22
Add failing tests for #1173: off-by-one discrepancy in the location of reported parser errors

hal7df:
This adds (currently failing) test cases for the issue identified in #1173. I've tried my best at adding a plethora of different variations of invalid JSON, but it's possible I might still be missing some edge cases. I also have written this to test every variation of invalid JSON against the four parser backends identified by @cowtowncoder in the issue comments.

### DIFF
--- a/src/test/java/com/fasterxml/jackson/failing/read/LocationOfError1173Test.java
+++ b/src/test/java/com/fasterxml/jackson/failing/read/LocationOfError1173Test.java
@@ -1,0 +1,326 @@
+package com.fasterxml.jackson.failing.read;
+
+import java.io.ByteArrayInputStream;
+import java.io.DataInput;
+import java.io.DataInputStream;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Stream;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import com.fasterxml.jackson.core.*;
+import com.fasterxml.jackson.core.async.ByteArrayFeeder;
+
+/**
+ * Tests that the {@link JsonLocation} attached to a thrown {@link JsonProcessingException} due to invalid json points
+ * to the correct character.
+ */
+public class LocationOfError1173Test extends BaseTest
+{
+    private static final JsonFactory JSON_F = new JsonFactory();
+
+    /** Represents the different parser backends */
+    public enum ParserVariant
+    {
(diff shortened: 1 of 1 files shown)


hal7df:
Curiously, `DataInput`-based parsers pass all of these tests, because they don't provide byte-/character-level position information. It's possible that they might still be susceptible to the off-by-one error if the error occurs across a line boundary, but I haven't been able to get that behavior to occur yet.

src/test/java/com/fasterxml/jackson/failing/read/LocationOfError1173Test.java
```diff
@@ -0,0 +1,326 @@
...
+            21
+        ),
+        new InvalidJson(
+            "Invalid JSON with raw unicode character",
+            // javac will parse the unicode control sequence, it will be passed to the parser as a raw unicode character
+            "{\"validJson\":\"\u274c\",\"right\", \"here\"}",
+            26,
+            24,
+            1,
+            25
```
hal7df, line 230: Parser backends aren't consistent about the column in this case, because the exact definition of "column" seems to depend on whether the parser is operating on a byte array or a character array. I'm open to suggestions about what to do with this case.

cowtowncoder: Column should be exactly same for byte/char -backed input, for ASCII characters (for multi-byte characters there is indeed difference). But I think few tests use characters outside ASCII range..

src/test/java/com/fasterxml/jackson/failing/read/LocationOfError1173Test.java
```diff
@@ -0,0 +1,326 @@
...
+        ),
+        ASYNC(
+            (String input) -> {
+                JsonParser parser = JSON_F.createNonBlockingByteArrayParser();
+                ByteArrayFeeder feeder = (ByteArrayFeeder) parser.getNonBlockingInputFeeder();
+                assertTrue(feeder.needMoreInput());
+
+                byte[] inputBytes = input.getBytes(StandardCharsets.UTF_8);
+                feeder.feedInput(inputBytes, 0, inputBytes.length);
+                feeder.endOfInput();
```
cowtowncoder, line 59: I don't think this works the way you'd expect... I think `endOfInput()` should not be called until all input is consumed.

cowtowncoder (approved):
Ok parameterization etc make things bit hard to follow but I assume things are fine.
I do have suspicion async-parser case won't work as expected but I'll merge and see how things fare.

src/test/java/com/fasterxml/jackson/failing/read/LocationOfError1173Test.java
cowtowncoder: Or maybe I misremember this part, as test appears to work fine :)

Issue #1117, opened 2023-09-30
Change default `RecylerPool` implementation to `newLockFreePool` (from `threadLocalPool`)


cowtowncoder:
With #1089 (and related) completed, Jackson 2.16 both allows for configuring alternate `RecyclerPool`s to use for buffer recycling and provides a set of alternatives, notably ones that do not use `ThreadLocal` for recycling.

To reduce likelihood of breakage, the default implementation for 2.16 remains ThreadLocal-based pool. This hopefully allows as to gather feedback, experiences by early adopters, on how alternative pools work (for example, Quarkus project is likely to experiment with alternatives). This, in turn, should let us decide on which alternative to use, along with global-vs-per-factory pools.
As the baseline, my icurrent thinking is that per-factory, unbounded/lock-free implementation might work well as the default.

cowtowncoder:
Ok, to get discussion started, I will propose we will use:

* `JsonRecyclerPool.newLockFreePool()`

as the default. This will have following benefits:

* Per-factory instances to reduce chance of global contention
* Has low minimal memory usage

As with any choice this will not be the optimal choice for all use cases. I am ready to hear alternative suggestions.

mariofusco:
Note that this is blocked by https://github.com/FasterXML/jackson-databind/issues/4321

cowtowncoder:
Note: issue @mariofusco outlined (thanks!) has been resolved now, so the default pool implementation may be changed.

PR #1187, opened 2024-01-14
Fix #1186: improve recycled buffer release logic

### LINKED ISSUE #1186: `BufferRecycler` should avoid setting replacement if one already returned, bigger
(based on discussion https://github.com/FasterXML/jackson/discussions/204 -- good suggestion by @kkkkkhhhh)

It seems that `BufferRecycler` will always replace assigned buffer when release method is called. But it would probably make sense to only replace `null` or smaller buffer, and avoid replacing bigger buffer with smaller.
While in the original expected usage sequence should always be "alloc / release / allow / release" (in which case "release" would be replacing `null`), there can be cases where this does not hold (multiple parsers/generators per thread, concurrently; but also just parser-and-generator case).

So let's add some basic checking into release method.

cowtowncoder:


pjfanning:
@cowtowncoder maybe I'm wrong but I still think we have a GC issue. https://github.com/FasterXML/jackson/discussions/204 mentions a GC issue. The new code still means that we need to GC an array - just possibly a different array (the smaller array). This still means GC. This code change may still be helpful generally but not be enough.

I think something more is at play. Without a reproducible test case, this is all supposition. Should we not ask ourselves why this user is getting so many extra arrays created in the first place? Ideally, we should only create so many arrays and we should have free slots to add them back to buffer recycler. I suppose that we might need to create a new array if the buffer recycler has no free entries (they are all in use).

pjfanning:
@cowtowncoder I think we are stuck if the OP won't give us more detail or a reproducible case. If you are using jackson-databind to deserialize lots of input files then you will be creating lots of Java instances, not just byte and char arrays. I see nothing in what the OP has shown that they have anything more than guesswork to point fingers specifically at the buffer recycling. There are no stats from Profiling tools. No heap dumps. Nada.

cowtowncoder:
@pjfanning right, I added similar note to #1186 (wrt this not avoiding possible GC for "extra" buffers).
And I agree that there is little evidence here. However, I do think that the change I made here is net positive.

But beside anything done in `BufferRecycler` container,  there is much more important work to be done for #1117: changing `RecyclerPool` default implementation.

The issue at that level is that with ThreadLocal-based pooling, there will only be single `BufferRecycler` instance per Thread. So if and when there are more than 1 instance of `JsonParser` / `JsonGenerator` used within same thread, reycling only avoids one of allocations per buffer-type.
But with all other (new) `ReyclerPool` implementations there will be distinct `BufferRecycler` per parser/generator.
This may lead to more buffer retention, but it should make recycling itself much more effective,

(sidenote: this is where exposing metrics for hit/recycle rate would be useful.. but that's much more work).

PR #1191, opened 2024-01-15
Fix #1190: add more testing for `JsonBufferRecylers` implementations

### LINKED ISSUE #1190: No tests for verifying basic functioning of `RecyclerPool` implementations (#1064)
(note: spun off of https://github.com/FasterXML/jackson-databind/issues/4321)

One major new feature in Jackson 2.16 was addition of pluggable `RecyclerPool`s (see #1064 f.ex).
But unfortunately their functioning was not tested beyond regression testing for the default, ThreadLocal-based legacy `RecyclerPool`.
Let's start by adding streaming-level testing first.

**EDIT**: there is actually one test -- `BufferRecyclerPoolTest` -- that verifies some aspects but is limited (only simple generation case, not parser), but we can do better.

cowtowncoder:


cowtowncoder:
@mariofusco Adding these just in case problem was reproducible at streaming level, but doesn't seem to be the case (which makes sense -- we did have some tests after all).

PR #1192, opened 2024-01-16
reuse in IOContext the BufferRecycler already contained in the ContentReference if any

mariofusco:
This is a tentative fix for the issue reported in https://github.com/FasterXML/jackson-databind/issues/4321

It tries to reuse the `BufferRecycler` already contained in the `ContentReference` (if any) also in the `IOContext`, instead of asking a new one from the pool regardless. In this case the `IOContext` is flagged for not being the owner of the `BufferRecycler`, so it won't attempt to mistakenly put it back into the pool when closed.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/JsonFactory.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonFactory.java
@@ -2185,8 +2185,12 @@ public class JsonFactory
         if (contentRef == null) {
             contentRef = ContentReference.unknown();
         }
+
+        BufferRecycler br = contentRef.getBufferRecycler();
+        boolean externalBufferRecycler = br != null;
         return new IOContext(_streamReadConstraints, _streamWriteConstraints, _errorReportConfiguration,
-                _getBufferRecycler(), contentRef, resourceManaged);
+                externalBufferRecycler ? br : _getBufferRecycler(), contentRef, resourceManaged)
+                .withExternalBufferRecycler(externalBufferRecycler);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


cowtowncoder:
Ok, first of all, thank you for proposing this patch.

My first thought is that `ContentReference` should not be used for this purpose. I realize it may be one of few convenient things being passed but it just... does not seem right conceptually.
So I would really like to figure out some other way.

mariofusco:
> My first thought is that `ContentReference` should not be used for this purpose. I realize it may be one of few convenient things being passed but it just... does not seem right conceptually. So I would really like to figure out some other way.

I agree that conceptually `ContentReference` is not made for this purpose. I wanted to achieve the goal of using the same `BufferRecycler` instance in both `IOContext` and `SegmentedStringWriter` in the less invasive and most backward compatible possible way. Please advice if you can find a better solution and let me know if I could do something else on this.

cowtowncoder:
@mariofusco Yes, we agree on goals. I have been trying to think of how to achieve this but haven't yet figured out a good way -- I think it is something that requires changes to maybe `JsonFactory` `createGenerator` method (adding new one(s)), and then `ObjectMapper` / `ObjectWriter` to use.

Approach would either be for `JsonFactory` to handle construction of `ByteArrayBuilder` / `SegmentedStringWriter` (and exposing via `JsonGenerator.getOutputTarget()`), or it accepting a callback for something similar to `Supplier<Writer>` / `Supplier<OutputStream>`.
Either way, the trick is that of getting shared `BufferRecycler` that `JsonGenerator` gets, to BAB/SSW.
There is also the sequencing challenge since `JsonGenerator` needs BAB/SSW to be given and thereby constructed first, before generator -- but `BufferRecycler` is currently accessed as part of generator construction.

cowtowncoder:
@mariofusco Ok, so I did #1193 as foundation for passing `BufferRecycler` along with `SegmentedStringWriter` and `ByteArrayBuilder`: the remaining part is extracting it out by `JsonFactory` and using. Plus hopefully verifying that usage works as expected (nothing breaks, first, and buffer reuse/recycling actually works which will be trickier to test).

cowtowncoder:
Ok, I ended up using quite a bit from this for #1196, the only meaningful difference "carrier" object; decided to use `Writer` / `OutputStream`.

PR #1196, opened 2024-01-19
Fix #1195: add actual `BufferRecycler` reuse

### LINKED ISSUE #1195: Use `BufferRecycler` provided by input (`InputStream`, `Reader`) or output (`OutputStream`, `Writer`) object if available
(note: follow-up to #1193)

Currently `BufferRecycler` passed to `JsonGenerator` and `JsonParser` is explicitly accessed from `RecyclerPool` configured with `JsonFactory`. But there are cases where `jackson-databind` has already allocated an instance for use (currently mostly for output aggregation) and it would make sense to allow that to be used instead.
Let's do this by checking if input source/output target implements `BufferRecycler.Gettable` to expose pre-configured `BufferRecycler`.

cowtowncoder:


cowtowncoder:
@mariofusco I think this is quite similar to what you suggested and should work well (not perfect obviously, but relatively limited changes). WDYT?
I think life-cycle wrt pool will work ok in this particular case (that is, use of ByteArrayBuilder/SegmentedStringWriter by databind), even if not quite as clean as ideally it'd be.

src/main/java/com/fasterxml/jackson/core/JsonFactory.java
```diff
@@ -2181,12 +2181,24 @@ public RecyclerPool<BufferRecycler> _getRecyclerPool() {
...
+            br = (content instanceof BufferRecycler.Gettable)
+                    ? ((BufferRecycler.Gettable) content).bufferRecycler()
+                    : null;
+        }
+        if (br == null) {
+            br = _getBufferRecycler();
         }
         return new IOContext(_streamReadConstraints, _streamWriteConstraints, _errorReportConfiguration,
-                _getBufferRecycler(), contentRef, resourceManaged);
+                br, contentRef, resourceManaged);
```
mariofusco, line 2201: This is very similar to what I did in my pull request, but in my opinion there is one important missing thing here: if the `BufferRecycler` hasn't been taken from the pool here but you're reusing the one coming from the `BufferRecycler.Gettable` [the IOContext should be aware that it is not the main owner of the resource](https://github.com/FasterXML/jackson-core/pull/1192/files#diff-6cb3e29fa26f4258ee5ad42c9a0a2bd91e053d2a7e7dbe16b67230dcf542ab14R2193) and [in this case it should NOT be its reposibility to put the `BufferRecycler` back into the pool when closed](https://github.com/FasterXML/jackson-core/pull/1192/files#diff-a252aa8da98b5f0fe4dfd326d160337c010be5921607f0a64df3bb9634779736R469).

cowtowncoder: That's part that is not ideal: in this case ownership is assumed to be passed over. I'll have to think about this a bit.
For existing implementations difference does not matter but other pools could assume this important.

The reason it should not matter a lot, at this point, is due to synchronization in`BufferRecycler`.

cowtowncoder: And yes, this is very similar: I liked many parts so wanted to start with those. And will likely copy the ownership part wrt `IOContext` too.

cowtowncoder: Added possibility of external life-cycle.

PR #1197, opened 2024-01-21
Add new `JsonWriteFeature.ESCAPE_FORWARD_SLASHES`

### LINKED ISSUE #507: Add `JsonWriteFeature.ESCAPE_FORWARD_SLASHES` to allow escaping of '/' for String values
Jackson 2.x only escapes minimum set of characters, as defined by JSON specification. This does not include forward slash character ('/'). But while legal, it turns out that more often than not users do want escaping, to guard against potential inclusion-in-HTML problems, particularly for embedded JSON constants in Javascript sources, in script tags.

Now: although it is possible to enable escaping already (via `CharacterEscapes`), it is bit verbose, and also adds some measurable (not huge, but not completely trivial) overhead.
So for 3.0 let's add this character as escape-by-default, but also add a simple mechanism for turning that off if feasible (`JsonWriteFeature`, most likely?).

JooHyukKim:
resolves #507

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
@@ -7,6 +7,7 @@ package com.fasterxml.jackson.core;
 import java.io.*;
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.util.Arrays;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
@@ -14,6 +15,7 @@ import java.util.concurrent.atomic.AtomicLong;
 import com.fasterxml.jackson.core.JsonParser.NumberType;
 import com.fasterxml.jackson.core.exc.StreamWriteException;
 import com.fasterxml.jackson.core.io.CharacterEscapes;
+import com.fasterxml.jackson.core.json.JsonWriteFeature;
 import com.fasterxml.jackson.core.type.WritableTypeId;
 import com.fasterxml.jackson.core.type.WritableTypeId.Inclusion;
 import com.fasterxml.jackson.core.util.JacksonFeatureSet;
@@ -32,6 +34,12 @@ import static com.fasterxml.jackson.core.JsonTokenId.*;
 public abstract class JsonGenerator
     implements Closeable, Flushable, Versioned
 {
+
+    /**
+     * @since 2.17
+     */
+    protected final static int INT_FORWARD_SLASH = 47;
+
     /**
      * Default set of {@link StreamWriteCapability}ies that may be used as
      * basis for format-specific readers (or as bogus instance if non-null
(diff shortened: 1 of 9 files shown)


JooHyukKim:
I figured we should do https://github.com/FasterXML/jackson-core/pull/1198 separately.

src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
```diff
@@ -2977,4 +3008,18 @@ protected void _writeSimpleObject(Object value) throws IOException
...
                 +value.getClass().getName()+")");
     }
+
+    /**
+     * Returns a new {@code int[]} of escape table with additional forward-slash escaping configured
+     * as configured by {@link JsonWriteFeature#ESCAPE_FORWARD_SLASHES}.
+     *
+     * @since 2.17
+     */
+    protected int[] escapeForwardSlash(int[] escapes) {
```
cowtowncoder, line 3018: Should not be added here since this is not JSON-specific type despite name. So rather add in `JsonGeneratorImpl`, shared base class for concrete ones

src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
```diff
@@ -2977,4 +3008,18 @@ protected void _writeSimpleObject(Object value) throws IOException
...
+        int[] esc = Arrays.copyOf(escapes, escapes.length);
```
cowtowncoder, line 3019: I think this should be optimized a bit by checking if any change is actually needed -- and only making defensive copy in that case.

However, this leads to my main concern here: copying of the array for every generator instance adds overhead, which can be non-trivial in case of writing small documents.
So I wonder if there's any way to avoid this overhead by pre-creating escaping table in `CharTypes`.

cowtowncoder (commented):
I think this is a good starting point: I can make changes I think are needed wrt placement of methods. The remaining question then is about avoiding overhead of copying the `_outputEscapes` encoding array for every generator.

cowtowncoder:
Ok, I changed things I wanted to change.

Looking at how escape-table is created, I don't think it is easy to try to make it happen by `JsonFactory` so let's forget about that wrt 2.x.

But it might still make sense to add an overload for `CharTypes.get7BitOutputEscapes()` (one that takes second argument, if forward slash is to be escaped), and be able to use eagerly constructed table unless (except if non-standard quote char is specified).

cowtowncoder:
@JooHyukKim Ok, I rewrote this a bit and I think it might be ready for merging. Thank you for providing the base & let me know if you think some changes are needed. Otherwise (or after changing) if you can convert it from draft to regular PR I can go ahead and merge it.

Thank you again!

JooHyukKim:
@cowtowncoder seems okay now 👍🏼

cowtowncoder:
Thank you @JooHyukKim. I'll go ahead and merge this then.

cowtowncoder (approved):

cowtowncoder:
Follow-up wrt 3.0: #1200

PR #1198, opened 2024-01-21
Add new method `JsonGenerator.isEnabled(JsonWriteFeature)`

JooHyukKim:
blocks #1197


cowtowncoder:
As per my note on the other PR: let's not add this -- there is a reason this method is not included in `JsonGenerator`.
Naming here is bit unfortunate, but basically:

1. `JsonGenerator` is API for generators, NOT specific to. JSON, but
2. `JsonWriteFeature` IS JSON-specific, and should not be referenced for general, non-JSON-specific code

Instead, access from actual JSON-backed implementations (`JsonGeneratorImpl`) could have this method -- but it is not necessarily even needed (existing code doesn't use this).

JooHyukKim:
> As per my note on the other PR: let's not add this -- there is a reason this method is not included in `JsonGenerator`. Naming here is bit unfortunate, but basically:

Reasonable. CLosed this accordingly.

PR #1203, opened 2024-01-25
Faster division by 1000

xtonik:
**Division by 1000** was replaced with more performant custom method using multiply and add. I suppose that this tiny method will be inlined, otherwise the code must be copied to places where is called.

Another similar simple optimization was also done - **multiplication of int by constant `0.80`** in integer (exactly in `long`) domain.

The validity is **verified by attached tests**. They are provided as a comparison to original implementation and thus they don't need to be executed again and again, so they are marked as a ignored not to cause running tests slowly - on my PC they take around 10 s.

**No behavior changes were done, performance improvement only.**

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/io/NumberOutput.java
+++ b/src/main/java/com/fasterxml/jackson/core/io/NumberOutput.java
@@ -245,6 +245,12 @@ public final class NumberOutput
         return _outputFullBillion((int) v, b, off);
     }
 
+    /**
+     * Optimized code for integer division by 1000; typically 50% higher
+     * throughput for calculation
+     *
+     * @since 2.17
+     */
     static int divBy1000(int number) {
         return (int) (number * 274_877_907L >>> 38);
     }
--- a/src/main/java/com/fasterxml/jackson/core/io/NumberOutput.java
+++ b/src/main/java/com/fasterxml/jackson/core/io/NumberOutput.java
@@ -246,7 +246,7 @@ public final class NumberOutput
     }
 
     static int divBy1000(int number) {
-        return (int) (number * 274877907L >>> 38);
+        return (int) (number * 274_877_907L >>> 38);
     }
 
     /*
--- a/src/main/java/com/fasterxml/jackson/core/io/NumberOutput.java
+++ b/src/main/java/com/fasterxml/jackson/core/io/NumberOutput.java
@@ -85,7 +85,7 @@ public final class NumberOutput
                 }
                 return _leading3(v, b, off);
             }
-            int thousands = v / 1000;
+            int thousands = divBy1000(v);
             v -= (thousands * 1000); // == value % 1000
             off = _leading3(thousands, b, off);
             off = _full3(v, b, off);
@@ -107,10 +107,10 @@ public final class NumberOutput
             }
             return _outputFullBillion(v, b, off);
         }
-        int newValue = v / 1000;
+        int newValue = divBy1000(v);
         int ones = (v - (newValue * 1000)); // == value % 1000
         v = newValue;
-        newValue /= 1000;
+        newValue = divBy1000(newValue);
         int thousands = (v - (newValue * 1000));
 
         off = _leading3(newValue, b, off);
@@ -136,7 +136,7 @@ public final class NumberOutput
                     off = _leading3(v, b, off);
                 }
             } else {
-                int thousands = v / 1000;
+                int thousands = divBy1000(v);
                 v -= (thousands * 1000); // == value % 1000
                 off = _leading3(thousands, b, off);
                 off = _full3(v, b, off);
@@ -153,10 +153,10 @@ public final class NumberOutput
             }
             return _outputFullBillion(v, b, off);
         }
-        int newValue = v / 1000;
+        int newValue = divBy1000(v);
         int ones = (v - (newValue * 1000)); // == value % 1000
         v = newValue;
-        newValue /= 1000;
+        newValue = divBy1000(newValue);
         int thousands = (v - (newValue * 1000));
         off = _leading3(newValue, b, off);
         off = _full3(thousands, b, off);
@@ -243,6 +243,10 @@ public final class NumberOutput
             off = _outputFullBillion((int) upper, b, off);
         }
         return _outputFullBillion((int) v, b, off);
+    }
+
+    static int divBy1000(int number) {
+        return (int) (number * 274877907L >>> 38);
     }
 
     /*
@@ -359,13 +363,13 @@ public final class NumberOutput
             if (v < 1000) {
                 return _leading3(v, b, off);
             }
-            int thousands = v / 1000;
+            int thousands = divBy1000(v);
             int ones = v - (thousands * 1000); // == value % 1000
             return _outputUptoMillion(b, off, thousands, ones);
         }
-        int thousands = v / 1000;
+        int thousands = divBy1000(v);
         int ones = (v - (thousands * 1000)); // == value % 1000
-        int millions = thousands / 1000;
+        int millions = divBy1000(thousands);
         thousands -= (millions * 1000);
 
         off = _leading3(millions, b, off);
@@ -385,9 +389,9 @@ public final class NumberOutput
 
     private static int _outputFullBillion(int v, char[] b, int off)
     {
-        int thousands = v / 1000;
+        int thousands = divBy1000(v);
         int ones = (v - (thousands * 1000)); // == value % 1000
-        int millions = thousands / 1000;
+        int millions = divBy1000(thousands);
 
         int enc = TRIPLET_TO_CHARS[millions];
         b[off++] = (char) (enc >> 16);
@@ -414,13 +418,13 @@ public final class NumberOutput
             if (v < 1000) {
                 return _leading3(v, b, off);
             }
-            int thousands = v / 1000;
+            int thousands = divBy1000(v);
             int ones = v - (thousands * 1000); // == value % 1000
             return _outputUptoMillion(b, off, thousands, ones);
         }
-        int thousands = v / 1000;
+        int thousands = divBy1000(v);
         int ones = (v - (thousands * 1000)); // == value % 1000
-        int millions = thousands / 1000;
+        int millions = divBy1000(thousands);
         thousands -= (millions * 1000);
 
         off = _leading3(millions, b, off);
@@ -440,9 +444,9 @@ public final class NumberOutput
 
     private static int _outputFullBillion(int v, byte[] b, int off)
     {
-        int thousands = v / 1000;
+        int thousands = divBy1000(v);
         int ones = (v - (thousands * 1000)); // == value % 1000
-        int millions = thousands / 1000;
+        int millions = divBy1000(thousands);
         thousands -= (millions * 1000);
 
         int enc = TRIPLET_TO_CHARS[millions];
--- a/src/main/java/com/fasterxml/jackson/core/sym/ByteQuadsCanonicalizer.java
+++ b/src/main/java/com/fasterxml/jackson/core/sym/ByteQuadsCanonicalizer.java
@@ -1096,6 +1096,7 @@ public final class ByteQuadsCanonicalizer
         return offset;
     }
 
+    // @since 2.17
     static int multiplyByFourFifths(int number) {
         return (int) (number * 3_435_973_837L >>> 32);
     }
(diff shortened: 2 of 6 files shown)


src/test/java/com/fasterxml/jackson/core/io/NumberOutputTest.java
```diff
@@ -0,0 +1,19 @@
...
+import static org.junit.Assert.assertEquals;
+
+public class NumberOutputTest {
+    @Test
+    @Ignore
+    public void testDivBy1000() {
+        for (long i = 0; i <= Integer.MAX_VALUE; i++) {
+            int number = (int) i;
+            int expected = number / 1000;
+            int actual = NumberOutput.divBy1000(number);
```
pjfanning, line 15: can you replace this test by 1 that tests a few common cases and some edge cases - you should also test the FourFifths code.

src/test/java/com/fasterxml/jackson/core/sym/ByteQuadsCanonicalizerTest.java
```diff
@@ -0,0 +1,19 @@
...
+import static org.junit.Assert.assertEquals;
+
+public class ByteQuadsCanonicalizerTest {
+    @Test
+    @Ignore
+    public void testMultiplyByFourFifths() {
+        for (long i = 0; i <= Integer.MAX_VALUE; i++) {
+            int number = (int) i;
+            int expected = (int) (number * 0.80);
+            int actual = ByteQuadsCanonicalizer.multiplyByFourFifths(number);
```
pjfanning, line 15: this is not a test we want in the source repo - write a test that tests a few important values.

pjfanning:
@xtonik thanks for the changes but this code would need to be sigificantly faster in order to justify the extra complexity - at least for me.

src/main/java/com/fasterxml/jackson/core/io/NumberOutput.java
```diff
@@ -245,6 +245,10 @@ public static int outputLong(long v, byte[] b, int off)
         return _outputFullBillion((int) v, b, off);
     }
 
+    static int divBy1000(int number) {
```
pjfanning, line 248: could mark this as final? likewise with the four-fifths method

src/test/java/com/fasterxml/jackson/core/io/NumberOutputTest.java
xtonik: That's why I market the tests as ignored. **They are exhaustive to be really sure, that the changes are valid not to break the code** when the change is really tiny and so important. Do you have some suggestion what values are important for this case? **What I can imagine is small values including zero, some (randomly selected?) from the middle and values near `Integer.MAX_VALUE`,** but still I have to say that so reduced tests are not enough to persuade at least me, that the code works.

EOLs I will add of course.

src/test/java/com/fasterxml/jackson/core/sym/ByteQuadsCanonicalizerTest.java
xtonik: Reply in previous comment

src/main/java/com/fasterxml/jackson/core/io/NumberOutput.java
xtonik: Of course I can, doesn't they supposed as being effectively final and thus treated as "real" final?

src/test/java/com/fasterxml/jackson/core/io/NumberOutputTest.java
pjfanning: an ignored test is a useless test - a test that validates thousands of values is not a test that we can add to CI

pjfanning: no test is a -1 from me

src/main/java/com/fasterxml/jackson/core/io/NumberOutput.java
xtonik: Yes, it is tiny improvement, but as I saw the same principle (division by multiply and add) used in the project, so I expected, that this changes will be welcomed. If you look to my fork of this repo, I have prepared several PRs of this kind, so it will be convenient to know beforehand if I even should make another proposals like this not to waste our time.

pjfanning: I'll let @cowtowncoder decide. For me, this sort of enhancement is difficult to maintain. If it is absolutely much faster then maybe this is worth it. We would need benchmarks.

xtonik: I can prepare benchmarks also, but as the improvement are tiny, they need to be focused on particular actions to be visible, but as you wish - I will look at _jackson-benchmarks_ and will try to find some convincing benchmarks or will try to create some for all my stuff. Then I filter my PRs to only ones, whose are "valuable".

src/test/java/com/fasterxml/jackson/core/io/NumberOutputTest.java
xtonik: **These test just only verify the validity of the changes and I agree they should not be included in CI.** I expected, that some responsible person will use them for verification only and remove them from repository before merge.

For this task is hard to imagine, for which numbers the results differ to ones returned by original code and thus what values are important. There is many tasks like this, where is convenient to perform some exhaustive testing.
Someone with more than basic mathematical background can predict them via manual calculation, but how you know that he did not some mistake?

I can easily rework them to be included in CI, but as I said the tests if they do not test whole integer range are not worthy.

cowtowncoder:
@xtonik  Is there some article or reference showing the idea behind technique? I can believe that FP division could be faster than integer division (FPUs boost fp arithmetics). Or links to other projects doing this?

src/main/java/com/fasterxml/jackson/core/sym/ByteQuadsCanonicalizer.java
```diff
@@ -1096,13 +1096,17 @@ private int _resizeAndFindOffsetForAdd(int hash) throws StreamConstraintsExcepti
         return offset;
     }
 
+    static int multiplyByFourFifths(int number) {
```
cowtowncoder, line 1099: Let's not do this -- this is not performance critical in any way, no point optimizing in this case (whereas `NumberOutput` is indeed performance sensitive)

src/test/java/com/fasterxml/jackson/core/io/NumberOutputTest.java
cowtowncoder: Tests that are not included by CI are, IMO, not very useful. So I agree with @pjfanning .
But I guess I can see the point of one-off before merging.

Still, in addition, I'd probably run through 1 .. 1M, them sampling, using `Random` or something.

src/test/java/com/fasterxml/jackson/core/sym/ByteQuadsCanonicalizerTest.java
```diff
@@ -0,0 +1,19 @@
+package com.fasterxml.jackson.core.sym;
+
+import org.junit.Ignore;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class ByteQuadsCanonicalizerTest {
```
cowtowncoder, line 8: Let's drop this test, along with optimization.

cowtowncoder:
I think I am open to change, provided there's something published about technique (ideally it's a well-known technique that I wasn't familiar with). Let's drop 80% calculation as it's not worth optimizing.

It'd be good to have `jmh`-based micro-benchmark as well; maybe I'll modify 

https://github.com/cowtowncoder/misc-micro-benchmarks

with it.

src/main/java/com/fasterxml/jackson/core/sym/ByteQuadsCanonicalizer.java
xtonik: The technique is used [in this project](https://github.com/FasterXML/jackson-core/blob/dc5b746f33aa96235040c291d910da15692504cd/src/main/java/com/fasterxml/jackson/core/io/schubfach/FloatToDecimal.java#L548) also among many [others](https://github.com/plokhotnyuk/jsoniter-scala/blob/7b3de2176d0e15457933e87d0080914545bca08a/jsoniter-scala-core/jvm/src/main/scala/com/github/plokhotnyuk/jsoniter_scala/core/JsonReader.scala#L2995). I have also verified this fact via my own benchmarks.

src/test/java/com/fasterxml/jackson/core/io/NumberOutputTest.java
xtonik: Yep, random select is good also as there is not only one number, which does not fit well when the values is greater than some threshold for selected magical constants for multiplication and bitshift. In fact the values for which is expression gives invalid results are repeating with period of divisor.

cowtowncoder:
Ok. So a quick test with `misc-microbenchmarks` gives these numbers with JDK 17:

```
Benchmark                      Mode  Cnt  Score   Error  Units
DivBy1000.perfHandOptimized   thrpt    5  5.022 ± 0.037  ops/s
DivBy1000.perfSimpleDivision  thrpt    5  3.566 ± 0.029  ops/s
```

that is, +50% throughput for operation itself. With JDK 8 bit less; and for some reason I don't quite know, no speedup with JDK 21 (maybe the "unoptimized" case is actually optimized by JDK under the hood).

Whether this translates into measurable improvement for number serialization is open question; I guess it shouldn't hurt.

Also: I modified unit test a bit by moving String concatenation out of busy loop and test over full `int` range takes only about 2 seconds. I'll probably tweak it a bit to have automated, always run part, and separate manually run full test.

cowtowncoder:
@xtonik I think we can make this happen, and although it's hard to know how meaningful for real use improvements are, change is still relatively small and tested (I modified unit test to run through 1/7 of full `int` range and leave on for CI).

But one thing before I can merge this is that if we don't have a CLA for you, we'd need one (just needs to be sent once before the first contribution). It's this one:

and usually easiest to print, fill & sign, scan/photo; then email `cla` at fasterxml dot com.
Once I get it I can proceed with last clean up and merge.

Looking forward to getting this merged, thank you in advance!

xtonik:
> @xtonik I think we can make this happen, and although it's hard to know how meaningful for real use improvements are, change is still relatively small and tested (I modified unit test to run through 1/7 of full `int` range and leave on for CI).
> 
> But one thing before I can merge this is that if we don't have a CLA for you, we'd need one (just needs to be sent once before the first contribution). It's this one:
> 
> https://github.com/FasterXML/jackson/blob/master/contributor-agreement.pdf
> 
> and usually easiest to print, fill & sign, scan/photo; then email `cla` at fasterxml dot com. Once I get it I can proceed with last clean up and merge.
> 
> Looking forward to getting this merged, thank you in advance!

Ok, I will subscribe it.

xtonik:
> Ok. So a quick test with `misc-microbenchmarks` gives these numbers with JDK 17:
> 
> ```
> Benchmark                      Mode  Cnt  Score   Error  Units
> DivBy1000.perfHandOptimized   thrpt    5  5.022 ± 0.037  ops/s
> DivBy1000.perfSimpleDivision  thrpt    5  3.566 ± 0.029  ops/s
> ```
> 
> that is, +50% throughput for operation itself. With JDK 8 bit less; and for some reason I don't quite know, no speedup with JDK 21 (**maybe the "unoptimized" case is actually optimized by JDK under the hood**).
> 
> Whether this translates into measurable improvement for number serialization is open question; I guess it shouldn't hurt.
> 
> Also: I modified unit test a bit by moving String concatenation out of busy loop and test over full `int` range takes only about 2 seconds. I'll probably tweak it a bit to have automated, always run part, and separate manually run full test.

As you said, the compiler for new Java versions does more optimizations than older ones. The same situation (not remembering from which version) is probably for `String.replace(String, String)`, when given string is one character long, then it is replaced by `String.replace(char, char)`. This is what is about [one of my prepared PR](https://github.com/FasterXML/jackson-core/compare/2.17...xtonik:jackson-core:REPLACE_CHAR_NOT_STRING_BY_REGEX).

xtonik:
To get more familiar where the magic number came from, there are registered sequences [A346495](https://oeis.org/A346495) and [A346496](https://oeis.org/A346496) for this.

xtonik:
I have just sent my CLA to given email as requested.

If someone is interested and have some time, I have prepared [another changes](https://github.com/xtonik/jackson-core/branches/active). As they are tiny, it takes couple of minutes to check if some of them is valuable to create PRs.

PR #1212, opened 2024-02-11
Start converting some unit tests to JUnit5; add new test base class

cowtowncoder:


JooHyukKim:
Are we having a base test class here? In jackson-databind wr have been doing static imports from DatabindTestUtil

cowtowncoder:
Yeah the only reason I didn't suggest it there is that I forgot that the problem with base class was extending JUnit 4 `TestCase` (or such). Otherwise base class seems to require slightly fewer imports.

For databind it is actually possible to use `DatabindTestUtil` as the base class too (although not necessarily clean).

PR #1220, opened 2024-02-15
Prefer bitwise OR over addition for UTF-8 multi-byte encoding

cowtowncoder:
(contributed by @xtonik)


cowtowncoder:
@xtonik Are bitwise logical operators faster than basic arithmetic (add, sub)? Back in the days costs were similar if not identical, but I haven't really been following on how CPU instruction costs fare these days.

So, basically, I assume this is expected to improve performance, but not sure how to verify: could build a micro-benchmark I guess but was guessing you have more information onthis.

cowtowncoder:
@xtonik Forgot to follow up on this -- do you have something (tests, articles) to support the idea that this helps speed up processing?

PR #1222, opened 2024-02-15
Improve UTF-8 encoding by rearranging sufficient space check

cowtowncoder:
(contributed by @xtonik)

...also arguably slightly improves readability of code.


cowtowncoder:
Oh. No, this fails unit tests and that's because it's not functionally equivalent -- `_allocMore()` does NOT resize current block; the underlying buffers are segmented.
So allocation must fully align with byte boundary, otherwise there'll be one or more "empty" slots in the last buffer.

So, cannot do this optimization.

Issue #1229, opened 2024-02-29
Add char[] versions for NumberInput parseFloat, parseDouble, parseBigInteger


devinrsmith:
I've got a high performance streaming use case where I'm handling some of the lower-level details. For example, translating a `VALUE_STRING` into a `double`. I'm relying on `NumberInput` with the appropriate `useFastParser` config, and would like to avoid extraneous `String` allocation when possible. The `BigDecimal` case is currently the most complete, as it works with `JsonParser#hasTextCharacters`:

```java
static BigDecimal parseStringAsBigDecimal(JsonParser parser) throws IOException {
    final boolean useFastParser = parser.isEnabled(StreamReadFeature.USE_FAST_BIG_NUMBER_PARSER);
    return parser.hasTextCharacters()
            ? NumberInput.parseBigDecimal(parser.getTextCharacters(), parser.getTextOffset(), parser.getTextLength(), useFastParser)
            : NumberInput.parseBigDecimal(parser.getText(), useFastParser);
}
```

The other similar cases aren't able to take advantage since the underlying `char[]` parsing methods aren't exposed:

```java
static float parseStringAsFloat(JsonParser parser) throws IOException {
    // TODO: improve when parser.hasTextCharacters()
    return NumberInput.parseFloat(parser.getText(), parser.isEnabled(StreamReadFeature.USE_FAST_DOUBLE_PARSER));
}

static double parseStringAsDouble(JsonParser parser) throws IOException {
    // TODO: improve when parser.hasTextCharacters()
    return NumberInput.parseDouble(parser.getText(), parser.isEnabled(StreamReadFeature.USE_FAST_DOUBLE_PARSER));
}

static BigInteger parseStringAsBigInteger(JsonParser parser) throws IOException {
    // TODO: improve when parser.hasTextCharacters()
    return NumberInput.parseBigInteger(parser.getText(), parser.isEnabled(StreamReadFeature.USE_FAST_BIG_NUMBER_PARSER));
}
```

It would be great to add `char[]` versions `NumberInput#parseFloat`,  `NumberInput#parseDouble`, and  `NumberInput#parseBigInteger`.

cowtowncoder:
Might make sense, esp. if performance benchmark shows meaningful improvement from avoided allocations (or rather, gc time). Challenges:

1. Number of permutations wrt fast/JDK parser
2. JDK implementation only takes `String` (which is probably why there's no `char[]` taking alternative), so might end up being no-win for default case

[...]

PR #1230, opened 2024-03-01
Improve perf of float and double parsing in TextBuffer

pjfanning:
* just an outline - not for merging
* relates to #1229 
* In TextBuffer, when parsing floats and doubles, avoid creating Strings first
* copies the logic of contentsAsString into the float and double methods so that we can best handle the various scenarios
  * if _resultString already exists, use it
  * if _resultArray already exists, parse from the array instead of creating a String
  * in the case that multiple segments are needed, we might as well fall back to contentsAsString in this case because we need to build a String with a StringBuilder anyway 

@cowtowncoder is this a long the right lines? If so, I can add tests. If we get this done for floats and doubles, we can come back to make similar changes for other number types. 

I'm aware that this does not address the deferred number parsing in ParserBase (where numbers can be stored as `_numberString` until we know what type we need). Storing the number as the wrong number type and later converting it to a different number type can lead to number conversion issues and can be more expensive.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/io/NumberInput.java
+++ b/src/main/java/com/fasterxml/jackson/core/io/NumberInput.java
@@ -401,6 +401,32 @@ public final class NumberInput
     }
 
     /**
+     * @param array a char array containing a number to parse
+     * @param useFastParser whether to use {@code FastDoubleParser}
+     * @return closest matching double
+     * @throws NumberFormatException if string cannot be represented by a double
+     * @since v2.18
+     */
+    public static double parseDouble(final char[] array, final boolean useFastParser) throws NumberFormatException {
+        return parseDouble(array, 0, array.length, useFastParser);
+    }
+
+    /**
+     * @param array a char array containing a number to parse
+     * @param offset the offset to apply when parsing the number in the char array
+     * @param len the length of the number in the char array
+     * @param useFastParser whether to use {@code FastDoubleParser}
+     * @return closest matching double
+     * @throws NumberFormatException if string cannot be represented by a double
+     * @since v2.18
+     */
+    public static double parseDouble(final char[] array, final int offset,
+                                     final int len, final boolean useFastParser) throws NumberFormatException {
+        return useFastParser ? JavaDoubleParser.parseDouble(array, offset, len) :
+                Double.parseDouble(new String(array, offset, len));
+    }
+
+    /**
      * @param s a string representing a number to parse
      * @return closest matching float
      * @throws NumberFormatException if string cannot be represented by a float where useFastParser=false
@@ -426,6 +452,32 @@ public final class NumberInput
             return JavaFloatParser.parseFloat(s);
         }
         return Float.parseFloat(s);
+    }
+
+    /**
+     * @param array a char array containing a number to parse
+     * @param useFastParser whether to use {@code FastDoubleParser}
+     * @return closest matching float
+     * @throws NumberFormatException if string cannot be represented by a float
+     * @since v2.18
+     */
+    public static float parseFloat(final char[] array, final boolean useFastParser) throws NumberFormatException {
+        return parseFloat(array, 0, array.length, useFastParser);
+    }
+
+    /**
+     * @param array a char array containing a number to parse
+     * @param offset the offset to apply when parsing the number in the char array
+     * @param len the length of the number in the char array
+     * @param useFastParser whether to use {@code FastDoubleParser}
+     * @return closest matching float
+     * @throws NumberFormatException if string cannot be represented by a float
+     * @since v2.18
+     */
+    public static float parseFloat(final char[] array, final int offset,
+                                   final int len, final boolean useFastParser) throws NumberFormatException {
+        return useFastParser ? JavaFloatParser.parseFloat(array, offset, len) :
+                Float.parseFloat(new String(array, offset, len));
     }
 
     /**
--- a/src/main/java/com/fasterxml/jackson/core/io/NumberInput.java
+++ b/src/main/java/com/fasterxml/jackson/core/io/NumberInput.java
@@ -405,7 +405,7 @@ public final class NumberInput
      * @param useFastParser whether to use {@code FastDoubleParser}
      * @return closest matching double
      * @throws NumberFormatException if value cannot be represented by a double
-     * @since v2.18
+     * @since 2.18
      */
     public static double parseDouble(final char[] array, final boolean useFastParser) throws NumberFormatException {
         return parseDouble(array, 0, array.length, useFastParser);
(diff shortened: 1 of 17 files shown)


src/main/java/com/fasterxml/jackson/core/io/NumberInput.java
```diff
@@ -400,6 +400,32 @@ public static double parseDouble(final String s, final boolean useFastParser) th
         return useFastParser ? JavaDoubleParser.parseDouble(s) : Double.parseDouble(s);
     }
 
+    /**
+     * @param array a char array containing a number to parse
+     * @param useFastParser whether to use {@code FastDoubleParser}
+     * @return closest matching double
+     * @throws NumberFormatException if value cannot be represented by a double
+     * @since v2.18
```
cowtowncoder, line 408: I think we use plain `2.18` for version

cowtowncoder:
@pjfanning Yes, this is what I had in mind. I'll have a look and suggest some changes. Since we are not (yet) changing calling from `ParserBase`, I think this could actually go in 2.17, even.

I also think that we can remove some of deferral because calls to, say, `getFloatValue()` does explicitly establish target type -- but this is easy enough to verify with more time wrt 2.18 once 2.17 is out.
But that question is separate from work to make `TextBuffer` pass more optimal representation.

src/main/java/com/fasterxml/jackson/core/util/TextBuffer.java
```diff
@@ -536,6 +536,37 @@ public char[] contentsAsArray() throws IOException {
      */
     public double contentsAsDouble(final boolean useFastParser) throws NumberFormatException {
         try {
+            if (_resultString == null) {
+                // Has array been requested? Can make a shortcut, if so:
+                if (_resultArray != null) {
+                    validateFPLength(_resultArray.length);
```
cowtowncoder, line 542: Not needed: `StreamReadConstraints.validateFPLength()` is called from `ParserBase.resetFloat()` as part of decoding. `ReaderBasedJsonParser, line 1465 f.ex calls:

```
        _textBuffer.resetWithShared(_inputBuffer, startPtr, len);
        // And there we have it!
        return resetFloat(neg, intLen, fractLen, expLen);
```

so this is already taken care of

src/main/java/com/fasterxml/jackson/core/util/TextBuffer.java
```diff
@@ -536,6 +536,37 @@ public char[] contentsAsArray() throws IOException {
...
+                    return NumberInput.parseDouble(_resultArray, useFastParser);
+                } else {
+                    // Do we use shared array?
+                    if (_inputStart >= 0) {
+                        if (_inputLen < 1) {
```
cowtowncoder, line 547: Similarly, decoder has already confirmed this won't be the case, no need to verify.

src/main/java/com/fasterxml/jackson/core/util/ReadConstrainedTextBuffer.java
```diff
@@ -26,4 +26,13 @@ protected void validateStringLength(int length) throws StreamConstraintsExceptio
     {
         _streamReadConstraints.validateStringLength(length);
     }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    protected void validateFPLength(int length) throws StreamConstraintsException
```
cowtowncoder, line 34: Won't be needed, as per my comments (parsers will call `validateFPLength()` around same time as setting `TextBuffer` contents; and this only if number decoded is lexically valid (will not reset to empty String).

src/main/java/com/fasterxml/jackson/core/util/TextBuffer.java
pjfanning: when I removed this I got a NPE in a new test (empty TextBuffer)

cowtowncoder: Right but we are not building a general purpose functionality: Parsers will never call (and are specifically not to ever call) method without populating buffer with lexically valid number.

src/test/java/com/fasterxml/jackson/core/util/TestTextBuffer.java
```diff
@@ -192,4 +196,62 @@ public void testGetSizeFinishCurrentSegmentAndResetWith() throws Exception {
...
+    }
+
+    public void testContentsAsFloatFastParser() throws IOException {
+        TextBuffer textBuffer = new TextBuffer(null);
+        textBuffer.resetWithString("1.2345678");
+        assertEquals(1.2345678f,  textBuffer.contentsAsFloat(true));
+    }
+
+    public void testEmptyContentsAsDouble() {
+        TextBuffer textBuffer = new TextBuffer(null);
```
cowtowncoder, line 226: Not legal test as per usage patterns: not just for empty case but more generally, parsers will tokenize numbers using JSON (etc) pattern that describes valid numbers.
This simplifies decoding in cases decoder knows it need not validate input (such as our optimized `int`/`long` decoding logic).

pjfanning: If `new TextBuffer(null)` is illegal, can we update the constructor to throw an exception?

cowtowncoder: Yes, that'd make sense

cowtowncoder: Actually no n/m, that's `BufferRecycler` -- `null` is accepted.

cowtowncoder (approved):

cowtowncoder:
I think I'll go ahead and merge this: it will be needed to tackle #1229.

src/main/java/com/fasterxml/jackson/core/util/TextBuffer.java
```diff
@@ -525,15 +524,41 @@ public char[] contentsAsArray() throws IOException {
...
+        // that do not require allocation first
+
+        // except _resultString first since it works best with JDK (non-fast parser)
+        if (_resultString != null) {
+            return NumberInput.parseDouble(_resultString, useFastParser);
+        }
+        if (_inputStart >= 0) { // shared?
+            return NumberInput.parseDouble(_inputBuffer, _inputStart, _inputLen, useFastParser);
+        }
+        if (_currentSize == 0) { // all content in current segment!
```
cowtowncoder, line 554: This was a bug, actually -- should use `_hasSegments` instead! Fixed in #1313.

src/main/java/com/fasterxml/jackson/core/util/TextBuffer.java
```diff
@@ -574,14 +599,41 @@ public float contentsAsFloat() throws NumberFormatException {
...
+        // that do not require allocation first
+
+        // except _resultString first since it works best with JDK (non-fast parser)
+        if (_resultString != null) {
+            return NumberInput.parseFloat(_resultString, useFastParser);
+        }
+        if (_inputStart >= 0) { // shared?
+            return NumberInput.parseFloat(_inputBuffer, _inputStart, _inputLen, useFastParser);
+        }
+        if (_currentSize == 0) { // all content in current segment!
```
cowtowncoder, line 629: Same here, fixed in #1313

PR #1241, opened 2024-03-16
Fix `NumberInput.looksLikeValidNumber()` implementation

pjfanning:
See https://github.com/FasterXML/jackson-databind/issues/4435

I'm not a regex expert and sort of hate them from a code readability and maintenance perspective.

This is my attempt at fixing the issue but the method is now so neutered, I really wonder if the method should be removed. It is also going to cause perf issues as the check is non-trivial.


EAlf91:
IMHO it can be removed completely. According to [RFC-8259](https://datatracker.ietf.org/doc/html/rfc8259#page-8) the rules for strings should be applied here for parsing from string inputs in jackson-databind and not number in this case. (see JsonTokenId.ID_STRING in [NumberDeserializers.java](https://github.com/FasterXML/jackson-databind/blob/d4570077445c34e2328df32a08b73d82ba0870ab/src/main/java/com/fasterxml/jackson/databind/deser/std/NumberDeserializers.java#L1026))

It seems that we're encountering this error because the prevalidation method _looksLikeValidNumber_ is misplaced for the given input, which is JSON-Type string and not number.

src/main/java/com/fasterxml/jackson/core/io/NumberInput.java
```diff
@@ -578,6 +578,7 @@ public static BigInteger parseBigIntegerWithRadix(final String s, final int radi
      * @since 2.17
      */
     public static boolean looksLikeValidNumber(final String s) {
-        return (s != null) && PATTERN_FLOAT.matcher(s).matches();
+        // the PATTERN_FLOAT match returns true for empty strings, so we need to check for digits separately
```
cowtowncoder, line 581: No need: caller will fail on some things that look like number but aren't. Will remove added check.

cowtowncoder:
(I thought I had added a comment here).

No. this cannot and should not be removed: while the location of method may be misplaced (it is really only needed by `jackson-databind`, not core streaming itself), it serves necessary function as can be seen from `NumberDeserializers`.
It has nothing to do wrt general validation of JSON Strings but needed for heuristics on when to apply maximum number length checks to avoid possibly DoS for decoding "too long" floating-point numbers.

Thank you @pjfanning for PR, will merge to get in 2.17.1

cowtowncoder (approved):

PR #1252, opened 2024-03-27
`ThreadLocalBufferManager` replace synchronized with `ReentrantLock`

pjfanning:
use ReentrantLock instead


cowtowncoder:
This is legacy thing, to be removed (already removed) from 3.0 so not sure if it's worth it. But I guess no strong feeling eithe rway.

cowtowncoder (approved):
Fine, let's just do it.

Issue #1260, opened 2024-04-15
Figure out why `LockFreePool` appears to cause unintended object retention (~= memory leak)


cowtowncoder:
Based on report problem like:

* https://github.com/FasterXML/jackson-module-scala/issues/672
* https://github.com/trinodb/trino/issues/21356

it looks as if the new recycler pool default (see #1117) -- Lock-free Pool -- results in (at least in some cases) unbounded increasing memory retention over time. From symptoms it would appear as if actual releases to pool happened more often than succesful reuse from the pool, as otherwise 2 operations should generally balance out over time (resulting in maximum pool size relative to maximum concurrent use).

We need to figure out what could cause such behavior, or if not, what is the root cause of observed problems.

In the meantime we also will need to do #1256 to allow use of 2.17(.1) without this issue.

cowtowncoder:
/cc @mariofusco @pjfanning -- I hope to learn if there is some specific usage scenario and create multi-threaded reproduction. I am suspecting that somehow reuse from the pool is failing, but perhaps not for every call. `jackson-benchmarks`, for example, does not exhibit this problem: but it also does not by default use heavy concurrency and tests do not run long enough to necessarily accumulate humongous buffers.

Also: it is interesting that the one concern I did have wrt unbounded pools ended up happening, it seems.
So it feels like my insistence on ideally having maximum size to avoid this problem (... although, doing so might hide an issue to) seems warranted. For whatever that is worth.

mariofusco:
HI @cowtowncoder I would like to investigate this memory issue, but at the moment I have no clue on how to reproduce it on my side. I also asked [here](https://github.com/trinodb/trino/issues/21356#issuecomment-2055760668) and I'm waiting for an answer. I'm reviewing its implementation again, and in all honesty it seems quite straightforward, so at the moment I don't know what to do about it. About making the LockFree pool bounded, this is of course feasible, but I'm against for a few reasons:

1. In general I don't like blind fixes
2. In this specific case making the pool bounded won't help as to investigate the problem, but only to put it under the carpet
3. Making that pool bounded won't be free however, both adding some further complexity to it and introducing a performance cost
4. We already have a bounded pool implementation, with performance very similar to the LockFree, so if we want a bounded pool as default implementation, let's just use it

mariofusco:
[...]

PR #1265, opened 2024-04-17
Part of #1260: write a manually run concurrency test to tease out problem with pool retention

cowtowncoder:
So, `RecyclerPoolTest` under `src/test/java/perf/` runs through 3 pool implementations with N threads (like 100) over some amount of time (30 or 60 seconds), and checks pooled entry count (pool.pooledCount()); occasionally printing size.
Hope is that this will show abnormal behavior seen in production environments.

### DIFF
--- a/src/test/java/perf/RecyclerPoolTest.java
+++ b/src/test/java/perf/RecyclerPoolTest.java
@@ -1,0 +1,159 @@
+package perf;
+
+import java.io.*;
+import java.nio.charset.StandardCharsets;
+import java.util.Random;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+
+import com.fasterxml.jackson.core.JsonFactory;
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.util.BufferRecycler;
+import com.fasterxml.jackson.core.util.JsonRecyclerPools;
+import com.fasterxml.jackson.core.util.RecyclerPool;
+
+/**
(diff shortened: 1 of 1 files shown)


cowtowncoder:
Running locally, with the latest code -- with 100 threads, 60 seconds for others, 120 seconds for lock-free, I get results:

```
Tests complete! Results:

 * Completed test of 'LockFreePool': max size seen = 22440
 * Completed test of 'ConcurrentDequePool': max size seen = 29
 * Completed test of 'BoundedPool': max size seen = 34
```

Interestingly enough, `LockFreePool` seems to hit its maximum within 1 second (!) of start, probably that is when highest contention is when things are warming up. After that time things always seem to settle.

But the other two implementations appear much better behaved. So I am starting to think that maybe default should actually be changed to `ConcurrentDequePool` as it has low minimum footprint (important for "single use" usage) while staying within expected boundaries.

mariofusco:
@cowtowncoder nice reproducer! I also gave a try and came to your similar conclusions, even though I made a few attempts and never seen numbers so high for the `LockFreePool`, but I guess this also depends on the underlying hardware. I also tried to modified that pool replacing the inner 3 loops before giving up on contention and creating a new resource on the acquire or throwing it away forever on release with an infinite loop and this indeed lowered the max size reached by the pool, but at the cost of a slightly lower throughput.

All this considered I agree on your suggestion of deprecating and ultimately removing the `LockFreePool` and using the `ConcurrentDequePool` as the default one for next releases.

pjfanning:
All of these new pools have much worse performance than the thread pool one. I feel stringly that it is much too early to make the default to use the really slow pools.

cowtowncoder:
At this point there is just one more thing I would like to confirm wrt performance before rollback for 2.18+ (for 2.17.1 it;s done): is the problem

1. Simply because the overhead of thread-contention for access in massively parallel (test) cases OR
2. do the pools fail to recycle buffers thereby being slower than thread-local - one

since (1) is something we perhaps cannot solve.

One way to figure out if (2) might occur would be to profile memory allocations: if recycling works as it should, new pool recycler would not do any more allocations than threadlocal-based pool.

Issue #1274, opened 2024-04-25
`NUL`-corrupted keys, values on JSON serialization


jaredstehler:
First posted in discussions: https://github.com/FasterXML/jackson/discussions/215

Jackson version: `2.12.6`

We have observed sporadic instances of serialized JSON messages containing corrupted keys and/or values (correct length, but some amount of string suffix is replaced with `NUL` (`0`) bytes).

From our metrics, it seemed to recur every 2-4 days on average. We have swapped in a fork of `jackson-core` with fields marked as `volatile` in `SerializedString` (https://github.com/HubSpot/jackson-core/commit/4d3bd2db409468e275ca4c7c471b523c5d1c18d9) and have not seen the issue in > 10 days now.

Even though we saw this once or twice a while back on JDK 17, it seems to have been exacerbated with JDK 21. We are also running on ARM64 instances, which could have some behavior contributing here.

I have not been able to reproduce this in isolation, unfortunately. Happy to help provide any additional information. We are working on an upgrade to jackson `2.18` but I see those fields are unchanged in that branch as well.

pjfanning:
Can you provide us with your code? Ideally as a PR but paste it here if a PR is not something you have time for.

PS We are unlikely to release a new v2.12 fix. The fix is likely to only go in more recent release lines - 2.17 and possibly 2.16, etc.

jaredstehler:
Oh I wasn't planning on pushing for a 2.12 release; we are definitely moving to a higher version. I only called it out as a clarifying point of context.

pjfanning:
> Which code are you looking for? The actual invocations to the ObjectMapper? Types being serialized?

I was asking for a PR with the changes you want to SerializedString.

jaredstehler:
ah, sorry. opened PR https://github.com/FasterXML/jackson-core/pull/1276

cowtowncoder:
Fix merged via #1297.

PR #1276, opened 2024-04-25
add volatile flag to SerializedString byte[] fields

jaredstehler:
Fixes https://github.com/FasterXML/jackson-core/issues/1274

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/io/SerializedString.java
+++ b/src/main/java/com/fasterxml/jackson/core/io/SerializedString.java
@@ -42,11 +42,11 @@ public class SerializedString
      *   and hopefully beyond.
      */
 
-    protected /*volatile*/ byte[] _quotedUTF8Ref;
+    protected volatile byte[] _quotedUTF8Ref;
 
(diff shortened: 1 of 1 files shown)


cowtowncoder:
Hi @jaredstehler! Have you been able to observe the different wrt old code, one with the fix?
I think you mentioned that the initial results seemed encouraging, just wanted to follow up.

If this fix needs to go in (which it sounds it does), it probably has negative performance effect so I want to try to make sure it does resolve the problem observed.

jaredstehler:
I ran with this fork in our environment for an extended test, and did not observe the issue for > 14 days. Subsequently, when rolling out the fork more fully after the test, we observed the issue during the time when a service had switched back to mainline 2.12.6.

I will continue to monitor and report back if we observe the issue for any service running our fork, which would rule out the fix, but our findings do seem to line up with this being the culprit.

cowtowncoder:
@jaredstehler Sounds good. I will off merging this for a bit now, given that there is plenty of time to get this in 2.18, and I don't intend to backport this in 2.17 (but even if I did, 2.17.1 was just released so 2.17.2 will probably take 2-4 months to release anyway).

PR #1278, opened 2024-04-26
Fix "iff" typo in Javadoc and comments, remove some redundant expressions

Philzen:
Analogous to https://github.com/FasterXML/jackson-databind/pull/4502

Also took the liberty to remove some code lines that came up in my IntelliJ highlighting (there's a couple more regarding redundant casts and using protected modifier in a final class – however i only applied the most obvious and safe ones that simplify the class).


pjfanning:
@Philzen could you hold your horses? Let's discuss the other PRs first.

cowtowncoder:
It's not a typo:

https://en.wikipedia.org/wiki/If_and_only_if

so only to be removed from cases where plain "if" is better -- but I try to add it only where "if and only if" is indeed the accurate usage.

Philzen:
Closed in favor of https://github.com/FasterXML/jackson-core/pull/1282

PR #1279, opened 2024-04-26
Remove duplicated condition

Philzen:
Haven't checked the bytecode if this would be optimized away by the compiler or not – so cannot say whether this would also pose a nice micro-optimization.

However it simplifies the condition and reduces the code base.

Kindly advise if this is the right target branch for the PR.


cowtowncoder:
I don't plan for new releases of 2.15 so this should be rebased for 2.18 (current default).

cowtowncoder (approved):

Philzen:
> … so this should be rebased for 2.18 (current default).

Done.

cowtowncoder (approved):

cowtowncoder:
LGTM, will merge. Thank you @Philzen !

PR #1280, opened 2024-04-26
Remove unnecessary semicolon

Philzen:


Philzen:
As per [ed8732d](https://github.com/FasterXML/jackson-core/pull/1278/commits/ed8732d741809624a48e781aea44cca46c96d719), moved this commit into a separate PR.

cowtowncoder:
Ok, going forward these are not kind of PRs that are super useful, and come down to stylistic differences (I somehow like trailing semicolon for enums, but YMMV). Same for semicolon for empty statements.
Does not make any difference to compiler, same resulting bytecode.

Still. I know others like minimalism so yeah, let's remove redundancy.

But like other PRs, let's re-base to 2.18.

Philzen:
> But like other PRs, let's re-base to 2.18.

Done.

cowtowncoder (approved):

PR #1281, opened 2024-04-26
Remove redundant suppression

Philzen:


Philzen:
As per https://github.com/FasterXML/jackson-annotations/pull/250#issuecomment-2080120796, moved this commit into a separate PR.

src/main/java/com/fasterxml/jackson/core/base/ParserBase.java
```diff
@@ -439,7 +439,6 @@ public boolean hasTextCharacters() {
         return false;
     }
 
-    @SuppressWarnings("resource")
```
cowtowncoder, line 442: I don't see this as redundant on Eclipse.

src/main/java/com/fasterxml/jackson/core/Base64Variant.java
```diff
@@ -669,7 +669,6 @@ public String encode(byte[] input, boolean addQuotes, String linefeed)
      *
      * @throws IllegalArgumentException if input is not valid base64 encoded data
      */
-    @SuppressWarnings("resource")
```
cowtowncoder, line 672: Does not exist in 2.18 any longer (converted to try () block).

cowtowncoder:
No longer relevant; one warning removed by 2.18, the other seems legit.

PR #1285, opened 2024-04-30
Fix #1277: add JDK 22 optimizations back to FDP

### LINKED ISSUE #1277: Add back Java 22 optimisation in fastdoubleparser
The pom.xml removes the Java 22 classes for fastdoubleparser. Maven Shade plugin settings. Will also need to add a relocation setting.

It should be safe to add them back now. Build tools have releases that support Java 22 format classes.

cowtowncoder:


cowtowncoder:
@pjfanning Is this what is needed or #1277 ?

cowtowncoder:
Ok, there seems to be an issue; maybe due to Shade plugin not working with JDK 22 bytecode.
There's a newer version; need to update version via `oss-parent` tomorrow.

cowtowncoder:
Ok so either Shade plugin, or latest Maven, fixed the problem: CI passes now.

pjfanning (approved):
Lgtm

PR #1296, opened 2024-06-08
Count JSON tokens

pjfanning:
Idea is to be able to apply a StreamReadConstraint that limits total JSON tokens that you can have in an input doc.

See https://github.com/FasterXML/jackson/discussions/231#discussioncomment-9603279

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/JsonParser.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonParser.java
@@ -782,6 +782,18 @@ public abstract class JsonParser
     }
 
     /**
+     * Get an approximate count of the number of tokens that have been read.
+     * This count is likely to be only updated if {@link StreamReadConstraints.Builder#maxTokenCount(long)}
+     * has been used to set a limit on the number of tokens that can be read.
+     *
+     * @return the number of tokens that have been read (-1 if the count is not available)
+     * @since 2.18
+     */
+    public long currentTokenCount() {
+        return -1L;
+    }
+
+    /**
      * Deprecated alias for {@link #currentLocation()} (removed from Jackson 3.0).
      *
      * @return Location of the last processed input unit (byte or character)
@@ -862,18 +874,6 @@ public abstract class JsonParser
     @Deprecated // since 2.17
     public void setCurrentValue(Object v) {
         assignCurrentValue(v);
-    }
-
-    /**
-     * Get an approximate count of the number of tokens that have been read.
-     * This count is likely to be only updated if {@link StreamReadConstraints.Builder#maxTokenCount(long)}
-     * has been used to set a limit on the number of tokens that can be read.
-     *
-     * @return the number of tokens that have been read (-1 if the count is not available)
-     * @since 2.18
-     */
-    public long getTokenCount() {
-        return -1L;
     }
 
     /*
--- a/src/main/java/com/fasterxml/jackson/core/JsonParser.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonParser.java
@@ -864,6 +864,18 @@ public abstract class JsonParser
         assignCurrentValue(v);
     }
 
+    /**
+     * Get an approximate count of the number of tokens that have been read.
+     * This count is likely to be only updated if {@link StreamReadConstraints.Builder#maxTokenCount(long)}
+     * has been used to set a limit on the number of tokens that can be read.
+     *
+     * @return the number of tokens that have been read (-1 if the count is not available)
+     * @since 2.18
+     */
+    public long getTokenCount() {
+        return -1L;
+    }
+
     /*
     /**********************************************************
     /* Buffer handling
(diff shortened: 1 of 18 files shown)


src/main/java/com/fasterxml/jackson/core/json/async/NonBlockingJsonParserBase.java
```diff
@@ -826,21 +826,20 @@ protected final JsonToken _eofAsNextToken() throws IOException {
             _handleEOF();
         }
         close();
-        return (_currToken = null);
+        return _updateToken(null);
```
cowtowncoder, line 829: _updateTokenToNull ?

src/main/java/com/fasterxml/jackson/core/json/async/NonBlockingUtf8JsonParserBase.java
```diff
@@ -467,7 +467,7 @@ private final JsonToken _finishBOM(int bytesHandled) throws IOException
         }
         _pending32 = bytesHandled;
         _minorState = MINOR_ROOT_BOM;
-        return (_currToken = JsonToken.NOT_AVAILABLE);
+        return _updateToken(JsonToken.NOT_AVAILABLE);
```
cowtowncoder, line 470: This would work incorrectly since it can happen multiple times in a row; should only update once token type is determined.
So maybe third method `_updateTokenNA()` or such, similar to `_updateTokenToNull()`?

src/main/java/com/fasterxml/jackson/core/json/async/NonBlockingJsonParserBase.java
pjfanning: done

src/main/java/com/fasterxml/jackson/core/base/ParserMinimalBase.java
```diff
@@ -782,6 +795,22 @@ protected final void _wrapError(String msg, Throwable t) throws JsonParseExcepti
         throw _constructReadException(msg, t);
     }
 
+    protected final JsonToken _updateToken(final JsonToken token) throws StreamConstraintsException {
+        _currToken = token;
+        if (streamReadConstraints().hasMaxTokenCount() && token != JsonToken.NOT_AVAILABLE) {
```
pjfanning, line 800: @cowtowncoder I'm doing the hasMaxTokenCount check here to try to minimise the impact of this code when hasMaxTokenCount=false. Testing for hasMaxTokenCount in the validateTokenCount will mean that we do a fair amount of unnecessary work if hasMaxTokenCount=false.

This change needs more testing and microbenchmarking but this is what I think is the best initial strategy.

cowtowncoder: As a further optimization, I think we should just have

    protected final boolean _hasTokenCountLimit;

initialized on parser being constructed. But need to see how to do that given that for now `StreamReadConstraints` are not properly passed to `ParserMinimalBase`. I can probably change code to pass it tho. Will create an issue and try to get that done.

I think we can also remove `JsonToken.NOT_AVAILABLE` check: there's the other method to call and only non-blocking/Async parsers (JSON, Smile) ever set that token, nothing else.

cowtowncoder:
Created #1310 as the issue covering this PR.

I think implementation makes sense, as-is. So that's fine. But merging to master/3.0 may be challenging, so what I think I will do is create another PR for a subset of this one, to handle parts that are likely to cause most problems. That can be merged in 2.18, then master, and then sync pr.

cowtowncoder:
@pjfanning Ok I think this is good: I merged update-token parts via separate PR into 2.18, this PR & master and remaining diffs are bit smaller and hopefully merge with few conflicts.
I assume this could change from draft to regular one now?
While perf tests might be nice, I feel confident that when disabled (no limit defined), overhead should be negligible.

cowtowncoder:
@pjfanning Is it ok to merge this? (only asking since it's still a Draft)

pjfanning:
@cowtowncoder I've marked this as ready for review

src/main/java/com/fasterxml/jackson/core/JsonParser.java
```diff
@@ -864,6 +864,18 @@ public void setCurrentValue(Object v) {
...
 
+    /**
+     * Get an approximate count of the number of tokens that have been read.
+     * This count is likely to be only updated if {@link StreamReadConstraints.Builder#maxTokenCount(long)}
+     * has been used to set a limit on the number of tokens that can be read.
+     *
+     * @return the number of tokens that have been read (-1 if the count is not available)
+     * @since 2.18
+     */
+    public long getTokenCount() {
```
cowtowncoder, line 875: Actually, I think that I will change naming to `currentTokenCount()`, to follow new(er) naming convention we have going for, wrt dynamic values like this.

cowtowncoder (approved):
LGTM

PR #1299, opened 2024-06-10
fix ignored maxDocLen in deprecated StreamReadConstraints constructor

pjfanning:
really just tidy up - this constructor is not likely used by anyone


cowtowncoder (approved):

cowtowncoder:
Good catch regardless, even if hopefully no one is bitter by this bug.

PR #1314, opened 2024-06-26
add _nullSafeUpdateToken

pjfanning:
I found use cases in jackson-databind and jackson-dataformats-text where the new token value can be null.


cowtowncoder (approved):
Makes sense; I was wondering about this when looking at other changes. Thanks!

PR #1331, opened 2024-09-11
uptake FastDoubleParser v1.0.1

pjfanning:
see https://github.com/FasterXML/jackson-databind/issues/4694#issuecomment-2344827965

Pick up fix for https://github.com/wrandelshofer/FastDoubleParser/issues/80.

Thanks to @wrandelshofer for his continued support and quick response times. 

https://github.com/wrandelshofer/FastDoubleParser/releases/tag/v1.0.1


cowtowncoder:
As per my note on https://github.com/FasterXML/jackson-databind/issues/4694 I think I'd like to see if there's any chance of getting FDP fix and new version before adding work-arounds.

In the meantime what would be valuable would be (for now) failing test case first, to check fix, guard against regression.

src/test/java/com/fasterxml/jackson/core/io/BigDecimalParserTest.java
```diff
@@ -51,6 +52,18 @@ void longValidStringFastParse() {
         assertEquals(EXP, BigDecimalParser.parseWithFastParser(num.toCharArray(), 0, num.length()));
     }
 
+    @Test
```
cowtowncoder, line 55: This is useful, although would probably be good to also test via `JsonParser` backends.

src/test/java/com/fasterxml/jackson/core/read/NonStandardNumberParsingTest.java
```diff
@@ -243,6 +243,20 @@ void leadingDotInNegativeDecimalAllowedReader() throws Exception {
         _testLeadingDotInNegativeDecimalAllowed(jsonFactory(), MODE_READER_THROTTLED);
     }
 
+    // https://github.com/FasterXML/jackson-databind/issues/4694
```
cowtowncoder, line 246: Can remove, added concurrently as `NumberParsing4694Test.java` under `src/test/java/.../failing`, to be merged in `NumberParsingTest` once fixed.

cowtowncoder (approved):
So: makes sense, but will wait for a bit before merging to see if there's any chance to get fix via FDP before 2.18.0 release. Or not.

src/test/java/com/fasterxml/jackson/core/read/NonStandardNumberParsingTest.java
```diff
@@ -21,6 +21,9 @@ class NonStandardNumberParsingTest
             .enable(JsonReadFeature.ALLOW_LEADING_DECIMAL_POINT_FOR_NUMBERS)
             .enable(JsonReadFeature.ALLOW_TRAILING_DECIMAL_POINT_FOR_NUMBERS)
             .build();
+    private final String ISSUE_4694_VALUE =
```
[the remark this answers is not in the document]
pjfanning: moved tests

cowtowncoder:
Thank you for clean up @pjfanning ! Will make sure we'll either add this work-around, or upgrade if FDP fix available, before 2.18.0 final.

src/main/java/com/fasterxml/jackson/core/io/BigDecimalParser.java
```diff
@@ -40,7 +40,17 @@ private BigDecimalParser() {}
      * @throws NumberFormatException
      */
     public static BigDecimal parse(String valueStr) {
-        return parse(valueStr.toCharArray());
+        try {
+            if (valueStr.length() < 500) {
+                return new BigDecimal(valueStr);
+            }
+            return JavaBigDecimalParser.parseBigDecimal(valueStr);
```
pjfanning, line 47: effectively, the same code as the parse(char[]) method but slight optimisation is that we avoid converting to a char array when we call the  JavaBigDecimalParser method

cowtowncoder (approved):

pjfanning:
@cowtowncoder The bug is an edge case, is this change worth backporting to 2.17? I'm on the fence.

cowtowncoder:
Merged, backporting in 2.17(.3) -- just the library upgrade tho, not other changes.

PR #1333, opened 2024-09-11
exception message generation uses too much memory

pjfanning:
I've looking at issues in BigDecimalParser and noticed that the exception message code that is meant to protect us from consuming too much memory is not efficient enough.

This PR adds methods to handle char arrays and avoids creating Strings for the potentially large arrays.


cowtowncoder (approved):
Good catch & thank you for doing smaller PRs to isolate work-arounds from long-term improvements.

PR #1332, opened 2024-09-11
Create BigDecimalParser4694Test.java

pjfanning:
see #1331

PR #1334, opened 2024-09-11
Create NonStandardNumberParsing4694Test.java

pjfanning:
another test for #1331 - I think this and #1332 are both useful in their own rights


cowtowncoder (approved):
Looks useful on its own right, agreed. Will probably rename as it's not testing non-standard numbers, after merging.

PR #1335, opened 2024-09-12
Write 4-byte characters (surrogate pairs) instead of escapes

rnetuka:
Issue: https://github.com/FasterXML/jackson-core/issues/223

A fix for issue #223. Some languages (Japanese, Chinese, ...) use 4-byte characters and it's annoying to have them marshaled as escapes. 

I've tried to fix this with minimum intrusions. I've only added this feature/fix to __writeStringSegment2_ method(s) and left the rest without it. If you like the changes for other methods calling __outputMultiByteChar_, just let me know.


pjfanning:
* this probably imposes a performance hit on all users
* some users may even prefer the existing behaviour
* I think this should be only applied if a new JsonGenerator.Feature is added and defaulted to false.

https://www.javadoc.io/static/com.fasterxml.jackson.core/jackson-core/2.17.2/com/fasterxml/jackson/core/JsonGenerator.Feature.html

pjfanning:
The change that you have made to PackageVersion.java.in breaks the build - please revert it.

rnetuka:
@pjfanning Thanks for your review. I've added the feature. Not sure if _StreamWriteFeature_ is the right place...

src/main/java/com/fasterxml/jackson/core/json/UTF8JsonGenerator.java
```diff
@@ -1510,7 +1515,14 @@ private final void _writeStringSegment2(final char[] cbuf, int offset, final int
                 outputBuffer[outputPtr++] = (byte) (0xc0 | (ch >> 6));
                 outputBuffer[outputPtr++] = (byte) (0x80 | (ch & 0x3f));
             } else {
-                outputPtr = _outputMultiByteChar(ch, outputPtr);
+                // multibyte character
+                if (Feature.WRITE_UTF8_SURROGATES.enabledIn(_features) && isSurrogatePair((char) ch) && offset < end) {
+                    char highSurrogate = (char) ch;
```
pjfanning, line 1520: it might be possible to create a class variable to hold the result of Feature.WRITE_UTF8_SURROGATES.enabledIn(_features) so that we don't need to recalculate this for every char.

pjfanning:
@rnetuka @cowtowncoder It looks like the Windows CI build cannot handle UTF-8 chars in source files.

rnetuka:
@pjfanning _WriterBasedJsonGenerator_ works fine and writes the surrogate pairs automatically.

As for _UTF-8 chars in source files_, do you mean the ones in the new test? If so, I can just replace them with escapes.

pjfanning:
> @pjfanning _WriterBasedJsonGenerator_ works fine and writes the surrogate pairs automatically.
> 
> As for _UTF-8 chars in source files_, do you mean the ones in the new test? If so, I can just replace them with escapes.

The Windows build doesn't like any of your non-ASCII chars. Could you have escape all the non-ASCII chars?

src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
```diff
@@ -202,6 +202,8 @@ public enum Feature {
          */
         WRITE_BIGDECIMAL_AS_PLAIN(false),
 
+        WRITE_UTF8_SURROGATES(false),
```
cowtowncoder, line 205: Needs Javadoc (like other options)

src/test/java/com/fasterxml/jackson/core/json/StringGenerationTest.java
```diff
@@ -86,6 +87,17 @@ void longerRandomMultiChunk() throws Exception
         }
     }
 
+    @Test
+    public void testWritingSurrogatePairs() throws IOException {
+        ByteArrayOutputStream stream = new ByteArrayOutputStream();
+        JsonGenerator generator = FACTORY.createGenerator(stream, JsonEncoding.UTF8).enable(Feature.WRITE_UTF8_SURROGATES);
+        String string = "システム\uD867\uDE3D"; // システム𩸽
```
cowtowncoder, line 94: Yes, def need to use Unicode escapes for Katakana characters.

src/test/java/com/fasterxml/jackson/core/json/StringGenerationTest.java
```diff
@@ -86,6 +87,17 @@ void longerRandomMultiChunk() throws Exception
         }
     }
 
+    @Test
+    public void testWritingSurrogatePairs() throws IOException {
```
cowtowncoder, line 91: Note: there is already

src/test/java/com/fasterxml/jackson/failing/Surrogate223Test.java

that hopefully now passes; if so, should be moved to another location from under `.../failing`

cowtowncoder:
@rnetuka Forgot to say: big thank you for contributing this!!! I'll try to make sure this gets in 2.18.0 release; as I said this looks pretty good already.

rnetuka:
@cowtowncoder Based on you review, I've updated the PR with following changes:
- moved the feature into the group _// Misc other_
- added javadoc for the feature
- reused _Surrogate223Test_ (which is now green) instead of creating new test method

I've renamed the new feature to _COMBINE_UNICODE_SURROGATES_ which I believe will better describe its purpose. I've also put the feature only to _JsonGenerator_ (previously I put it also in _StreamWriteFeature_, but that would make it kind of duplicate). 

Thank you for you review. If you have any more comments, just let me know.

src/test/java/com/fasterxml/jackson/core/json/StringGenerationTest.java
rnetuka: Test deleted. Surrogate223Test cover the new feature instead.

src/test/java/com/fasterxml/jackson/core/json/StringGenerationTest.java
rnetuka: Confirmed. Surrogate223Test is green now.

src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
rnetuka: Javadoc added

cowtowncoder:
@rnetuka Just realized I don't have a CLA from you. However, I do have corporate-CLA from Red Hat -- and it looks like you are employed by Red Hat. If this is the case, I think you are covered and we are good.
Please let me know if this is the case (will remove "cla-needed" label).

Also: I see failed unit tests -- is this PR still work-in-progress?
(I had to approve CI run since this is your first contribution, hopefully it will now re-run on all pushes)

rnetuka:
@cowtowncoder Yes, I'm Red Hat employee.

As for the test, I've noticed _Surrogate223Test#surrogatesCharBacked_ is written for an option _to escape the surrogate pairs_ for _StringWriter_/_WriterBasedJsonGenerator_. _WriterBasedJsonGenerator_ always combines the surrogate pairs together, so I've removed the latter part of the test. Please, take a look and confirm this is ok for you.

To sum it up, if merged at this point:
_UTF8JsonGenerator_ escapes 4-byte characters by default
_UTF8JsonGenerator_ can be configured to combine the bytes instead
_WriterBasedJsonGenerator_ always combines the bytes together

src/main/java/com/fasterxml/jackson/core/json/UTF8JsonGenerator.java
```diff
@@ -2133,6 +2156,13 @@ protected final void _outputSurrogates(int surr1, int surr2) throws IOException
         bbuf[_outputTail++] = (byte) (0x80 | (c & 0x3f));
     }
 
+    private int _outputSurrogatePair(char highSurrogate, char lowSurrogate, int outputPtr) {
+        String s = String.valueOf(highSurrogate) + lowSurrogate;
+        byte[] bytes = s.getBytes(StandardCharsets.UTF_8);
+        System.arraycopy(bytes, 0, _outputBuffer, outputPtr, bytes.length);
```
cowtowncoder, line 2162: Yikes. This is rather wasteful way of doing this... Let me think we really shouldn't have do too this much allocation of things.

rnetuka: Would
```
private int _outputSurrogatePair(char highSurrogate, char lowSurrogate, int outputPtr) {
        int unicode = 0x10000;
        unicode += (highSurrogate & 0x03FF) << 10;
        unicode += (lowSurrogate & 0x03FF);

        byte[] utf8 = new byte[4];
        utf8[0] = (byte) (0b11110000 + ((unicode & 0b00000000_00011100_00000000_00000000) >> 18));
        utf8[1] = (byte) (0b10000000 + ((unicode & 0b00000000_00000011_11110000_00000000) >> 12));
        utf8[2] = (byte) (0b10000000 + ((unicode & 0b00000000_00000000_00001111_11000000) >> 6));
        utf8[3] = (byte) (0b10000000 + (unicode & 0b00000000_00000000_00000000_00111111));

        System.arraycopy(utf8, 0, _outputBuffer, outputPtr, utf8.length);
        return outputPtr + utf8.length;
}
```
be better?

src/main/java/com/fasterxml/jackson/core/json/UTF8JsonGenerator.java
```diff
@@ -1517,14 +1511,17 @@ private final void _writeStringSegment2(final char[] cbuf, int offset, final int
...
                 outputBuffer[outputPtr++] = (byte) (0x80 | (ch & 0x3f));
             } else {
-                // multibyte character
-                if (combineSurrogates && isSurrogatePair((char) ch) && offset < end) {
-                    char highSurrogate = (char) ch;
-                    char lowSurrogate = cbuf[offset++];
-                    outputPtr = _outputSurrogatePair(highSurrogate, lowSurrogate, outputPtr);
-                } else {
-                    outputPtr = _outputMultiByteChar(ch, outputPtr);
+                // 3- or 4-byte character
```
rnetuka, line 1514: Only a 4-byte characters (i.e. U+010000 - U+10FFFF). Characters that are encoded as UTF-8 3-byte falls to _outputMultiByteChar method

cowtowncoder (approved):

src/main/java/com/fasterxml/jackson/core/json/UTF8JsonGenerator.java
cowtowncoder: Copied with minor modifications; also realized unit test was not verifying actual round-trip encode/decode, added. Now catches some of problems I tried.
(ideally we'd have more tests for surrogate pairs but will have to do for now).

Will merge.

cowtowncoder:
Thank you for contributing this, @rnetuka ! Will merge in 2.18 for inclusion in 2.18.0 release.

PR #1337, opened 2024-09-21
add tests for ints with 19 chars

pjfanning:
* based on https://github.com/plokhotnyuk/jsoniter-scala/issues/1195
* with int processing, it is common to count the chars and parse as a Long if the chars are <= 19 and BigInteger if above
* this approach can be risky because a few values may be at a little too big to parse as a long and you could get parse issues

### DIFF
--- a/src/test/java/com/fasterxml/jackson/core/read/NumberParsingTest.java
+++ b/src/test/java/com/fasterxml/jackson/core/read/NumberParsingTest.java
@@ -454,11 +454,31 @@ class NumberParsingTest
         BigInteger biggie = new BigInteger(NUMBER_STR);
 
         for (int mode : ALL_MODES) {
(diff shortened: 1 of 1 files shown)


cowtowncoder (approved):

src/test/java/com/fasterxml/jackson/core/read/NumberParsingTest.java
```diff
@@ -463,6 +463,26 @@ void bigNumbers() throws Exception
         }
     }
 
+    @Test
+    void intsWith19Chars() throws Exception
+    {
+        final String[] values = new String[] {
+            "9223372036854775808", "9999999999999999999"
```
cowtowncoder, line 470: Makes sense, could probably have based first on `Long.MAX_VALUE`? But will merge as-is.

cowtowncoder (approved):

cowtowncoder:
Thanks!

PR #1338, opened 2024-09-23
Fix "can not" -> "cannot"

TheJavaGuy:
Fixes various typos (mostly of `can not` -> `cannot` kind). Rationale is given in https://writingexplained.org/cannot-or-can-not-difference

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,7 +25,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java_version: ['8', '11', '17', '21', '22']
+        java_version: ['8', '11', '17', '21', '23']
         os: ['ubuntu-22.04']
         include:
           - java_version: '8'
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,7 +17,7 @@ permissions:
 
 jobs:
   release:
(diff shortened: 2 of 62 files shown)


cowtowncoder:
100% agree with @pjfanning: I'm fine with changes to Javadocs, but let's not change exception messages for 2.x.

So would it be possible to change this PR to only have Javadoc changes.
Additionally if it feels useful -- I am non-native english speaker so do not have a gut feelings -- could have another PR against `master` (Jackson 3.0) for Exception message changes: with 3.0 there are API changes anyway and we could make the change.

cowtowncoder:
Would be happy to apply just Javadoc changes...

cowtowncoder:
Applied Javadoc/comment-only changes to `2.18`. The rest can't go in 2.18, but I am thinking that if we want these, they really need to in `master` (for Jackson 3.0). This since while some changes are ok with minor (2.19) versions, I think these changes have a bit too high a chance to break some brittle code somewhere to warrant the risk.

This cannot be done with rebase unfortunately, but if anyone wants to take a stab at that I'd be happy to review and merge. Leaving this PR open as reminder for now.

cowtowncoder:
Ok, made in-comment changes in 2.18, and everything in 3.0/master. Closing this PR.

PR #1339, opened 2024-09-24
Fix #1340 by adding missing "provides" in `module-info.java`

### LINKED ISSUE #1340: Missing `JsonFactory` "provides" SPI with JPMS in `jackson-core` module
SPI does not work in Java projects using JPMS, since `module-info.java` does not contains any `provides` directive.

To reproduce, I created a Java project with:
- `module-info.java`
  ```java
  module foo {
      requires com.fasterxml.jackson.core;
      uses com.fasterxml.jackson.core.JsonFactory;
  }
  ```
- `main`
  ```java
  public static void main(String[] args) {
        Optional<JsonFactory> jf = ServiceLoader.load(JsonFactory.class).findFirst();
        System.out.println(jf.isPresent()); // --> false
  }
  ```

sdyura:
(Fixes #1340)

if we have a src/main/resources/META-INF/services/com.fasterxml.jackson.core.JsonFactory
this works in projects that have no module-info.java, but if a project does have module-info.java
then we need to also add the java modules way of provides

this fixes `ServiceLoader<JsonFactory> sl = ServiceLoader.load(JsonFactory.class);` not working (returning empty)
for projects that have a module-info.java

### DIFF
--- a/release-notes/VERSION-2.x
+++ b/release-notes/VERSION-2.x
@@ -45,6 +45,8 @@ a pure JSON library.
   of Tokens allowed per document#
 #1331: Update to FastDoubleParser v1.0.1 to fix `BigDecimal` decoding proble
  (fixed by @pjfanning)
+#1340: Missing `JsonFactory` "provides" SPI with JPMS in `jackson-core` module
+ (contributed by @sdyura)
 
 2.17.2 (05-Jul-2024)
 
--- a/src/moditect/module-info.java
+++ b/src/moditect/module-info.java
@@ -21,6 +21,7 @@ module com.fasterxml.jackson.core {
     // 03-Oct-2019, tatu: [core#567] Add self-use to avoid warnings
     uses com.fasterxml.jackson.core.ObjectCodec;
 
+    // 25-Sep-2024: [core#1340] Need to explicitly add even in base
     provides com.fasterxml.jackson.core.JsonFactory with
-            com.fasterxml.jackson.core.JsonFactory;
+        com.fasterxml.jackson.core.JsonFactory;
 }


pjfanning:
Would it be possible to provide something to test this with? Even a standalone project.

pjfanning:
Or even links to docs that describe this issue and/or fix generally.

sdyura:
to test you can run `ServiceLoader<JsonFactory> sl = ServiceLoader.load(JsonFactory.class);` in a project, it will work fine (return 1 item) if the project has no `module-info.java`, but as soon as you add `module-info.java` to the project, it will break (return 0 items)

here is a simple project that fails when run: https://github.com/user-attachments/files/17117454/arangodb-java-driver-demo.zip

sdyura:
it is standard java to Always put the ServiceLoader loaded class in BOTH the "META-INF/services" AND module-info.java, you can see in the other jackson module this is the case: 
 - https://github.com/FasterXML/jackson-databind/blob/2.18/src/main/resources/META-INF/services/com.fasterxml.jackson.core.ObjectCodec
 - https://github.com/FasterXML/jackson-databind/blob/2.18/src/moditect/module-info.java#L42

pjfanning:
This will likely be merged but can't users who need this today create their own jar that has a module-info.class in like described in #1341. The workaround jar might also need the META-INF/services/com.fasterxml.jackson.core.JsonFactory file.

The same users who think this is needed yesterday could also build their own jackson-core with this amendment in it.

cowtowncoder:
I can backport to, say, 2.16 and 2.17 but unlikely to release new patches for pre-2.17.

I assume this has been tested to work? (even if just manually)

cowtowncoder:
Backport in 2.16 and 2.17 in case new versions might be released. But the next version that has the fix will be 2.18.0.

rashtao:
Hi @cowtowncoder ,
I have manually checked that the changes work when building from `2.18`. Thanks!

PR #1341, opened 2024-09-25
SPI with JPMS

rashtao:
fixes https://github.com/FasterXML/jackson-core/issues/1340

---

Ideally this should be backported also to older branches.


pjfanning:
duplicate of https://github.com/FasterXML/jackson-core/pull/1339

PR #1347, opened 2024-10-05
add workflows for suggestions on PRs

timo-a:


pjfanning:
We don't typically start many PRs at master level. So I would prefer not to merge this.

Most PRs start on the latest dev 2.x branch - currently 2.19. And the changes are then forward fit.

pjfanning:
I'm not making any accusations but I am super wary of any PRs that modify GitHub workflows. I am a strong believer that you need a much higher level of trust when it comes to PRs like this.
* I hope @timo-a doesn't take this personally but they don't have much verifiable GitHub commit history
* Attacks on the build systems of FOSS projects is a real issue and a major concern. It's a terrible thing but we have people trying to damage the reputation of FOSS by submitting malicious build changes - and I repeat that I am not saying that this is one.
* One of the sneakiest ways to get a malicious change to the build system merged is for the attacker to claim that they are making the build system more secure - and again, I am not claiming that this is such a case.

pjfanning (changes requested):
-1 from me

.github/workflows/receive-pr.yml
```diff
@@ -0,0 +1,55 @@
...
+          repository: ${{github.event.pull_request.head.repo.full_name}}
+      - uses: actions/setup-java@v4
+        with:
+          java-version: '21'
+          distribution: 'temurin'
+          cache: 'maven'
+
+      # Capture the PR number
+      # https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#using-data-from-the-triggering-workflow
+      - name: Create pr_number.txt
```
pjfanning, line 35: this is not a good way to get the PR number - try first answer in https://stackoverflow.com/questions/59077079/how-to-get-pull-request-number-within-github-actions-workflow

pom.xml
```diff
@@ -298,4 +298,37 @@ tools.jackson.core.*;version=${project.version}
...
+  <profiles>
+    <profile>
+      <id>openrewrite</id>
+      <!-- `mvn -P openrewrite org.openrewrite.maven:rewrite-maven-plugin:run` -->
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.openrewrite.maven</groupId>
+            <artifactId>rewrite-maven-plugin</artifactId>
+            <version>5.41.0</version>
```
pjfanning, line 311: This can't be used on the 2.x branches because it would cause a huge diff

timo-a:
> We don't typically start many PRs at master level. So I would prefer not to merge this.
> 
> Most PRs start on the latest dev 2.x branch - currently 2.19. And the changes are then forward fit.

I have created a second PR against the proper target branch: https://github.com/FasterXML/jackson-core/pull/1348.

> [...] I am super wary of any PRs that modify GitHub workflows [...]

I am not taking this personally, you are correct to be wary of changes regarding the workflow. I would even go further and say that such changes should be measured against a higher standard than trust towards a contributor or their verifiable GitHub commit history. They should be not be approved if the maintainers do not understand their contents. (For security but also sustainability, I am making no commitment to provide fixes if needed) I am happy to answer questions regarding the workflow changes. I also understand if the review takes some time or ultimately results in rejection of the PR.

pjfanning:
We only need one PR.

pom.xml
timo-a: I can try to put the info from the profile directly into the command in the workflow, I'll just have to figure out how to specify the version number there.
The advantage of having the profile is in the pom is that you can reuse it to apply the changes locally. On the other hand if all the information is in the workflow then the workflow does not depend on the pom and you do not run the risk that changes to the pom inadvertently change the workflow behavior so this might be safer.

pjfanning: We have the imports ordered on master so we can afford to support this and have full reordering - it is just not a good idea to have this on the 2.x branches.

I still have my objections to using any build tooling that doesn't have multitudes of users and oodles of history to establish trust.

My preference is not merge any of this. Low cost to benefit ratio to me.

If we insist on merging something like this - I would prefer to look at established tools to do this.

I code mainly in Scala and use Scalafmt a lot. But if someone came to me with a custom project and asked me to replace Scalafmt, I would need a load of convincing.

timo-a: I see your point. Openrewrite is still new. It is the best tool I know for automatic refactoring and by extension it can also reformat code to a custom style. 
_(tangent: I believe that it is a time sink to define a [custom code style](https://github.com/FasterXML/jackson/blob/master/contribution/jackson-coding-style-guide.md#jackson-coding-style-guide) without also having tools to enforce it. Thus my main suggestion remains to [use an established code style](https://github.com/FasterXML/jackson-databind/pull/4448#discussion_r1544730223) which can be easily enforced. But @cowtowncoder is [not fan](https://github.com/FasterXML/jackson/pull/201#issuecomment-1848851178) of comprehensive auto-formatters and the google formatter in particular and them having fun maintaining the project is ultimately more important to its success than adherence to a code style.)_
I tried to see if I can use Openrewrite to enforce key aspects of the code style on PRs and thereby relieve the reviewer. I think such a workflow is really useful to fix common sonarqube findings on PRs in an industrial setting. However, I'm not sure if there are enough PRs on this project to justify adding that complexity to the project.

pjfanning: Again trying not to cast stones - but your PR includes call to a lib that is under your control. 
io.github.timo-a:rewrite-recipe-starter - this is a project with 0 stars, 0 forks

.github/workflows/receive-pr.yml
timo-a: I don't understand, both my workflow and the first stackoverflow answer use `github.event.number` to get the PR number. In order for the other workflow to get the PR number, this workflow then uploads the number as an artifact which the other workflow can download. I don't think this can be done with environment variables as used in the stackoverflow answer.

PR #1353, opened 2024-10-28
use fastdoubleparser 1.0.90

pjfanning:
an issue in ZIng JVM - and possibly other JVMs

https://github.com/wrandelshofer/FastDoubleParser/releases/tag/v1.0.90

It does seem like a bug in the JVM but seems best to use the workaround


cowtowncoder (approved):

cowtowncoder:
Thanks! Looks like 2.0.0 released too; probably more for Jackson 2.19?

PR #1354, opened 2024-10-31
Fix #1328: optimize `JsonPointer.head()`

### LINKED ISSUE #1328: Optimize handling of `JsonPointer.head()`
(note: this is a placeholder for reported but not verified performance concern)

Looks like implementation of `JsonPointer.head()` could be problematic for deeply nested pointer instances: and while this is unlikely to be of general security concern, it could be problematic for legitimate use. And as per comments in class, optimization already considered earlier... so let's see how easy it's to do.

cowtowncoder:


src/main/java/com/fasterxml/jackson/core/JsonPointer.java
```diff
@@ -814,31 +833,38 @@ private static void _appendEscape(StringBuilder sb, char c) {
...
+        int offset = 0;
 
-    protected JsonPointer _constructHead(int suffixLength, JsonPointer last)
-    {
-        if (this == last) {
-            return EMPTY;
+        while (current != last) {
+            JsonPointer nextSegment = new JsonPointer(current,
+                    fullString, offset);
+            offset += suffixLength;
```
pjfanning, line 856: @cowtowncoder This code adds the suffix Len to the prefix/offset Len. I think this is the cause of #1425.

cowtowncoder: Yes, exactly. Not sure why I chose that, makes no sense.

I am trying to remember if this is the one case where I did ask ChatGPT help with recursion or not (it was for something within this class). If so would be hilarious fail on my part.

At any rate, fixed via https://github.com/FasterXML/jackson-core/pull/1426 -- code is bit tricky to reason about but I think it works better now.

PR #1356, opened 2024-11-05
Make `JsonGenerator::writeTypePrefix` method to not make a `WRAPPER_ARRAY` when `typeIdDef.id == null`

Gems:
This PR addresses [GitHub issue #4772](https://github.com/FasterXML/jackson-databind/issues/4772).

**Summary**

In certain cases where `@JsonTypeInfo` is used with `JsonTypeInfo.Id.DEDUCTION`, `typeDef.id` may resolve to `null`. For example, in the [referenced issue](https://github.com/FasterXML/jackson-databind/issues/4772), this occurs with a class based on the `Collection` interface, resulting in `typeDef.id` being `null`.

**Change Details**

This update modifies the `sonGenerator::writeTypePrefix` method so that it doesn’t generate a wrapping array with a `"null"` key for classes with a `null` `typeDef.id`. The current behavior, which includes an unnecessary wrapping array labeled by `"null"`, can be misleading. By skipping this wrapper, we aim to make the serialization output more intuitive and aligned with user expectations.

**Rationale**

The previous behavior of adding a wrapping array with a "null" key was confusing and did not provide meaningful value to users. Removing it simplifies the output and makes the serialization format clearer, especially in cases involving classes based on generic interfaces like `Collection`.

**Impact**

This change will make the output format more predictable and consistent, especially for users employing `@JsonTypeInfo` with `JsonTypeInfo.Id.DEDUCTION`. Existing tests and validation tools should ensure that this change does not disrupt compatibility with existing deserialization setups.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
@@ -10,6 +10,7 @@ import java.math.BigInteger;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.Objects;
 
 import com.fasterxml.jackson.core.JsonParser.NumberType;
 import com.fasterxml.jackson.core.exc.StreamWriteException;
@@ -1978,12 +1979,13 @@ public abstract class JsonGenerator
         } else {
             // No native type id; write wrappers
             // Normally we only support String type ids (non-String reserved for native type ids)
-            String idStr = (id instanceof String) ? (String) id : String.valueOf(id);
+            String idStr = (id instanceof String) ? (String) id : Objects.toString(id, null);
             typeIdDef.wrapperWritten = true;
 
             Inclusion incl = typeIdDef.include;
             // first: can not output "as property" if value not Object; if so, must do "as array"
             if ((valueShape != JsonToken.START_OBJECT)
+                    && idStr != null 
                     && incl.requiresObjectContext()) {
                 typeIdDef.include = incl = WritableTypeId.Inclusion.WRAPPER_ARRAY;
             }
--- a/src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
@@ -1969,39 +1969,47 @@ public abstract class JsonGenerator
      */
     public WritableTypeId writeTypePrefix(WritableTypeId typeIdDef) throws IOException
     {
-        final boolean hasStartObjectWritten = canWriteTypeId()
-                // just rely on native type output method (sub-classes likely to override)
-                ? writeTypePrefixWithTypeId(typeIdDef)
-                // No native type id; write wrappers
-                : writeTypePrefixWrapper(typeIdDef);
+        // First: is this a native type id? If so, just use write method, be done
+        if (canWriteTypeId()) {
+            // just rely on native type output method (sub-classes likely to override)
+            return _writeNativeTypePrefix(typeIdDef);
+        }
+        
+        // No native type id; write wrappers
+        final boolean wasStartObjectWritten = _writeTypePrefixWrapper(typeIdDef);
 
         // and finally possible start marker for value itself:
         switch (typeIdDef.valueShape) {
         case START_OBJECT:
-            if (!hasStartObjectWritten)
+            if (!wasStartObjectWritten) {
                 writeStartObject(typeIdDef.forValue);
+            }
             break;
         case START_ARRAY:
             // should we now set the current object?
             writeStartArray();
             break;
+        default: // otherwise: no start marker
         }
 
         return typeIdDef;
     }
 
-    private boolean writeTypePrefixWithTypeId(WritableTypeId typeIdDef) throws IOException {
+    // @since 2.19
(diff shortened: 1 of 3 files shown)


src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
```diff
@@ -1978,12 +1979,13 @@ public WritableTypeId writeTypePrefix(WritableTypeId typeIdDef) throws IOExcepti
         } else {
             // No native type id; write wrappers
             // Normally we only support String type ids (non-String reserved for native type ids)
-            String idStr = (id instanceof String) ? (String) id : String.valueOf(id);
+            String idStr = (id instanceof String) ? (String) id : Objects.toString(id, null);
```
cowtowncoder, line 1982: Nice! TIL there's 2-arg variant. Will need to start using more widely.

Gems: I also learned it not that long time ago 😄

src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
```diff
@@ -1978,12 +1979,13 @@ public WritableTypeId writeTypePrefix(WritableTypeId typeIdDef) throws IOExcepti
...
             typeIdDef.wrapperWritten = true;
 
             Inclusion incl = typeIdDef.include;
             // first: can not output "as property" if value not Object; if so, must do "as array"
             if ((valueShape != JsonToken.START_OBJECT)
+                    && idStr != null
```
cowtowncoder, line 1988: I guess I don't quite understand logic here: change will prevent changing of output for some cases from non-Array to Array. But not prevent output or anything like that.

So... what use case being resolved by this change?

cowtowncoder:
One process thing: beyond code review, before merging I'd need CLA:

(unless already done earlier; one CLA per contributor is fine for all Jackson contributions).
The usual way is to print, fill & sign, scan/photo, email to `cla` at fasterxml dot com.

cowtowncoder:
Ah ok: so we don't want type id like:

```
["null", [ 1, 2, "foo" ]]
```

which I totally agree with.

However, actual `null` value seems more reasonable.

Alternatively another maybe even better way would be to consider `null` typeId to mean "no type wrapping" (neither ARRAY nor OBJECT). But the change as shown doesn't, I think, do that.

One problem with changes to Type Id handling is that it is feature `jackson-databind` only uses, but nothing in `jackson-core`. So unit test suite does not cover it well (although there is `src/test/java/com/fasterxml/jackson/core/write/WriteTypeIdTest.java`).

Oh: which reminds me: we'd definitely need a test added to `src/test/java/com/fasterxml/jackson/core/write/WriteTypeIdTest.java` to verify expected behavior.

src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
Gems: Seeing your comment in the PR comment I assume that you have a good grasp of the idea for change. I continue commenting there as find that context more relevant.

Gems:
I have to say that I don't have sufficient visibility into the design idea of the `writeTypePrefix` and thus it's a bit more stubbing in the dark rather than fully conscious decision-making having full picture in mind.

Though at least it's thought-provoking, and this was my goal.

I support the idea to "consider `null` `typeId` to mean "no type wrapping" (neither ARRAY nor OBJECT)". And you are right, the provided change doesn't cover it fully.

I'll do the drill related to CLA and improve the change in this PR some time soon.

Thanks for your feedback as usual.

Gems:
I did the changes and sent the signed agreement.

src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
```diff
@@ -1968,63 +1969,87 @@ public void writeTypeId(Object id) throws IOException {
...
+        case START_ARRAY:
             // should we now set the current object?
             writeStartArray();
+            break;
         }
+
         return typeIdDef;
     }
 
+    private boolean writeTypePrefixWithTypeId(WritableTypeId typeIdDef) throws IOException {
```
cowtowncoder, line 1993: Let's change to `protected` for overridability

src/main/java/com/fasterxml/jackson/core/JsonGenerator.java
```diff
@@ -1968,63 +1969,87 @@ public void writeTypeId(Object id) throws IOException {
...
+        writeTypeId(typeIdDef.id);
+
+        return false;
+    }
+
+    /**
+     * Writes a wrapper for the type id; if necessary.
+     * @return True if start of an object has been written, False otherwise.
+     */
+    private boolean writeTypePrefixWrapper(WritableTypeId typeIdDef) throws IOException {
```
cowtowncoder, line 2004: Let's change to `protected` for overridability

cowtowncoder:
@Gems looks good -- did just a quick first pass, need to take some more time, but I think this should work. Got CLA so we are good with that too.

Will try local build to hopefully ensure that format modules (esp. YAML, XML) are ok with this change (should be but it's potentially brittle area)

cowtowncoder:
Ok, some of my logic changes proved wrong wrt short-circuiting handling: cause issue with YAML. Will revert.

cowtowncoder (approved):
LGTM -- made some stylistic changes, but essentially included as-is.

Issue #1289, opened 2024-05-06
NegativeArraySizeException and ArrayIndexOutOfBoundsException in ByteQuadsCanonicalizer


richard-timpson:
## Versions:
Jackson Version: 2.12.7
Java Version: 11

## Context:

We operate a large installation of the open source [Spinnaker CD platform](https://github.com/spinnaker). Spinnaker uses a complicated json object to manage and facilitate context during a pipeline execution. At our scale this execution json object can be extremely large for certain use cases. There is an [endpoint](https://github.com/spinnaker/gate/blob/master/gate-web/src/main/groovy/com/netflix/spinnaker/gate/controllers/ApplicationController.groovy#L122) exposed to the Spinnaker UI that will return all of the executions for a given application. 

For this endpoint we frequently see a single payload over 2Gb, and sometimes up to 10Gb. 

## Problem:

Lately we have been seeing at least two different exceptions originate from the ByteQuadsCanonicalizer class. Here are stack traces for both. 

### NegativeArraySizeException

Error message: `"Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception [Request processing failed; nested exception is retrofit.RetrofitError: com.fasterxml.jackson.databind.JsonMappingException: -2147481843 (through reference chain: java.util.ArrayList[0])] with root cause"`
Stack Trace:
```
java.lang.NegativeArraySizeException: -2147483128
     java.base/java.util.Arrays.copyOf(Arrays.java:3793)
     com.fasterxml.jackson.core.sym.ByteQuadsCanonicalizer._appendLongName(ByteQuadsCanonicalizer.java:997)
     com.fasterxml.jackson.core.sym.ByteQuadsCanonicalizer.addName(ByteQuadsCanonicalizer.java:866)
     com.fasterxml.jackson.core.json.UTF8StreamJsonParser.addName(UTF8StreamJsonParser.java:2389)
     com.fasterxml.jackson.core.json.UTF8StreamJsonParser.findName(UTF8StreamJsonParser.java:2273)
     com.fasterxml.jackson.core.json.UTF8StreamJsonParser.parseLongName(UTF8StreamJsonParser.java:1862)
     com.fasterxml.jackson.core.json.UTF8StreamJsonParser.parseMediumName2(UTF8StreamJsonParser.java:1844)
     com.fasterxml.jackson.core.json.UTF8StreamJsonParser.parseMediumName(UTF8StreamJsonParser.java:1801)
[...]

PR #1358, opened 2024-11-09
fail if int overflow in _longNameOffset

pjfanning:
* relates to #1289 
* the class relies on an array so can't handle the case where _longNameOffset ends up greater than Integer.MAX_VALUE
* I think it is helpful to fail with a readable exception - instead of array size exceptions
* It may or may not be useful to include details about how to disable canonicalisation

Only people who are seriously abusing JSON and creating names in their JSON that exceed 2 billion chars will hit this.


cowtowncoder (approved):
Looks good -- I'll modify message slightly but logic is solid!

cowtowncoder:
Good idea: will merge.

Issue #1359, opened 2024-11-13
Non-surrogate characters being incorrectly combined when `JsonWriteFeature.COMBINE_UNICODE_SURROGATES_IN_UTF8` is enabled


jgosselin-accesso:
Version: 2.18.0+

Hello! I believe we've found a bug in the encoding of UTF-8 JSON when the `COMBINE_SURROGATES_AS_UTF8` feature is enabled whereby some characters in the BMP, not belonging to the surrogate blocks, are incorrectly marked as such and therefore combined. This leads to invalid encoding results around those characters.

This happens with some characters that are specifically in the BMP but beyond the surrogate blocks. Notably this includes half- and full-width characters. Characters in the surrogate blocks of D800 - DFFF are correctly treated as surrogates and properly combined when the feature is enabled, and characters before those blocks are also encoded correctly as usual.

Here is a reproducible example:

```java
JsonFactory factory = new JsonFactory();  
ByteArrayOutputStream out = new ByteArrayOutputStream();  
JsonGenerator gen = factory.createGenerator(out);  
  
gen.writeStartObject();  
gen.enable(JsonGenerator.Feature.COMBINE_UNICODE_SURROGATES_IN_UTF8);  
  
// Inside the BMP, beyond surrogate block; 0xFF0C - full-width comma 
gen.writeStringField("test_full_width", "foo" + new String(Character.toChars(0xFF0C)) + "bar");  
  
// Inside the BMP, beyond surrogate block; 0xFE6A - small form percent
gen.writeStringField("test_small_form", "foo" + new String(Character.toChars(0xFE6A)) + "bar");  
  
// Inside the BMP, before the surrogate block; 0x3042 - Hiragana A
gen.writeStringField("test_hiragana", "foo" + new String(Character.toChars(0x3042)) + "bar");  
  
// Outside the BMP; 0x1F60A - emoji
gen.writeStringField("test_emoji", new String(Character.toChars(0x1F60A)));  
  
gen.writeEndObject();  
gen.close();  
  
System.out.write(out.toByteArray());
```

This prints the following (image to show the broken encoding):

<img width="794" alt="Pasted image 20241113150011" src="https://github.com/user-attachments/assets/7bbf1fc3-1426-41bc-8a51-ea3e7eb4a844">

[...]

PR #1360, opened 2024-11-14
fix issue with incorrect matches for some surrogate chars

pjfanning:
see #1359

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/json/UTF8JsonGenerator.java
+++ b/src/main/java/com/fasterxml/jackson/core/json/UTF8JsonGenerator.java
@@ -1511,7 +1511,7 @@ public class UTF8JsonGenerator
                 outputBuffer[outputPtr++] = (byte) (0x80 | (ch & 0x3f));
             } else {
                 // 3- or 4-byte character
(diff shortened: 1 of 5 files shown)


src/main/java/com/fasterxml/jackson/core/json/UTF8JsonGenerator.java
```diff
@@ -2249,8 +2249,8 @@ private byte[] getHexBytes() {
     }
 
     // @since 2.18
-    private boolean _isSurrogateChar(int ch) {
-        return (ch & 0xD800) == 0xD800;
+    private static boolean _isStartOfSurrogatePair(final int ch) {
+        return (ch & 0xF800) == 0xD800;
```
pjfanning, line 2253: I made this change based on comments in #1360 but would like to have it reviewed

cowtowncoder: Updated my comment: I think mask actually needs to be `0xFC00` for First surrogate (i.e. not match Second part of surrogate pair). Will change.

PR #1369, opened 2024-12-23
add propertyNamesSpliterator

pjfanning:
will likely break downstream Jackson projects until we implement the equivalent method

### DIFF
--- a/src/main/java/tools/jackson/core/TreeNode.java
+++ b/src/main/java/tools/jackson/core/TreeNode.java
@@ -6,6 +6,7 @@
 package tools.jackson.core;
 
 import java.util.Iterator;
+import java.util.Spliterator;
 
 /**
  * Marker interface used to denote JSON Tree nodes, as far as
@@ -226,6 +227,17 @@ public interface TreeNode
      *   has (if Object node); empty {@link Iterator} otherwise (never {@code null}).
      */
     Iterator<String> propertyNames();
+
+    /**
+     * Method for accessing names of all properties for this node via a <code>Spliterator</code> ,
+     * if (and only if) this node is an Object node. Number of property names accessible
+     * will be {@link #size}.
+     *
+     * @return A <code>Spliterator</code> for traversing names of all properties this Object node
+     *   has (if Object node); empty {@link Spliterator} otherwise (never {@code null}).
+     * @since 3.0
+     */
+    Spliterator<String> propertyNamesSpliterator();
 
     /**
      * Method for locating node specified by given JSON pointer instances.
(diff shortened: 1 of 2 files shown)


JooHyukKim:
Is this part of any JSTEP? If not I think we can add somewhere in JSTEP for migration doc later on.

cowtowncoder:
> Is this part of any JSTEP? If not I think we can add somewhere in JSTEP for 2->3 migration doc later on.

Would belong under https://github.com/FasterXML/jackson-future-ideas/wiki/JSTEP-3 and I think @pjfanning probably noticed my adding #4863 (sort of related). `TreeNode` additions are bit limited due to Java limitation on co-variance (cannot add return value of `Stream<TreeNode>` because not legal to override with `Stream<JsonNode>`) but can do this for property names (since keys are still `String`s).

cowtowncoder:
@pjfanning Quick question on need/rationale here: I forget details of how `Spliterator`s are useful (as building blocks for Streams... ? ). So would this be useful for actual use cases?

pjfanning:
I saw the changes for stream support and spliterators are most used as an iterator that can be split and processed in parallel (eg parallel streams).
The changes don't look very drastic so I'm happy to not merge the changes and wait to see if there is explicit demand for this.

cowtowncoder:
> I saw the changes for stream support and spliterators are most used as an iterator that can be split and processed in parallel (eg parallel streams). The changes don't look very drastic so I'm happy to not merge the changes and wait to see if there is explicit demand for this. What I mean is that changes like this could be added at any time if needed and they don't really need a major release.

Correct, except for one thing: to add method in `TreeNode` there either needs to be default method implementation, or needs to go in a major version.

I'll have a look at `jackson-databind` side and see if that makes.

Also, fwtw, I try to keep [JSTEP-3](https://github.com/FasterXML/jackson-future-ideas/wiki/JSTEP-3) updated with changes to `JsonNode` / `TreeNode`, although it's obviously challenging to keep related-but-not-synced things in-sync manually (github not having something for "uber-issues" like... epics or something).

pjfanning:
I've added a propertyNamesSpliterator default impl.

cowtowncoder (approved):

cowtowncoder:
Realized somewhat late that name needs to be `propertyNameSpliterator()` since we will use:

1. Plulars for actual `Collection`/`Set` (JsonNode.values())`
2. But Singulars for streams (`valueStream()`) and thereby spliterators as well.

Will change accordingly.

JooHyukKim:
> Realized somewhat late that name needs to be `propertyNameSpliterator()` since we will use:

Is there a written form of this convention somewhere? We could probably go make one in the main '/Jackson' repo

cowtowncoder:
Right: there's JSTEP-6 (https://github.com/FasterXML/jackson-future-ideas/wiki/JSTEP-6) for this. But I need to update that wrt this part of naming.

Issue #1370, opened 2024-12-29
Convert handling of failing tests (JSTEP-10)


JooHyukKim:
As part of [JSTEP-10](https://github.com/FasterXML/jackson-future-ideas/wiki/JSTEP-10), specifically...

    Change 2 : Replace maven-surefire-plugin with Jackson's new failing test management

PR #1371, opened 2024-12-29
[JSTEP-10] Add failing test utils 

JooHyukKim:
part of #1370.

Moving tests & cleaning up will follow

### DIFF
--- a/src/test/java/com/fasterxml/jackson/core/testutil/failure/JacksonTestFailureExpected.java
+++ b/src/test/java/com/fasterxml/jackson/core/testutil/failure/JacksonTestFailureExpected.java
@@ -1,11 +1,11 @@
 package com.fasterxml.jackson.core.testutil.failure;
 
-import org.junit.jupiter.api.extension.ExtendWith;
-
 import java.lang.annotation.ElementType;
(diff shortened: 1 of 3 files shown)


cowtowncoder (approved):

cowtowncoder:
@JooHyukKim One idea -- not sure how easy to do, but possible -- it is possible to export and publish separate test jar by Maven, to share common test classes. If so, that should probably be done from `jackson-core` since almost everything else depends on it (directly or indirectly). If so, it could include these 3 classes, plus some version of current `JUnit5TestBase` (which can then be extended by module-specific classes like `DatabindTestUtil`)

JooHyukKim:
Hmmmm, I clearly  see how we could benefit from it. But I am not sure if it is more beneficial to do it during 2->3 transition phase.

I am concerned that it would sort of expand `JSTEP-10`'s scope more, which also means we have less time to touch other `TODO 3.0` things as well.

cowtowncoder:
@JooHyukKim Maybe we should allow bit of duplication at first, and get bigger conversion done now. And then test-jar publishing could be for 3.0 only, as the next step.

So let's defer reusability concern after more of conversion has been done.

PR #1372, opened 2024-12-30
[JSTEP-10] Migrate from using `/failing` tests to `/tofix` tests

### LINKED ISSUE #1370: Convert handling of failing tests (JSTEP-10)

JooHyukKim:
Resolves #1370


src/test/java/com/fasterxml/jackson/core/testutil/failure/JacksonTestFailureExpectedInterceptor.java
```diff
@@ -29,6 +29,39 @@ public void interceptTestMethod(Invocation<Void> invocation,
...
+    /**
+     * Interceptor API method that is called to intercept test template
+     * method invocation, like {@link org.junit.jupiter.params.ParameterizedTest}.
+     *<p>
+     * Before failing passing test case, it checks if the test should pass according to the
+     * {@link ExpectedPassingTestCasePredicate} provided in the
+     * {@link JacksonTestFailureExpected} annotation.
+     */
+    @Override
+    public void interceptTestTemplateMethod(Invocation<Void> invocation,
```
JooHyukKim, line 41: This API is for `@ParameterizedTest`.
For every method invocation, this method is called.

src/test/java/com/fasterxml/jackson/core/testutil/failure/JacksonTestFailureExpectedInterceptor.java
```diff
@@ -29,6 +29,39 @@ public void interceptTestMethod(Invocation<Void> invocation,
...
+
+        // Before we fail the test for passing,
+        // check if the test should pass according to the predicate given
+        ExpectedPassingTestCasePredicate predicate = findzExpectedPassingTestCasePredicate(invocationContext);
+        if (predicate != null) {
+            if (predicate.shouldPass(invocationContext.getArguments())) {
+                // do-nothing, we do not expect an exception
+                return;
+            }
+        }
```
JooHyukKim, line 61: What we are trying to do here is use `ExpectedPassingTestCasePredicate` to allow passing test being part of test cases of provided by `@ParameterizedTest` this way we can keep test cases in one place (like `@MethodSource`)

JooHyukKim:
@cowtowncoder Now that we have additional implementation like one here, around `@ParameterizedTest`, now I feel more need for [the test jar deployment you suggested.](https://github.com/FasterXML/jackson-core/pull/1371#issuecomment-2564919026). But still hold [the same about the timing](https://github.com/FasterXML/jackson-core/pull/1371#issuecomment-2565333296)(when to do)

src/test/java/com/fasterxml/jackson/core/testutil/failure/JacksonTestFailureExpectedInterceptor.java
```diff
@@ -29,6 +29,39 @@ public void interceptTestMethod(Invocation<Void> invocation,
...
+    @Override
+    public void interceptTestTemplateMethod(Invocation<Void> invocation,
+            ReflectiveInvocationContext<Method> invocationContext, ExtensionContext extensionContext)
+        throws Throwable
+    {
+
+        try {
+            invocation.proceed();
+        } catch (Throwable t) {
+            // do-nothing, we do expect an exception
```
cowtowncoder, line 49: Is there an expectation of always getting an exception, to indicate failing test? I guess this is true in that assert-methods do throw assertion exceptions.

src/test/java/com/fasterxml/jackson/core/tofix/LocationOfError1180Test.java
```diff
@@ -195,4 +199,18 @@ public String toString()
         public final int lineNr;
         public final int columnNr;
     }
+
+    public static class ShouldPredicate1180Test
+            implements ExpectedPassingTestCasePredicate
+    {
+        @Override
+        public boolean shouldPass(List<Object> arguments) {
```
cowtowncoder, line 207: I am not so sure about this -- seems bit fragile and not very easy to understand. Is something like this expected for all "tofix" parameterized tests?

cowtowncoder: Oh. Looks like this is just one case, maybe there is something special about that one.

src/test/java/com/fasterxml/jackson/core/testutil/failure/ExpectedPassingTestCasePredicate.java
```diff
@@ -0,0 +1,15 @@
+package com.fasterxml.jackson.core.testutil.failure;
+
+import java.util.List;
+
+/**
+ * Interface that defines a predicate to determine if a test case is expected to pass.
```
cowtowncoder, line 6: I assume "pass" here mean that test completes without exception? (if so, fine, if not should probably be explained further).

cowtowncoder (approved):

PR #1374, opened 2024-12-31
Fix #1373: change return type of `TreeNode.propertyNames()`

### LINKED ISSUE #1373: Change `TreeNode.propertyName()` return type from `Iterator<JsonNode>` to `Collection<JsonNode>` (3.0)

cowtowncoder:


src/main/java/tools/jackson/core/TreeNode.java
```diff
@@ -227,7 +228,7 @@ public interface TreeNode
      * @return An iterator for traversing names of all properties this Object node
      *   has (if Object node); empty {@link Iterator} otherwise (never {@code null}).
      */
```
pjfanning, line 230: @cowtowncoder the javadoc is out of date (mentions Iterator). Can the new the new javadoc mention if the collection returned is the live version of the data or a safe copy?

If this no longer returns iterator, then we probably don't need the new spliterator method either. Users can get their own iterator or spliterator from the collection.

cowtowncoder: Thanks will update. Live-ness is up to implementation tho, `TreeNode` does not control that aspect.

PR #1375, opened 2024-12-31
Remove spliterator

pjfanning:
* not really useful any more now that we return the Collection
* if we really want to still support returning the spliterator then we should also support returning the iterator - I still think that users can get these from them from the collection


cowtowncoder (approved):

cowtowncoder:
Ok. Easy come, easy go.

Issue #1380, opened 2025-01-09
Change 3.0 to use `module-info.java` directly (and not via Moditect)


cowtowncoder:
In Jackson 2.x we use Moditect plugin to add `module-info.class`, instead of relying on `javac`. This because build is done using JDK 8 which cannot handle `module-info.java`.

But with Jackson 3.0 using JDK 17 or above we can and should convert.

NOTE: all modules should do this, except for `jackson-annotations` which remains JDK 8 built for now.

cowtowncoder:
> I think module-info.class needs to be put in META-INF/versions/11 (or 9). I'm not sure if [fb5ad1b](https://github.com/FasterXML/jackson-core/commit/fb5ad1b8071a948a92b37904dd5838b4b84aba0a) would do that.

Not sure why -- minimum JDK for master/3.0 is JDK 17. Do you have a link to something that explains possible reasons to do so? (docs I saw just talked about `src/main/java/` (and `src/test/java` for test version)

Although I guess I could try and see what happens. :)

pjfanning:
> > I think module-info.class needs to be put in META-INF/versions/11 (or 9). I'm not sure if [fb5ad1b](https://github.com/FasterXML/jackson-core/commit/fb5ad1b8071a948a92b37904dd5838b4b84aba0a) would do that.
> 
> Not sure why -- minimum JDK for master/3.0 is JDK 17. Do you have a link to something that explains possible reasons to do so? (docs I saw just talked about `src/main/java/` (and `src/test/java` for test version)
> 
> But either way I think it's shade plugin that is somehow filtering/not-copying module-info.class, regardless of where it needs to go.

You are probably right - that the module-info.class only needs to be in META-INF/versions when you still support Java 8.

cowtowncoder:
I think we should not put it under `META-INF/versions/` any longer for 3.0.0 -- but not sure how to prevent Moditect from doing that...

pjfanning:
@cowtowncoder I asked a GenAI tool and it agrees that module-info.class should be in jar root dir if we don't support Java 8 any more. The META-INF/versions location for module-info.class is only needed if we need to support Java 8 users.

cowtowncoder:
Alas, after trying to change this, I  suspect we actually still need it, unfortunately... because of shading of different versions of "FastDoubleParser" (I also consulted GenAI which showed how to change things, but no change in inclusion).

pjfanning:
[...]

PR #1381, opened 2025-01-09
Trying to implement #1380: remove use of Moditect plug-in for 3.0

cowtowncoder:
For #1380


cowtowncoder:
@GedMarc Ok, I managed to put this together so that it finally builds & tests run. But for some reason, `module-info.class` is not included it seems.

I was hoping you could sanity check whether changes make sense (plus if you have an idea of where `module-info` disappears, that'd be nice bonus too :) )

cowtowncoder:
FUDGE. It is all thanks to %)@$(#)^$#^# `maven-shade-plugin` which ALWAYS drops `module-info.class`: first if "minimize" is enabled and then, well, drop it anyway no matter what.
Apparently no way to configure it to be retained.

Will see if there's a way to RE-ADD it tho.

cowtowncoder:
Ok and the blocker is this issue:

https://issues.apache.org/jira/browse/MSHADE-261

Because Maven shade plug-in will drop `module-info.class` (starting with 3.1.0) if:

1. "minimize" is used, OR
2. ... just for funsies even if not (there's second check)

cowtowncoder:
Ok this has been tough one, but at this point build fully works, ending up with `module-info.class` under JDK 17 folder (`META-INF/versions/17/module-info.class`). To make that happen:

1. Build uses `src/main/java/module-info.java` as it should
2. Tests use `src/main/java/module-info.java` as they should
3. Jar plugin is instructed to drop `/module-info.class` because...
4. we need Moditect to re-build and stash `module-info.java` as `META-INF/versions/17/module-info.class` (and it won't if there is already `module-info.class` hence previous removal)
5. maven-shade-plugin cannot forcibly remove `module-info.class` (it does from main level but NOT from jvm-specific ones. There!)

which is pretty convoluted way to make it. But I could not figure a simpler way... :-(
Still, seems fine.

Also had to simplify FastDoubleParser shading; use of names did not work for some reason.

Next to tackle https://github.com/FasterXML/jackson-databind/issues/4891 (will try to get that working before merging this in).

cowtowncoder:
@GedMarc @pjfanning @JooHyukKim Just FYI: getting things properly JPMS modularized is pretty challenging. But after `jackson-core` and `jackson-databind`, hopefully bit more straight-forward (will see if shading should be removed from Afterburner/Blackbird/MrBean tho)

PR #1386, opened 2025-01-12
Fix #1385: produce test-support jar

### LINKED ISSUE #1385: Create `jackson-core-[VERSION]-tests.jar` to contain shared test utility classes
There is a bit of test functionality that should be shared, to avoid duplicating test classes in every Jackson component: most notably new `tools.jackson.core.testutil.failure.JacksonTestFailureExpected` annotation.

It seems possible to create separate test jar, see f.ex: https://maven.apache.org/plugins/maven-jar-plugin/examples/create-test-jar.html

cowtowncoder:


cowtowncoder:
@JooHyukKim Let's see if this works: I can see test jar but need to see how to use it from `jackson-databind` once published.

Also: thinking of adding this only for 3.0/master at first; if it works could backport in 2.19 I suppose.

JooHyukKim:
Nice didn't know there was such thing.
Hopefully maintenance wrt artifact publishing cost is low :-o

cowtowncoder:
> Nice didn't know there was such thing. Hopefully maintenance wrt artifact publishing cost is low :-o

Should be no worse than regular publishing/dependencies; we should get -SNAPSHOT builds automatically along with regular artifact.

That is, if it all works out. We'll see. I'll merge this and see how things go.

PR #1388, opened 2025-01-13
Further work on #1385: try moving test classes to different package

cowtowncoder:
Doing this is required to properly separate jackson-core packages (actual streaming package) and test support classes published/exported for other components' tests to use.
(no package splitting with JPMS)


src/main/java/tools/jackson/core/io/JsonStringEncoder.java
```diff
@@ -34,10 +34,10 @@ public final class JsonStringEncoder
     //    to estimate ok initial encoding buffer, switch to segmented for
     //    possible (but rare) big content
 
-    final static int MIN_CHAR_BUFFER_SIZE = 16;
-    final static int MAX_CHAR_BUFFER_SIZE = 32000; // use segments beyond
-    final static int MIN_BYTE_BUFFER_SIZE = 24;
-    final static int MAX_BYTE_BUFFER_SIZE = 32000; // use segments beyond
+    public final static int MIN_CHAR_BUFFER_SIZE = 16;
```
cowtowncoder, line 37: Not ideal, but none of visibility increases should be problematic per se.
Should minimize testing based on of such constants of course but first things first.

cowtowncoder:
Ok: things work at this end so let's go ahead and see if we can make `jackson-databind` work too.

PR #1391, opened 2025-01-25
Fix issue where the parser can read back old number state when parsing later numbers

pjfanning:
* new version of #1389 but targeted at 2.17 branch
* only the original test in https://github.com/FasterXML/jackson-databind/pull/4918 is affected by the bug - similar tests (V2, V3, etc) are not affected.
* Aim is to use this fix for patch releases but to later try a bigger change to better manage number state in the parser code - in the 2.19 development branch

### DIFF
--- a/release-notes/VERSION-2.x
+++ b/release-notes/VERSION-2.x
@@ -14,6 +14,12 @@ a pure JSON library.
 === Releases ===
 ------------------------------------------------------------------------
 
+2.17.4 (not yet released)
+
+#1391: Fix issue where the parser can read back old number state when
+  parsing later numbers
+ (fix contributed by @pjfanning)
+
 2.17.3 (01-Nov-2024)
 
 #1331: Update to FastDoubleParser v1.0.1 to fix `BigDecimal` decoding problem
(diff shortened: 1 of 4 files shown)


cowtowncoder:
Ok that sounds reasonable. One reason why I think this fix is not ideal is that it actually does not clear the state unless value is accessed -- so it just covers this specific case and not others. But at the same it is somewhat minimal fix which is a plus for patch releases.

cowtowncoder:
Hmmmh. Playing with tests, the only fix that has any effect here is in `getDoubleValue()`, line 864. But it looks like `_numberString` is actually being cleared via `convertNumberToDouble()` which calls `_getNumberDouble()` where `_numberString` is actually cleared.

I think I'll trim down changes slowly to parts that seem useful in general (namely, immediate return after "convertXxx()" calls). And then look into `2.19` for more general clearing.

cowtowncoder:
Merged all the way to `master`: for `2.19` and `master` has clearing of `_numberString` on `reset[Number]()` methods as safeguard.

Thank you for test & fix @pjfanning !

Issue #1397, opened 2025-02-04
Jackson changes additional values to infinite in case of special JSON structures and existing infinite values


Rodenstock:
With infinite values of float types and the following JSON structure, Jackson unexpectedly changes additional values since version 2.18.0.
Radius is originally set to infinity (1.7976931348623157e+308). After selecting the value as JSONB from a PostgreSQL database and serializing it to JSON, the value of center.x is also set to infinity (1.7976931348623157e+308).

All versions greater than or equal to 2.18.0 are affected.
The last working version is 2.17.3. 

```json
{
  "results": [
    {
      "radius": 179769313486231570000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000,
     "type": "center",
      "center": {
        "x": -11.0,
        "y": -2.0
      }
    }
  ]
}
```

Jackson serializes the JSON-Object into the following JSON:

```json
{
  "results": [
    {
      "radius": 1.7976931348623157e+308,
     "type": "center",
      "center": {
        "x": 1.7976931348623157e+308,
        "y": -2.0
      }
    }
  ]
}
```

Below is the code for reproducing the issue.

```java
package org.example;

import com.fasterxml.jackson.annotation.JsonSubTypes;
import com.fasterxml.jackson.annotation.JsonTypeInfo;
import com.fasterxml.jackson.databind.ObjectMapper;

import java.io.IOException;
import java.util.Arrays;

class Point {
    private Double x;
    private Double y;

    public Double getX() {
        return x;
    }

    public void setX(Double x) {
        this.x = x;
    }

    public Double getY() {
        return y;
    }

    public void setY(Double y) {
        this.y = y;
    }
}

@JsonTypeInfo(
        use = JsonTypeInfo.Id.NAME,
        include = JsonTypeInfo.As.EXISTING_PROPERTY,
        property = "type",
        visible = true)
@JsonSubTypes(@JsonSubTypes.Type(value = CenterResult.class, name = "center"))
abstract class Result {
    private String type;

    public String getType() {
        return type;
    }

    public void setType(String type) {
        this.type = type;
    }
}

class CenterResult extends Result {
    private Point center;

    private Double radius;

    public Double getRadius() {
        return radius;
    }

    public void setRadius(Double radius) {
        this.radius = radius;
    }

[...]

PR #1400, opened 2025-02-04
Changes post #4917: forcibly clear `_numberString` when new number token set

pjfanning:
* based on extra report in #1397 
* additional test case added


cowtowncoder:
Ok so we are thinking of just biting the bullet and applying full fix? That sort of makes sense -- I was wondering if the small patch addressing just one aspect might be too little.

cowtowncoder (approved):

PR #1404, opened 2025-02-20
add utf8 parsing test

pjfanning:
Useful to test that we don't suffer from https://github.com/FasterXML/jackson-dataformats-text/issues/497 in the jackson-core utf8 parsing code.


cowtowncoder (approved):
LGTM

PR #1411, opened 2025-03-05
Add `JsonFactoryBuilder.configureForJackson2()`, `JsonFactoryBuilder builderWithJackson2Defaults()`

pjfanning:
* Relates to https://github.com/FasterXML/jackson-databind/issues/4986

### DIFF
--- a/src/main/java/tools/jackson/core/json/JsonFactory.java
+++ b/src/main/java/tools/jackson/core/json/JsonFactory.java
@@ -183,12 +183,8 @@ public class JsonFactory
 
     /**
      * Main factory method to use for constructing {@link JsonFactory} instances with
-     * different configuration. The builder returned uses default settings more closely
-     * matched the default configs used in Jackson 2.x versions.
-     * <p>
-     *     This method is still a work in progress and may not yet fully replicate the
-     *     default settings of Jackson 2.x.
-     * </p>
+     * different configuration.
+     *
      * @return Builder instance to use
      */
     public static JsonFactoryBuilder builder() {
@@ -197,7 +193,12 @@ public class JsonFactory
 
     /**
      * Factory method to use for constructing {@link JsonFactory} instances with
-     * different configuration.
+     * different configuration. The builder returned uses default settings more closely
+     * matched the default configs used in Jackson 2.x versions.
+     * <p>
+     *     This method is still a work in progress and may not yet fully replicate the
+     *     default settings of Jackson 2.x.
+     * </p>
      *
      * @return Builder instance to use
      */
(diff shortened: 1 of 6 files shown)


src/main/java/tools/jackson/core/json/JsonFactory.java
```diff
@@ -183,14 +183,32 @@ public JsonFactoryBuilder rebuild() {
 
     /**
      * Main factory method to use for constructing {@link JsonFactory} instances with
-     * different configuration.
-     *
+     * different configuration. The builder returned uses default settings more closely
```
cowtowncoder, line 186: I think this Javadoc is accidentally swapped with next one?

pjfanning: fixed just now

cowtowncoder:
Makes sense as a starting point, but if the idea is to add same functionality for other backends, should probably implement in a way that can share general (non-JSON-specific) settings.
So parts in `TokenStreamFactory`/`TSFBuilder`, parts in `JsonFactory`/`JsonFactoryBuilder`.

And in fact, instead of (or in addition to?) having all of these in Builder factory method, could perhaps have builder member method like `TSFBuilder.configureForJackson2()`.
This could be called separately, as well as part of factory method?

pjfanning:
> Makes sense as a starting point, but if the idea is to add same functionality for other backends, should probably implement in a way that can share general (non-JSON-specific) settings. So parts in `TokenStreamFactory`/`TSFBuilder`, parts in `JsonFactory`/`JsonFactoryBuilder`.
> 
> And in fact, instead of (or in addition to?) having all of these in Builder factory method, could perhaps have builder member method like `TSFBuilder.configureForJackson2()`. This could be called separately, as well as part of factory method?

I thought about having it on the builder but my worry is that users will then use annoying ordering in chaining their commands to configure the builder. Someone is bound to come along and create a build, set some config then call `configureForJackson2()` and that could undo some of their pre-set config. I guess you could say that about any command chaining on the builder but the actions of `configureForJackson2()` are a bit opaque - you need to read the code to see hat it does.

cowtowncoder:
@pjfanning I still suggest adding configure method in Builder instance even if called from static factory method that creates Builders. That allows composition of `TSFBuilder` (format-agnostic settings) and format-specific actually builder (like `JsonFactoryBuilder`).

Configure method can then be alternatively called by someone being only handed Builder, as opposed to only being available from Factory.

pjfanning:
> configureForJackson2

I added configureForJackson2 in latest commit.

cowtowncoder (approved):

Issue #1144, opened 2023-11-24
`FilteringParserDelegate` can go into an infinite loop if underlying parser is non-blocking


simonbasle:
Jackson version (BOM): `2.15.2`

While experimenting with `JsonPointer` in Spring Framework, I tried to build up on the existing `FilteringParserDelegate` and associated `JsonPointerBasedTokenFilter`. Unfortunately, this doesn't seem to work with a non-blocking `JsonParser` delegate, for several reasons:
 - it doesn't delegate method `canParseAsync()` (ends up returning the default interface implem => `false`)
 - it doesn't delegate method `getNonBlockingInputFeeder()` (same as above => `null`)
 - it isn't aware of `JsonToken.NOT_AVAILABLE`

While the first two are easily circumvented by extending `FilteringParserDelegate`, the last one is the truly problematic one. It results in going down the code path of a "scalar value", and will continuously call `delegate.nextToken()` which continues to result in `NOT_AVAILABLE`.

This leads to an infinite loop when attempting to filter based on a `JsonPointer` on top of a non-blocking parser.

Is there anything fundamentally preventing `FilteringParserDelegate` from being compatible with non-blocking parsers that I might have overlooked? Otherwise I think it's pretty close, hopefully that can be fixed 😄

pjfanning:
Can you provide reproducible test cases? Volunteers appreciate time savers like that.

simonbasle:
Sure! Here is a testcase, probably needs some refinement (especially around
asserting timeout with assertj).
Please note that JUnit `Timeout` rule doesn't work to cut the loop short if
the call isn't offloaded to another thread.

```
    public void testFilteringNonBlockingParser() throws IOException {
        JsonFactory factory = new JsonFactoryBuilder().build();
        JsonParser nonBlockingParser =
factory.createNonBlockingByteArrayParser();
        ByteArrayFeeder inputFeeder = (ByteArrayFeeder)
nonBlockingParser.getNonBlockingInputFeeder();
        String json = "{\"first\":1,\"second\":2}";
        byte[] jsonBytes = json.getBytes(StandardCharsets.UTF_8);

        JsonParser filteringParser = new
FilteringParserDelegate(nonBlockingParser, new
JsonPointerBasedFilter("/second"),
                TokenFilter.Inclusion.ONLY_INCLUDE_ALL, false);

        ScheduledExecutorService executor =
Executors.newSingleThreadScheduledExecutor();
        executor.schedule(() -> {
[...]

PR #1414, opened 2025-03-24
Skip RS CTRL-CHAR to support JSON Text Sequence (RFC7464)

iifawzi:
Hello, This PR tries to resolve https://github.com/FasterXML/jackson-core/issues/633

This is my first PR for the project, so I could be missing a point. Feel free to guide me through points that might need to be added. 

I didn't implement it in the data input parser as it seems like we're skipping the validation of WS
https://github.com/FasterXML/jackson-core/blob/99f749d958a55013724571d825625e1032af9d4b/src/main/java/com/fasterxml/jackson/core/json/UTF8DataInputJsonParser.java#L2299-L2304
so not sure if we need to have it now, the same way we do for other parsers

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/JsonParser.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonParser.java
@@ -194,6 +194,19 @@ public abstract class JsonParser
         ALLOW_UNQUOTED_CONTROL_CHARS(false),
 
         /**
+         * Feature that determines whether parser will allow
+         * Record Separator (RS) control character ({@code 0x1E})
+         * as part of ignorable whitespace in JSON input, similar to the TAB character.
+         * <p>
+         * Since the official JSON specification permits only a limited set of control
+         * characters as whitespace, this is a non-standard feature and is disabled by default.
+         * </p>
+         *
+         * @since 2.18
+         */
+        ALLOW_RS_CONTROL_CHAR(false),
+
+        /**
          * Feature that can be enabled to accept quoting of all character
          * using backslash quoting mechanism: if not enabled, only characters
          * that are explicitly listed by JSON specification can be thus
--- a/src/main/java/com/fasterxml/jackson/core/JsonParser.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonParser.java
@@ -202,7 +202,7 @@ public abstract class JsonParser
          * characters as whitespace, this is a non-standard feature and is disabled by default.
          * </p>
          *
-         * @since 2.18
+         * @since 2.19
          */
         ALLOW_RS_CONTROL_CHAR(false),
 
--- a/src/main/java/com/fasterxml/jackson/core/JsonParser.java
+++ b/src/main/java/com/fasterxml/jackson/core/JsonParser.java
@@ -194,16 +194,9 @@ public abstract class JsonParser
         ALLOW_UNQUOTED_CONTROL_CHARS(false),
 
         /**
-         * Feature that determines whether parser will allow
-         * Record Separator (RS) control character ({@code 0x1E})
-         * as part of ignorable whitespace in JSON input, similar to the TAB character.
-         * <p>
-         * Since the official JSON specification permits only a limited set of control
-         * characters as whitespace, this is a non-standard feature and is disabled by default.
-         * </p>
-         *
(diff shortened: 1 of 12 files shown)


src/main/java/com/fasterxml/jackson/core/base/ParserMinimalBase.java
```diff
@@ -768,7 +769,13 @@ protected void reportUnexpectedNumberChar(int ch, String comment) throws JsonPar
     
     protected void _throwInvalidSpace(int i) throws JsonParseException {
         char c = (char) i;
-        String msg = "Illegal character ("+_getCharDesc(c)+"): only regular white space (\\r, \\n, \\t) is allowed between tokens";
+        String msg = "Illegal character ("+_getCharDesc(c)+"): only regular white space (\\r, \\n, \\t, \\u001E) is allowed between tokens";
```
iifawzi, line 772: Given the note we have below, not sure if still need to include it here. I believe it's better to be removed given it's managed by a feature (not regular?)

src/main/java/com/fasterxml/jackson/core/JsonParser.java
```diff
@@ -193,6 +193,19 @@ public enum Feature {
...
+        /**
+         * Feature that determines whether parser will allow
+         * Record Separator (RS) control character ({@code 0x1E})
+         * as part of ignorable whitespace in JSON input, similar to the TAB character.
+         * <p>
+         * Since the official JSON specification permits only a limited set of control
+         * characters as whitespace, this is a non-standard feature and is disabled by default.
+         * </p>
+         *
+         * @since 2.18
```
pjfanning, line 205: 2.19 is the right value here

src/main/java/com/fasterxml/jackson/core/json/JsonReadFeature.java
```diff
@@ -80,6 +80,19 @@ public enum JsonReadFeature
...
+    /**
+     * Feature that determines whether parser will allow
+     * Record Separator (RS) control character ({@code 0x1E})
+     * as part of ignorable whitespace in JSON input, similar to the TAB character.
+     * <p>
+     * Since the official JSON specification permits only a limited set of control
+     * characters as whitespace, this is a non-standard feature and is disabled by default.
+     * </p>
+     *
+     * @since 2.18
```
pjfanning, line 92: 2.19

src/main/java/com/fasterxml/jackson/core/JsonParser.java
iifawzi: corrected, thanks

src/main/java/com/fasterxml/jackson/core/JsonParser.java
```diff
@@ -193,6 +193,12 @@ public enum Feature {
         @Deprecated
         ALLOW_UNQUOTED_CONTROL_CHARS(false),
 
+        /**
+         * @deprecated Use {@link com.fasterxml.jackson.core.json.JsonReadFeature#ALLOW_RS_CONTROL_CHAR} instead
```
cowtowncoder, line 197: May look funny but `JsonParser.Feature` is to be removed from 3.0 yet is needed internally (all `StreamReadFeature`s and `JsonReadFeature`s must map back to one entry).
But we don't want users to use `JsonParser.Feature` any more.

cowtowncoder (approved):

cowtowncoder:
@iifawzi Implementation looks good; refactored things slightly wrt tests but functionally it was fine.
Did not remember that `DataInput`-backed parser skips WS validation; agree that we'll go with it for now.

The only concern I have is merging to 3.0/master -- that will be some fun. But I'll take care of it.

src/main/java/com/fasterxml/jackson/core/JsonParser.java
iifawzi: It's good to know about that to help take care of it next time. 
Thanks for your collaboration on the PRs!

cowtowncoder:
@iifawzi thank you! It is great to hear that! And thank you for your contributions so far.

On RS in content: unless there's strong use case for allowing it (that is, some actually needing it for specific use), I'd probably not want to add support for in-content. #1144 would be good to try to figure out.

PR #1415, opened 2025-03-30
Fixing parser incorrect location when fed with non-zero offset

### LINKED ISSUE #1412: More cases of Non-blocking parser reporting incorrect locations when fed with non-zero offset
Related to https://github.com/FasterXML/jackson-core/issues/531 -- it looks like there's still an offset calculation glitch when `ByteArrayFeeder.feed()` is called with a non-zero array offset.

Jackson version: 2.18.3

I've attached a slightly more comprehensive unit test. `feedByteByByteFromOffsetZero()` passes, but the others fail.

Thanks!

[ByteArrayFeederOffsetTest.java.txt](https://github.com/user-attachments/files/19330280/ByteArrayFeederOffsetTest.java.txt)

iifawzi:
Hello, this PR is an attempt to resolve #1412 
The issue with the changed line is that it was double counting the offset as `_inputPtr` keeps increasing

This change ensures we track the relative position of each token in the JSON input stream, rather than being affected by the absolute array indices used when feeding the data.

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/json/async/NonBlockingUtf8JsonParserBase.java
+++ b/src/main/java/com/fasterxml/jackson/core/json/async/NonBlockingUtf8JsonParserBase.java
@@ -138,7 +138,8 @@ public abstract class NonBlockingUtf8JsonParserBase
 
         // No: fresh new token; may or may not have existing one
         _numTypesValid = NR_UNKNOWN;
-        _tokenInputTotal = _currInputProcessed + _inputPtr;
+        _tokenInputTotal = _currInputProcessed + (_inputPtr - _currBufferStart);
+
         // also: clear any data retained so far
         _binaryValue = null;
         int ch = getNextUnsignedByteFromBuffer();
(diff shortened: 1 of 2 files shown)


cowtowncoder:
Thank you! I'll need to double-check change wrt rest of code but sounds legit.

Aside from that, one thing we'd need before merging in the fix is CLA:

(unless you have sent one earlier)
which is needed before merging the first contribution. Usual way is to print it, fill & sign, scan/photo, email to `cla` at fasterxml dot com. Once we have it I can proceed and merge the fix.

Thank you in advance!

iifawzi:
> Thank you! I'll need to double-check change wrt rest of code but sounds legit.
> 
> Aside from that, one thing we'd need before merging in the fix is CLA:
> 
> https://github.com/FasterXML/jackson/blob/master/contributor-agreement.pdf
> 
> (unless you have sent one earlier) which is needed before merging the first contribution. Usual way is to print it, fill & sign, scan/photo, email to `cla` at fasterxml dot com. Once we have it I can proceed and merge the fix.
> 
> Thank you in advance!


Thank you @cowtowncoder, I forgot to check the guide. 
Signed and sent. 

Let me know if I can assist with any further verifications in any other repository.

cowtowncoder (approved):
LGTM

PR #1417, opened 2025-04-04
Add opt-in support for non blocking parser usage with filter delegate

### LINKED ISSUE #1144: `FilteringParserDelegate` can go into an infinite loop if underlying parser is non-blocking

iifawzi:
Hello, this PR is an initial attempt to resolve #1144 
- It allows explicit usage of a non-blocking parser with the filter delegate through second constructor.
- It throws an error if the NOT_AVAILABLE token is reported (it means the input isn't fed fully at once), and this fixes the infinite loop issue reported

The only limitation is that this is a breaking change, but this remains a limitation for any other way (Factory, etc..)


src/main/java/com/fasterxml/jackson/core/filter/FilteringParserDelegate.java
```diff
@@ -113,13 +114,55 @@ public FilteringParserDelegate(JsonParser p, TokenFilter f,
...
+    }
+
+    /**
+     * @param p Parser to delegate calls to
+     * @param f Filter to use
+     * @param inclusion Definition of inclusion criteria
+     * @param allowMultipleMatches Whether to allow multiple matches
+     * @param allowNonBlockingParser If true, permits NonBlockingJsonParser
+     * @throws IllegalArgumentException if NonBlockingJsonParser is used without explicit permission
+     */
```
cowtowncoder, line 141: Needs `@since 2.19`

src/main/java/com/fasterxml/jackson/core/filter/FilteringParserDelegate.java
```diff
@@ -113,13 +114,55 @@ public FilteringParserDelegate(JsonParser p, TokenFilter f,
...
      * @param inclusion Definition of inclusion criteria
      * @param allowMultipleMatches Whether to allow multiple matches
+     * @throws IllegalArgumentException if NonBlockingJsonParser is used without explicit permission
      */
     public FilteringParserDelegate(JsonParser p, TokenFilter f,
             TokenFilter.Inclusion inclusion, boolean allowMultipleMatches)
     {
         super(p);
+
+        if (p instanceof NonBlockingJsonParser) {
```
cowtowncoder, line 124: Relying on type works only JSON-backed async-parsers (there's one for Smile, which does not extend `NonBlockingJsonParser`).

But there's a better way: check 

    JsonParser.canParseAsync()

which returns `true`

src/test/java/com/fasterxml/jackson/core/json/async/AsyncTokenFilterTest.java
```diff
@@ -35,6 +37,25 @@ public TokenFilter includeProperty(String name) {
         JsonToken.END_OBJECT
     };
 
+    @Test
+    void filteredNonBlockingParserNotExplicitlyAllowed() throws IOException {
+        NonBlockingJsonParser nonBlockingParser = (NonBlockingJsonParser) JSON_F.createNonBlockingByteArrayParser();
+        assertThrows(IllegalArgumentException.class, () ->
```
cowtowncoder, line 43: It'd be good to check exception message, not just type.

cowtowncoder (commented):
Looks pretty good, thank you for submitting.

I added couple of notes for changes, after which I can go ahead and merge this.
Thanks!

src/main/java/com/fasterxml/jackson/core/filter/FilteringParserDelegate.java
iifawzi: Ahh, I didn't even know about Smile. Do you think we should make the error more generic as well?  now it's "`NonBlockingJsonParser` requires explicit permission."

iifawzi:
> Looks pretty good, thank you for submitting.
> 
> I added couple of notes for changes, after which I can go ahead and merge this. Thanks!

Thank you, comments have been resolved.

src/main/java/com/fasterxml/jackson/core/filter/FilteringParserDelegate.java
cowtowncoder: Yes, good idea, should probably include actual class name (`p.getClass().getSimpleName()`?) and mention of it being async due `canParseAsync()` returns `true`.

cowtowncoder: Fwtw would be nice to get async parser for CBOR (being quite similar to Smile). But writing one is lots of work so I haven't had time (or itch to be sure) to implement it.

Of course one could write one for most format backends, but CBOR would be very similar to Smile which would make it easiest.

iifawzi: Done, refactored! Not entirely sure about the phrasing though

On another note, thanks for pointing out the CBOR support — that’s definitely something worth exploring next, alongside digging deeper into the project with these fixes. I think it could be a great addition to Jackson as well.

cowtowncoder:
@iifawzi Thank you again! This is another GREAT contribution, something that had been waiting for someone with time & interest in solving the issue. I love when things work like here.

I am hoping to release 2.19.0-rc1 very soon now, possibly today.

iifawzi:
> @iifawzi Thank you again! This is another GREAT contribution, something that had been waiting for someone with time & interest in solving the issue. I love when things work like here.
> 
> I am hoping to release 2.19.0-rc1 very soon now, possibly today.

Thank you @cowtowncoder, couldn't be done without your colaboration on help into making the project more contributor-welcome by your quick addressing of PRs and failing tests! This's the power of open-source. Thanks goes to you as well. 

Looking forward for the release!

Issue #1421, opened 2025-04-24
Make `JsonReadContext` non-final


isalvadori:
After opening a discussion [here](https://github.com/FasterXML/jackson/discussions/272) about the need (or not) for the `JsonReadContext` to be `final`, it turned out there is no strong reason behind having it final.
Thus, I would like to propose to remove the `final` on that class.
This will allow easier customization of the context.

maho7791:
I wanted to mention that we would need this also for the 3.x versions.
We currently move our EMF mapping for jackson3.

cowtowncoder:
Yes, we merge forward changes in most cases, and do so here.

maho7791:
This was faster than light!!!

cowtowncoder:
@maho7791  fwtw this was due to lucky timing. It's not always this fast -- but sometimes that happens which is great :)

PR #1422, opened 2025-04-24
Allow `JsonReadContext` to be extended

pjfanning:
relates to #1421


cowtowncoder:
We actually need this against 2.19, can merge forward but not backwards.

Will copy this -- thank you @pjfanning !

cowtowncoder:
Re-created as #1423.

Issue #1429, opened 2025-05-09
Clarification: How to properly use maxDocumentLength


r4fterman:
Hi Jackson Team,
I am having an issue getting the max document length enforced. Hope someone can help me out on this.

I run the following unit test using jackson-core:2.18.1:

```java
    @Test
    void shouldFailOnLargeDocument() {
        StreamReadConstraints constraints = StreamReadConstraints.builder()
                .maxDocumentLength(1L)
                .build();

        JsonFactory factory = JsonFactory.builder()
                .streamReadConstraints(constraints)
                .build();

        ObjectMapper mapper = new ObjectMapper(factory);

        String json = "{\"key\":\"value\"}";

        assertThrows(StreamConstraintsException.class, () -> mapper.readTree(json));
    }
```

As the maxDocumentLength is set lower than the actual length of the json document, I expect the object mapper to throw an exception.

Instead JUnit tells me:
```text
org.opentest4j.AssertionFailedError: Expected com.fasterxml.jackson.core.exc.StreamConstraintsException to be thrown, but nothing was thrown.

	at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:152)
	at org.junit.jupiter.api.AssertThrows.assertThrows(AssertThrows.java:73)
	at org.junit.jupiter.api.AssertThrows.assertThrows(AssertThrows.java:35)
	at org.junit.jupiter.api.Assertions.assertThrows(Assertions.java:3082)
	at ...
```

For all other limits on StreamReadConstraints an expection is thrown as expected. Only for maxDocumentLength is not working out. Where do I make a mistake?

Thanks, Roman

[UPDATE]

Seems that JSON parser calls the maxDocumentLength validation only if the input data exceeds a certain size. If the input data is too small, the maxDocumentLength is not enforced by default.

pjfanning:
For performance reasons, the check is only done when a new batch of data is requested by the parser, not after every byte is read. The code is in jackson-core if you want to try the tests and see how it is implemented.
So, in short, the document parsing will fail at a doc size greater than or equal to the configured maxDocumentLength.
It may be feasible to do an exact check when the parser completes the parsing as this is just one additional check per document (ie fail if you have 1 byte more than the maxDocumentLength).
The real intent is not for someone to set a limit of 1 or 100 but a value set in Kb or Mb and that the maxDocumentLength will fail the stream parsing if the data length is a decent chunk bigger than the configured maxDocumentLength.

[...]

PR #1430, opened 2025-05-09
update docs about maxDocumentLength to say the check is not exact

pjfanning:
see #1429


cowtowncoder (approved):
Very good clarification. Will backport to 2.18/2.19 as well.

PR #1439, opened 2025-05-30
Fix #1438: change `ParserBase.close()` to clear `_currToken`

### LINKED ISSUE #1438: `ParserBase.close()` does not clear `_currToken`
Looks like default implementation of `close()` in `ParserBase` is missing this line:

    _currToken = null;

leaving "dangling" current token. Should be changed.

(note: found when testing changes for #1433)

cowtowncoder:


src/main/java/com/fasterxml/jackson/core/base/ParserBase.java
```diff
@@ -403,6 +403,8 @@ public void overrideCurrentName(String name) {
             // 19-Jan-2018, tatu: as per [core#440] need to ensure no more data assumed available
             _inputPtr = Math.max(_inputPtr, _inputEnd);
             _closed = true;
+            // 30-May-2025, tatu: was missing before 2.20
+            _currToken = null;
```
pjfanning, line 407: could this be viewed as a bug and potentially considered for 2.19.1?

cowtowncoder: Potentially too risky for 2.19.1, even if bug (which it seemed to me it is). That's why targeting 2.20.

But... I am now re-considering this change: the reason being that there are 2 test failures from `jackson-databind` for verifying information from `InvalidFormatException`.
(found out via cascading rebuilds).
This is done through exception-included `JsonParser` and it appears that parser gets closed before test has a chance to access the current token.
I'll see if there is something in the test to change, and perhaps `MismatchedInputException` ought to get "current token" directly. But there are many other things parser would have access to, at point of throwing an exception.

Will have to think about this a bit...

cowtowncoder: Ah. Test uses `ObjectReader.readValue(String)` which will create `JsonParser` but also close it (since caller wouldn't have access to it).

So it's probably ok -- can change test, but also explain why new behavior makes sense and how to avoid `JsonParser.close()` from occurring.

cowtowncoder: Hmmmh. I think there's need for new `StreamReadFeature` flag for this, actually.

Issue #1446, opened 2025-06-09
Invalid package reference to "java.lang.foreign" from `com.fasterxml.jackson.core:jackson-core` (from `FastDoubleParser`) [Android]


baywet:
Adding v2.18.3 in an android project leads to the following error message when running gradle build.

```
Error: Invalid package reference in com.fasterxml.jackson.core:jackson-core; not included in Android: java.lang.foreign. Referenced from com.fasterxml.jackson.core.internal.shaded.fdp.v2_18_3.FastDoubleSwar
```

No workaround is known, besides downgrading the dependency.

More context:
- https://github.com/microsoft/kiota-java/actions/runs/15538363819/job/43742832455?pr=1932
- https://github.com/microsoft/kiota-java/pull/1932
- https://github.com/Azure/azure-sdk-for-java/pull/44968

cowtowncoder:
I don't know what could be done here, although this seems odd: `FastDoubleParser` (https://mvnrepository.com/artifact/ch.randelshofer/fastdoubleparser/1.0.90) is included as shaded component, and it has not dependencies of its own.

So is there any chance Android SDK tooling might have an issue to fix? (or maybe this works with certain version of ASDK).

Also worth checking if Jackson 2.19.0 is any different in this regard (none of changes seems relevant but just in case).

cowtowncoder:
Ok. Perhaps this is due to FastDoubleParser version upgrade included in 2.18.1. If so, 2.18.0 might actually work, as it still depends on `1.0.1` instead of `1.0.90`

But I really hope this is not it because... that'd be difficult thing to reconcile wrt compatibility.

/cc @pjfanning .

pjfanning:
FastDoubleSwar is multi-release jar class. That is, a class where the jar has different versions of the class targeted at different Java versions. These versioned classes appear in META-INF/versions/11 where 11 is the Java version.
https://openjdk.org/jeps/238

java.lang.foreign is pretty new so FastDoubleSwar would only be set up to use this package for the latest Java versions (eg Java 21).

Whatever Android set up is being use by @baywet seems not to be able to handle JEP 238.

Is this not Android's problem? The JEP is not new. If they want to use Java classes, shouldn't they be up to date with the JEPs?

baywet:
@cowtowncoder thanks for the quick reply!

Here is the full repo (in android studio):
1. create a new project (empty activity, api level 26)
[...]

PR #1447, opened 2025-06-09
Remove Java 22 files (`FastDoubleParser`)

pjfanning:
To address #1446 

I got build warning about the META-INF/LICENSE and MANIFEST.MF too so I added them to the exclude.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -171,6 +171,8 @@ com.fasterxml.jackson.core.*;version=${project.version}
                     <exclude>META-INF/LICENSE</exclude>
                     <exclude>META-INF/MANIFEST.MF</exclude>
                     <exclude>META-INF/versions/**/module-info.*</exclude>
+                    <!-- 09-Jun-2025, PJ: [core#1446] Exclude Multi-Release classes
+                       for Java 22 due to issues in Android Linting tool -->
                     <exclude>META-INF/versions/22/**</exclude>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -168,7 +168,10 @@ com.fasterxml.jackson.core.*;version=${project.version}
                 <filter>
                   <artifact>ch.randelshofer:fastdoubleparser</artifact>
                   <excludes>
+                    <exclude>META-INF/LICENSE</exclude>
+                    <exclude>META-INF/MANIFEST.MF</exclude>
                     <exclude>META-INF/versions/**/module-info.*</exclude>
+                    <exclude>META-INF/versions/22/**</exclude>
```
cowtowncoder, line 176: Could you add an XML comment with reference to this issue (or one that triggered this), to explain why we are excluding JDK 22 implementation from Jackson 2.x.
For now I think 3.0 can retain newer versions.

pjfanning: comment added

pjfanning: it would probably be useful to port the LICENSE and MANIFEST.MF excludes to Jackson 3

cowtowncoder: Agreed. I'll first merge other exclusions up to `2.x`

cowtowncoder (approved):
LGTM

PR #1448, opened 2025-06-09
Upgrade `FastDoubleParser` to 2.0.1

pjfanning:
* not ready yet - let's wait till after #1447 is dealt with
* aim would be to update the versioned class excludes in #1447 to match what is in fastdoubleparser 2.0.1 (this jar has Java 23 classes but no Java 22 ones)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -175,6 +175,7 @@ com.fasterxml.jackson.core.*;version=${project.version}
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-shade-plugin</artifactId>
+        <version>3.6.0</version>
         <configuration>
           <promoteTransitiveDependencies>true</promoteTransitiveDependencies>
           <createDependencyReducedPom>true</createDependencyReducedPom>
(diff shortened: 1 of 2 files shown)


cowtowncoder:
It seems like we should probably only include LTS-matched classes, so leave JDK 21 but drop 23.

pjfanning:
This is currently blocked. maven-shade-plugin is omitting some new subpackages from FastDoubleParser. I have tried a few changes to the config setup but the classes are still missing so far.

pjfanning:
The issue seems that the `<minimizeJar>true</minimizeJar>` seems to lead to a lot of the FastDoubleParser classes being removed, presumably because we don't need them.

We could merge this and if anything breaks in other Jackson builds, we could revert this.

pom.xml
```diff
@@ -175,6 +175,7 @@ com.fasterxml.jackson.core.*;version=${project.version}
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-shade-plugin</artifactId>
+        <version>3.6.0</version>
```
cowtowncoder: Not needed, should come from "oss-parent", will remove.

cowtowncoder:
Sure, I can merge this now and we can see what (if anything) happens downstream.

PR #1465, opened 2025-08-18
Pre-size flattened parser sequences

schlosna:
- [ ] Needs benchmark & allocation profile


cowtowncoder:
Finding it bit unlikely that this should ever be a performance bottleneck for regular usage. So benchmark would def be useful.

schlosna:
> Finding it bit unlikely that this should ever be a performance bottleneck for regular usage. So benchmark would def be useful.


I'm fine with closing this out for now until I can devote more time to this.

This popped up in some production performance profiles from a serDe heavy service, but I need a bit more telemetry to create a realistic benchmark that doesn't arbitrarily skew the common cases and justify the complexity.

cowtowncoder:
@schlosna Sounds reasonable, would be interested in details if you ever find time to follow up.

PR #1474, opened 2025-09-16
Fixes the segment logic from splitting surrogates in half and not encoding correctly

### LINKED ISSUE #1473: Bug with Surrogates + Segments in the `UTF8JsonGenerator`
The UTF8JsonGenerator splits a string into segments without considering that it might cut the string exactly in between the high and low surrogate chars, which makes the generator escape surrogates instead of combining them when that feature is enabled. 

All cases where the segment is split must check if the final character is not the beginning of a surrogate (`_isStartOfSurrogatePair`) and adjust the segment `len` based on it (-1). 

https://github.com/FasterXML/jackson-core/blob/7ae2b8b9ea5d82c1b8d8ea543eb9e5577c0bff63/src/main/java/com/fasterxml/jackson/core/json/UTF8JsonGenerator.java#L1346

Does this make sense?

vitorpamplona:
Basically, the `combineSurrogates` logic doesn't run when the two characters of a surrogate are exactly at the end of one segment and the beginning of another.

Fixes https://github.com/FasterXML/jackson-core/issues/1473

### DIFF
--- a/src/main/java/com/fasterxml/jackson/core/json/UTF8JsonGenerator.java
+++ b/src/main/java/com/fasterxml/jackson/core/json/UTF8JsonGenerator.java
@@ -1375,7 +1375,7 @@ public class UTF8JsonGenerator
     {
         do {
             int len = Math.min(_outputMaxContiguous, totalLen);
-            // avoids splitting surrogates between two segments.
+            // [core#1473]: avoid splitting surrogates between two segments.
             // if len == 1 (edge case) don't apply to avoid infinite loop
             if (len > 1 && _isStartOfSurrogatePair(cbuf[offset + len-1])) {
                 --len;
@@ -1393,7 +1393,7 @@ public class UTF8JsonGenerator
     {
         do {
             int len = Math.min(_outputMaxContiguous, totalLen);
-            // avoids splitting surrogates between two segments.
+            // [core#1473]: avoid splitting surrogates between two segments.
             // if len == 1 (edge case) don't apply to avoid infinite loop
             if (len > 1 && _isStartOfSurrogatePair(text.charAt(offset + len-1))) {
                 --len;
@@ -1884,7 +1884,7 @@ public class UTF8JsonGenerator
     {
         do {
             int len = Math.min(_outputMaxContiguous, totalLen);
-            // avoids splitting surrogates between two segments.
+            // [core#1473]: avoid splitting surrogates between two segments.
             // if len == 1 (edge case) don't apply to avoid infinite loop
             if (len > 1 && _isStartOfSurrogatePair(utf8[offset + len-1])) {
                 --len;
--- a/src/main/java/com/fasterxml/jackson/core/json/UTF8JsonGenerator.java
+++ b/src/main/java/com/fasterxml/jackson/core/json/UTF8JsonGenerator.java
@@ -1346,7 +1346,7 @@ public class UTF8JsonGenerator
             int len = Math.min(_outputMaxContiguous, left);
             // [core#1473]: avoid splitting surrogates between two segments.
             // if len == 1 (edge case) don't apply to avoid infinite loop
-            if (len > 1 && _isStartOfSurrogatePair(text.charAt(len-1))) {
+            if (len > 1 && _isStartOfSurrogatePair(text.charAt(offset + len-1))) {
                 len--;
             }
             if ((_outputTail + len) > _outputEnd) { // caller must ensure enough space
@@ -1377,7 +1377,7 @@ public class UTF8JsonGenerator
             int len = Math.min(_outputMaxContiguous, totalLen);
             // avoids splitting surrogates between two segments.
             // if len == 1 (edge case) don't apply to avoid infinite loop
-            if (len > 1 && _isStartOfSurrogatePair(cbuf[len-1])) {
+            if (len > 1 && _isStartOfSurrogatePair(cbuf[offset + len-1])) {
                 len--;
             }
             if ((_outputTail + len) > _outputEnd) { // caller must ensure enough space
@@ -1395,7 +1395,7 @@ public class UTF8JsonGenerator
             int len = Math.min(_outputMaxContiguous, totalLen);
             // avoids splitting surrogates between two segments.
             // if len == 1 (edge case) don't apply to avoid infinite loop
-            if (len > 1 && _isStartOfSurrogatePair(text.charAt(len-1))) {
+            if (len > 1 && _isStartOfSurrogatePair(text.charAt(offset + len-1))) {
                 len--;
             }
             if ((_outputTail + len) > _outputEnd) { // caller must ensure enough space
@@ -1886,7 +1886,7 @@ public class UTF8JsonGenerator
             int len = Math.min(_outputMaxContiguous, totalLen);
             // avoids splitting surrogates between two segments.
             // if len == 1 (edge case) don't apply to avoid infinite loop
-            if (len > 1 && _isStartOfSurrogatePair(utf8[len-1])) {
+            if (len > 1 && _isStartOfSurrogatePair(utf8[offset + len-1])) {
                 len--;
             }
             _writeUTF8Segment(utf8, offset, len);
--- a/src/main/java/com/fasterxml/jackson/core/json/UTF8JsonGenerator.java
+++ b/src/main/java/com/fasterxml/jackson/core/json/UTF8JsonGenerator.java
@@ -1345,7 +1345,8 @@ public class UTF8JsonGenerator
         while (left > 0) {
             int len = Math.min(_outputMaxContiguous, left);
             // avoids splitting surrogates between two segments.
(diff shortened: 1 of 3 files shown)


src/test/java/com/fasterxml/jackson/core/write/UTF8GeneratorTest.java
```diff
@@ -105,6 +108,39 @@ void surrogatesWithRaw() throws Exception
...
+        JsonGenerator g = JSON_F.createGenerator(bb);
+        g.enable(JsonGenerator.Feature.COMBINE_UNICODE_SURROGATES_IN_UTF8);
+
+        g.writeStartArray();
+        g.writeString(VALUE);
+        g.writeEndArray();
+
+        g.close();
+
+        String result = new String(bb.toByteArray());
```
pjfanning: Could you specify StandardCharsets.UTF_8 so that we are not reliant on the default encoding (that can differ depending on OS, etc.?

src/main/java/com/fasterxml/jackson/core/json/UTF8JsonGenerator.java
```diff
@@ -1344,6 +1344,9 @@ private final void _writeStringSegments(String text, boolean addQuotes) throws I
 
         while (left > 0) {
             int len = Math.min(_outputMaxContiguous, left);
+            if (_isStartOfSurrogatePair(text.charAt(len-1))) {
```
cowtowncoder: Could you add a brief comment above, on why this is done (to avoid incorrect splitting of surrogate pairs so encoding works).

src/test/java/com/fasterxml/jackson/core/write/UTF8GeneratorTest.java
```diff
@@ -105,6 +109,39 @@ void surrogatesWithRaw() throws Exception
...
+        // We need a string with length 1001 where the surrogate is
+        // at 1000 and 1001 positions
+        int count = 999;
+        char[] chars = new char[count];
+        java.util.Arrays.fill(chars, 'x');
+        String base = new String(chars);
+
+        final String VALUE = base + "\uD83E\uDEE1";
+
+        // alas, we have to pull the recycler directly here...
```
cowtowncoder: Hmmh. Why are we using `BufferRecycler` (or `ByteArrayBuilder`) at all here?

cowtowncoder:
note: I think `src/main/java/com/fasterxml/jackson/core/json/WriterBasedJsonGenerator.java` wouldn't be affected as it's up to underlying `java.io.Writer` to handle escaping.

cowtowncoder:
Looks good overall, happy to get the fix!

One process thing: we'd need CLA (unless you have sent one earlier). It's here:

https://github.com/FasterXML/jackson/blob/main/contributor-agreement.pdf

and the usual way is to print, fill & sign, scan/photo, email to `cla` at fasterxml dot com.

Once I get it I can do final review, approve and merge PR.

I think this is also safe enough to go in `2.20` patch (I can cherry-pick the fix).

src/test/java/com/fasterxml/jackson/core/write/UTF8GeneratorTest.java
vitorpamplona: Added

src/test/java/com/fasterxml/jackson/core/write/UTF8GeneratorTest.java
vitorpamplona: Trying to simulate my own use case. I don't want a string at the end of this. I want a UTF-8 byte array.

src/main/java/com/fasterxml/jackson/core/json/UTF8JsonGenerator.java
vitorpamplona: added

vitorpamplona:
Just sent the CLA

src/test/java/com/fasterxml/jackson/core/write/UTF8GeneratorTest.java
cowtowncoder: Ah, then just use `ByteArrayOutputStream`? But either way let's not use `BufferRecycler` unless it's required part of reproduction; keep it (mostly) internal thing.

vitorpamplona: Cool, all done

src/test/java/com/fasterxml/jackson/core/write/UTF8GeneratorTest.java
```diff
@@ -105,6 +106,36 @@ void surrogatesWithRaw() throws Exception
         jp.close();
     }
 
+    @Test
```
cowtowncoder: Will move to test class that verifies #223 handling (where `COMBINE_UNICODE_SURROGATES_IN_UTF8` implemented).

src/main/java/com/fasterxml/jackson/core/json/UTF8JsonGenerator.java
```diff
@@ -1344,6 +1344,10 @@ private final void _writeStringSegments(String text, boolean addQuotes) throws I
 
         while (left > 0) {
             int len = Math.min(_outputMaxContiguous, left);
+            // [core#1473]: avoid splitting surrogates between two segments.
+            if (_isStartOfSurrogatePair(text.charAt(len-1))) {
+                len--;
```
cowtowncoder: Will need to consider case of `len == 1`, avoid infinite loop.
(extreme edge case, but possible I think)

cowtowncoder: Also: is missing `offset` -- will check wrong character if multiple segments actually written.

src/main/java/com/fasterxml/jackson/core/json/UTF8JsonGenerator.java
```diff
@@ -1370,6 +1374,10 @@ private final void _writeStringSegments(char[] cbuf, int offset, int totalLen) t
     {
         do {
             int len = Math.min(_outputMaxContiguous, totalLen);
+            // avoids splitting surrogates between two segments.
+            if (_isStartOfSurrogatePair(cbuf[len-1])) {
```
cowtowncoder: Same as above: check that `len > 1` and apply `offset.

src/main/java/com/fasterxml/jackson/core/json/UTF8JsonGenerator.java
```diff
@@ -1383,6 +1391,10 @@ private final void _writeStringSegments(String text, int offset, int totalLen) t
     {
         do {
             int len = Math.min(_outputMaxContiguous, totalLen);
+            // avoids splitting surrogates between two segments.
+            if (_isStartOfSurrogatePair(text.charAt(len-1))) {
```
cowtowncoder: same

src/main/java/com/fasterxml/jackson/core/json/UTF8JsonGenerator.java
```diff
@@ -1869,6 +1881,10 @@ private final void _writeUTF8Segments(byte[] utf8, int offset, int totalLen)
     {
         do {
             int len = Math.min(_outputMaxContiguous, totalLen);
+            // avoids splitting surrogates between two segments.
+            if (_isStartOfSurrogatePair(utf8[len-1])) {
```
cowtowncoder: and same

cowtowncoder (changes requested):
I think there's a bug (not using `offset`) and infinite-loop edge case (`len == 1`) to address.

src/main/java/com/fasterxml/jackson/core/json/UTF8JsonGenerator.java
vitorpamplona: Good catch on the len == 1. I can't see the issue with the offset. It shouldn't be affected by this. What do you mean by "apply"?

cowtowncoder: @vitorpamplona Check needs to be something like:

    _isStartOfSurrogatePair(text.charAt(offset + len - 1)))

since we have the call:

     _writeStringSegment(text, offset, len);

so, writes are for segments like `[0-999], [1000-1999]`.

Ideally test would try actual segmented case.

So basically with initial offset of `0` code works, but if there's more content to write, second check is for wrong character.

vitorpamplona: ohhh got it.

cowtowncoder (approved):
LGTM

cowtowncoder:
Looks good, hoping to sanity check later, merge.

cowtowncoder (approved):

cowtowncoder:
Merged into 2.x, backported in 2.20 (for 2.20.1).

Thank you @vitorpamplona ! This is very useful fix (even if not common bug).

PR #1488, opened 2025-10-29
Potential fix to DataInput-backed parser, number parsing

cowtowncoder:
Created by Claude, needs actual tests (possibly extend/uncomment existing ones?)


cowtowncoder:
I don't think this makes sense, in the end -- `DataInput` has different handling of End-of-Input, should not hide (and in particular only in some places). Closing.

Issue #363, opened 2017-03-22
Check why UTF-8 decoding does not fail on illegal encodings of Surrogate codes


cowtowncoder:
See https://github.com/FasterXML/jackson-dataformats-binary/pull/65 for background: it looks like UTF-8 json parser is not checking for case of 3-byte character being surrogate: I think it should.
(although there is still the whole issue of why JSON escapes may be used in such a way...).

cowtowncoder:
Fixed for 2.21.0, 3.0.2

PR #1494, opened 2025-11-02
Add failing tests for, fix issue #363

cowtowncoder:


AfzalivE:
@cowtowncoder Since this change, the json is failing to parse with `StreamRead Invalid UTF-8: Illegal surrogate character 0xd83d`. Is this expected?

```java
      private final JsonFactory FACTORY = newStreamFactory();

      /**
       * This test fails with:
       * StreamReadException: Invalid UTF-8: Illegal surrogate character 0xd83d
       */
      @Test
      void acceptJsonEscapedSurrogatePairThumbsUp() throws Exception
      {
          // JSON with escaped surrogate pair: {"\ud83d\udc4d": "value"}
          // \ud83d\udc4d is the JSON representation of the thumbs up emoji
          String json = "{\"\\ud83d\\udc4d\": \"value\"}";
          byte[] doc = json.getBytes("UTF-8");

          try (JsonParser p = FACTORY.createParser(ObjectReadContext.empty(), doc)) {
              assertToken(JsonToken.START_OBJECT, p.nextToken());
              assertToken(JsonToken.PROPERTY_NAME, p.nextToken());
              // The escaped surrogate pair should decode to the thumbs up emoji
              assertEquals("\uD83D\uDC4D", p.currentName());
              assertToken(JsonToken.VALUE_STRING, p.nextToken());
              assertEquals("value", p.getString());
              assertToken(JsonToken.END_OBJECT, p.nextToken());
          }
      }
```

cowtowncoder:
@AfzalivE I think your test might be broken due to use of `byte[] doc = json.getBytes("UTF-8");` -- it encodes Surrogate characters as UTF-8 code points which are not legal, as they should only be used in UTF-16 (and UCS-2 variation) content. UTF-8 can express underlying code (https://unicodeplus.com/U+1F44D) using 4-byte sequence.

From the test:

```
/**
 * Tests for [jackson-core#363]: UTF-8 parser should reject 3-byte UTF-8 sequences
 * that encode surrogate code points (U+D800 to U+DFFF), which are illegal in UTF-8.
 */
```

so, yes, this is intentional: UTF-8 content being fed is invalid: surrogate characters are not valid in UTF-8 encoded content.

AfzalivE:
@cowtowncoder I appreciate the reply.

In the minimum repro unit tests below, `acceptJsonEscapedSurrogatePairInFieldName` is failing but `acceptJsonEscapedSurrogatePairInStringValue` passes. I don't know nearly enough about unicode but seems like either both should fail or both should pass (as evidenced by similar tests in `UTF8SurrogateValidation363Test`). Seems like the field name codepath is doing something different.

```java
    @Test
    void acceptJsonEscapedSurrogatePairInFieldName() throws Exception
    {
        // JSON: {"\ud83d\udc4d":"value"}
        byte[] doc = new byte[] {
            '{', '"',
            '\\', 'u', 'd', '8', '3', 'd',  // JSON escape: \ud83d (high surrogate)
            '\\', 'u', 'd', 'c', '4', 'd',  // JSON escape: \udc4d (low surrogate)
            '"', ':', '"', 'v', 'a', 'l', 'u', 'e', '"',
            '}'
        };

        try (JsonParser p = FACTORY.createParser(doc)) {
            assertToken(JsonToken.START_OBJECT, p.nextToken());
            assertToken(JsonToken.FIELD_NAME, p.nextToken());
            // The escaped surrogate pair should decode to U+1F44D (thumbs up emoji)
            assertEquals("\uD83D\uDC4D", p.currentName());
            assertToken(JsonToken.VALUE_STRING, p.nextToken());
            assertEquals("value", p.getText());
            assertToken(JsonToken.END_OBJECT, p.nextToken());
        }
    }

    /**
     * Test that JSON escape sequence \ud83d\udc4d in string value is accepted.
     *
     * JSON: {"key":"\ud83d\udc4d"}
     */
    @Test
    void acceptJsonEscapedSurrogatePairInStringValue() throws Exception
    {
        // JSON: {"key":"\ud83d\udc4d"}
        byte[] doc = new byte[] {
            '{', '"', 'k', 'e', 'y', '"', ':', '"',
            '\\', 'u', 'd', '8', '3', 'd',  // JSON escape: \ud83d (high surrogate)
            '\\', 'u', 'd', 'c', '4', 'd',  // JSON escape: \udc4d (low surrogate)
            '"',
            '}'
        };

        try (JsonParser p = FACTORY.createParser(doc)) {
            assertToken(JsonToken.START_OBJECT, p.nextToken());
            assertToken(JsonToken.FIELD_NAME, p.nextToken());
            assertEquals("key", p.currentName());
            assertToken(JsonToken.VALUE_STRING, p.nextToken());
            // The escaped surrogate pair should decode to U+1F44D (thumbs up emoji)
            assertEquals("\uD83D\uDC4D", p.getText());
            assertToken(JsonToken.END_OBJECT, p.nextToken());
        }
    }
```

Here's what Claude is saying about this fwiw:

> When parsing field names, the code at lines 2025-2045 re-encodes the decoded escape sequence value (e.g., `0xD83D` from `\ud83d`) as a 3-byte UTF-8 sequence (`0xED 0xA0 0xBD)` into the quads buffer, which later gets rejected by `addName()` as an illegal surrogate—whereas string value parsing avoids this entirely by storing the decoded value directly into a `char[]` buffer where Java natively handles surrogate pairs.

cowtowncoder:
@AfzalivE could you please file a new, follow-up issue with above info? (with ref to this issue as background).

It sounds like there may be something wrong after all. Field name decoding is indeed quite specialized, for performance reasons -- so error handling may be incorrect.
Use of Unicode escape should allow use of surrogate codes, as per JSON spec.

cowtowncoder:
In hindsight, this validation should have been opt-in (or, maybe, opt-out) gated by new `StreamReadFeature`. But too late to do now.

Issue #679, opened 2021-02-06
Number parsing should fail for trailing dot (period)


cowtowncoder:
Looks like content such as:

```
[ 100.300.30 ]
```

passes and reports number token `100.300`, only failing when trying to access to the following token.
This should be caught earlier.

cowtowncoder:
Looks like this affects non-root numeric values -- at root level (where it is not that valuable, alas) existence of separator is checked as expected.
Should also check in other contexts but things get much more ... interesting... since there are legitimately more potentially valid characters compared to root level (where only white-space is allowed as separators)

TheSasch:
Why do you think that "100k" should throw an Exception? Wouldn't it be better that it gets a String result instead?

cowtowncoder:
If you mean "document" like:

```
100k
```

that should fail not being valid JSON value (and by extension neither JSON document).
The default behavior Jackson aims for is to only accept well-formed valid JSON content.

There is then a separate question of optionally allowing deviations; a few cases exist (enabled by `JsonReadFeature`).
In those cases it is also important to define exactly what else would be accepted.
You may file an issue to request acceptance for specific additional classes if they are something you would like to see.

TheSasch:
Oh thank you. I did also had a look at the RFC for JSON and didn't know that a String must always have " surrounded. I thought I've seen JSON in the past without it. Like in YML.

I had a look at the Unit Tests you wrote and also found the first possible problem within UTF8StreamJsonProcessor. At the End of _ParsePosNumber you don't stop when there is another char than a Number. (After the if statement where you know that an "e", "E" or "." would get you to a floating number.

One possible solution would be to show if the current char is a whitespace and continue then. If it's not then there must be an error. But this is only a first look in the code and for now I don't know what standard you have to throw exceptions. And also the other positions are still left where that could be a problem.

cowtowncoder:
Note: #105 somewhat related.

Issue #105, opened 2013-09-21
Parser parsers numbers eagerly; does not report error with missing space


cowtowncoder:
(note: moved from https://github.com/FasterXML/jackson-databind/issues/260)

---

If you give input like:

```
123true
```

it will first successfully parse integer 123, and then boolean `true`. This should instead result in a parse exception.

PR #1510, opened 2025-11-17
Possible fix for #679

cowtowncoder:
Possible fix for and #679 (#105 was fixed concurrently) -- need to see how intrusive this would be (and what performance consequences)...


cowtowncoder:
Not happy with implementation, let's retry in future.

Issue #1527, opened 2025-12-12
`TokenStreamFactory.createNonBlockingByteBufferParser()` return type wrong


hb16:
Both createNonBlockingByteArrayParser() and createNonBlockingByteBufferParser() in TokenStreamFactory have the same type bound for the return type:

`P extends JsonParser & ByteArrayFeeder`

I'd expect the type for createNonBlockingByteBufferParser() to be

`P extends JsonParser & ByteBufferFeeder`.

Is this a copy-paste error or am I missing something?

pjfanning:
@hb16 the method signature is wrong - can you use JsonFactory for now? JsonFactory has the method and you can cast its return type to ByteBufferFeeder. We will try to fix the return types in an upcoming release.

cowtowncoder:
Definitely copy-paste fail. Embarrassing.

cowtowncoder:
Fixed by #1528

pjfanning:
You'd really have expected that the Java parser would have prevented this.

cowtowncoder:
@pjfanning I think part of the problem was override in `JsonFactory` that used too generic return type, instead of indicating mismatch with `TokenStreamFactory` base typing.

PR #1528, opened 2025-12-12
adjust return types on non-blocking parsers

pjfanning:
relates to #1527 - one possible solution

### DIFF
--- a/release-notes/CREDITS
+++ b/release-notes/CREDITS
@@ -17,3 +17,8 @@ Sven Döring (@sdoeringNew)
 Nicholas Rayburn (@nrayburn-tech)
  * Contributed #1514: Additional configuration to closer match Jackson 2 behavior
   (3.1.0)
+
+Henri Biström (@hb16)
+ * Reported #1527: `TokenStreamFactory.createNonBlockingByteBufferParser()` return type wrong
+  (3.1.0)
+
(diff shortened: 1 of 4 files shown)


cowtowncoder:
Tricky... Ok first things first.

Fix in `TokenStreamFactory` needs to go in 3.x (for 3.1), even if it was (not sure it is?) not binary compatible -- although maybe it is binary compatible. But that needs to be fixed regardless.

`JsonFactory` is trickier. I think this would be source-compatible but binary incompatible (since type-erased return type is part of method signature). Those are nasty. So while improvement, I am not sure if fix could be worse than the problem.

Could you split this in 2 part: I'll merge 1st part immediately and we can consider 2nd separately?

**EDIT**: assuming this is possible to do (2nd part independent)

pjfanning:
JsonFactory change reverted.

cowtowncoder:
> ## 📈 Overall Code Coverage
> Metric 	Coverage 	Change
> **Instructions** 	coverage **82.80%** 	📈 **+5.541%**
> **Branches** 	branches **75.17%** 	📈 **+0.691%**
> 
> > Overall project coverage from JaCoCo test results. Change values compare against the latest base branch build.

Not sure why Code Coverage delta so wrong :-(

Issue #1288, opened 2024-05-06
Add new method like `JsonParser.readText(Writer)` (and implementation) for truly non-buffering reads


cowtowncoder:
(note: off-shoot of https://github.com/FasterXML/jackson/discussions/221)

Currently there is no way to access contents of long JSON String values without buffering of all contents in memory: even metho

    JsonParser.getText(Writer)

will temporarily buffer contents in `TextBuffer` (but avoids construction of big `String` unlike `getText()`).
But it would be possible and desirable to have an alternative that would decode and provide content without buffering; similar to how Stax2-compliant XML parsers (Aalto, Woodstox) do it.

Semantic difference would be that such method, f.ex

    JsonParser.readText(Writer)

would read, pass and _consume_ contents, such that after call finishes, parser now points to the end of text value, and text value could no longer be read.
Because of this change, we cannot change implementation of existing `getText(Writer)`

I will also update Javadocs of `getText(Writer)` to indicate use of buffering.

cowtowncoder:
Merged impl in 3.1.0 -- thank you @sri-adarsh-kumar !

PR #1543, opened 2026-02-09
Add `JsonParser.readString(Writer)` for truly non-buffering string reads

sri-adarsh-kumar:
Implements #1288 (originally #15) to support streaming large JSON string values without buffering entire content in memory.

### DIFF
--- a/src/main/java/tools/jackson/core/JsonParser.java
+++ b/src/main/java/tools/jackson/core/JsonParser.java
@@ -918,8 +918,10 @@ public abstract class JsonParser
      *
      * @throws JacksonIOException for low-level read issues, or failed write using {@link Writer}
      * @throws tools.jackson.core.exc.StreamReadException for decoding problems
+     *
+     * @since 3.1
      */
-    public abstract int readText(Writer writer) throws JacksonException;
+    public abstract long readString(Writer writer) throws JacksonException;
 
     /**
      * Method similar to {@link #getString()}, but that will return
--- a/src/main/java/tools/jackson/core/JsonParser.java
+++ b/src/main/java/tools/jackson/core/JsonParser.java
@@ -897,6 +897,31 @@ public abstract class JsonParser
     public abstract int getString(Writer writer) throws JacksonException;
 
     /**
+     * Method to read the textual representation of the current token in chunks and
+     * pass it to the given Writer, without buffering the whole String value.
+     * Functionally same as calling:
+     *<pre>
+     *  writer.write(parser.getString());
+     *</pre>
+     * but allows implementations to stream the decoded content directly without
+     * storing it in {@link TextBuffer}.
+     *<p>
+     * NOTE: Unlike {@link #getString(Writer)}, this method <b>consumes</b> the
+     * contents of a {@link JsonToken#VALUE_STRING} token, advancing the parser
+     * so that the underlying String value is no longer available via other
+     * {@code getString*} accessors. This method is primarily intended for very
+     * large String values where buffering would be prohibitive.
+     *
+     * @param writer Writer to write String value to
+     *
+     * @return The number of characters written to the Writer
+     *
+     * @throws JacksonIOException for low-level read issues, or failed write using {@link Writer}
+     * @throws tools.jackson.core.exc.StreamReadException for decoding problems
+     */
+    public abstract int readText(Writer writer) throws JacksonException;
+
+    /**
      * Method similar to {@link #getString()}, but that will return
      * underlying (unmodifiable) character array that contains
      * textual value, instead of constructing a String object
--- a/src/main/java/tools/jackson/core/JsonParser.java
+++ b/src/main/java/tools/jackson/core/JsonParser.java
@@ -897,27 +897,45 @@ public abstract class JsonParser
     public abstract int getString(Writer writer) throws JacksonException;
 
     /**
-     * Method to read the textual representation of the current token in chunks and
-     * pass it to the given Writer, without buffering the whole String value.
-     * Functionally same as calling:
+     * Method to read the textual representation of the current {@link JsonToken#VALUE_STRING}
+     * token in chunks and stream it directly to the given Writer, without buffering the entire
+     * String value in memory. Functionally same as calling:
      *<pre>
      *  writer.write(parser.getString());
      *</pre>
-     * but allows implementations to stream the decoded content directly without
-     * storing it in {@link TextBuffer}.
+     * but streams the decoded content directly without storing it in {@link TextBuffer},
+     * making it suitable for arbitrarily large strings without memory constraints.
      *<p>
-     * NOTE: Unlike {@link #getString(Writer)}, this method <b>consumes</b> the
-     * contents of a {@link JsonToken#VALUE_STRING} token, advancing the parser
-     * so that the underlying String value is no longer available via other
-     * {@code getString*} accessors. This method is primarily intended for very
-     * large String values where buffering would be prohibitive.
+     * NOTE: This method <b>consumes</b> the contents of the {@link JsonToken#VALUE_STRING}
+     * token, advancing the parser state so that the underlying String value is <b>no longer
+     * available</b> via {@link #getString()} or other {@code getString*} accessors after
+     * this method completes. This differs from {@link #getString(Writer)} which preserves
+     * the buffered string for subsequent access.
+     *<p>
+     * NOTE: This method is primarily intended for very large JSON string values (megabytes
+     * or larger) where full buffering would be prohibitive. For typical string sizes, prefer
+     * {@link #getString()} or {@link #getString(Writer)} which provide more convenient access.
+     * The implementation uses a 512-character intermediate buffer for efficient bulk writes
+     * to the Writer.
+     *<p>
+     * NOTE: This method <b>does</b> enforce
+     * {@link tools.jackson.core.StreamReadConstraints#maxStringLength()} validation during
+     * streaming, checking the string length at buffer boundaries and at completion. Strings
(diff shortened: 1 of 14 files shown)


cowtowncoder:
First of all: thank you for contributing this. I'll need to review it carefully.

But one big question on this:

> this method bypasses TextBuffer entirely and StreamReadConstraints validation, enabling processing of strings larger than maxStringLength limit.

I am not sure this is a good idea -- should constraints really be ignored by this method? Is that a feature not bug?

/cc @pjfanning @JooHyukKim

src/main/java/tools/jackson/core/JsonParser.java
```diff
@@ -896,6 +896,31 @@ public Boolean nextBooleanValue() throws JacksonException {
...
+     * large String values where buffering would be prohibitive.
+     *
+     * @param writer Writer to write String value to
+     *
+     * @return The number of characters written to the Writer
+     *
+     * @throws JacksonIOException for low-level read issues, or failed write using {@link Writer}
+     * @throws tools.jackson.core.exc.StreamReadException for decoding problems
+     */
+    public abstract int readText(Writer writer) throws JacksonException;
```
cowtowncoder: New naming: should be "readString()" (as per 3.0 style).

Also: should return `long` not `int` to support truly long content.

cowtowncoder: Also needs

    @since 3.1

src/main/java/tools/jackson/core/json/async/NonBlockingJsonParserBase.java
```diff
@@ -437,6 +437,12 @@ public int getString(Writer writer) throws JacksonException
         return 0;
     }
 
+    @Override
+    public int readText(Writer writer) throws JacksonException
+    {
+        return getString(writer);
```
cowtowncoder: Cannot really be supported reliably for non-blocking parser -- should this instead throw `UnsupportedOperationException`? Because sometimes it could work (enough content fed), other times not (not yet complete String)?

Or should it return -1 as marker for "Not yet fully available"?

NOTE: existing `getString(Writer)` implementation looks wrong, unfortunately. Should also similarly either fail, or return marker.

JooHyukKim:
StreamReadConstraints is the almost-only safe way to have some sort of control gate over what inputs applications would receive.

I think we should not bypass.

src/main/java/tools/jackson/core/json/async/NonBlockingJsonParserBase.java
JooHyukKim: I agree

src/main/java/tools/jackson/core/JsonParser.java
sri-adarsh-kumar: Changed as per suggestion.

src/main/java/tools/jackson/core/json/async/NonBlockingJsonParserBase.java
sri-adarsh-kumar: Changed to throw UnsupportedOperationException

sri-adarsh-kumar:
> First of all: thank you for contributing this. I'll need to review it carefully.
> 
> But one big question on this:
> 
> > this method bypasses TextBuffer entirely and StreamReadConstraints validation, enabling processing of strings larger than maxStringLength limit.
> 
> I am not sure this is a good idea -- should constraints really be ignored by this method? Is that a feature not bug?
> 
> /cc @pjfanning @JooHyukKim

Thanks for the comments, I have added checks on maxStringLength from StreamReadConstraints

pjfanning:
> StreamReadConstraints is the almost-only safe way to have some sort of control gate over what inputs applications would receive.
> 
> I think we should not bypass.

I agree. Constraints should be applied. Users can override the constraint defaults.

src/main/java/tools/jackson/core/base/ParserMinimalBase.java
```diff
@@ -547,6 +547,12 @@ public int getString(Writer writer) throws JacksonException
         return str.length();
     }
 
+    @Override
+    public long readString(Writer writer) throws JacksonException
```
pjfanning: Can we just call this  'readFromWriter'? 'readString' is a very strange name when input is not a String.

cowtowncoder: Hmmmh. I think "readString()" is consistent with "getText()", accessing (mostly?) JSON String tokens.
"readFromWriter()" is... bit vague.
While I can see confusion I prefer `readString()` here.

src/main/java/tools/jackson/core/json/ReaderBasedJsonParser.java
```diff
@@ -2268,6 +2301,55 @@ protected final void _skipString() throws JacksonException
...
+                    }
+                    if (i < INT_SPACE) {
+                        _inputPtr = inPtr;
+                        _throwUnquotedSpace(i, "string value");
+                    }
+                }
+            }
+            writer.write(c);
+            if (++count > maxStringLen) {
+                _streamReadConstraints.validateStringLength((int) count);
```
pjfanning: Can you check if count > max int?

src/main/java/tools/jackson/core/json/UTF8StreamJsonParser.java
```diff
@@ -3108,6 +3140,88 @@ protected void _skipString() throws JacksonException
...
+                }
+                while (ptr < max) {
+                    c = inputBuffer[ptr++] & 0xFF;
+                    if (codes[c] != 0) {
+                        _inputPtr = ptr;
+                        break ascii_loop;
+                    }
+                    writer.write((char) c);
+                    if (++count > maxStringLen) {
+                        _streamReadConstraints.validateStringLength((int) count);
```
pjfanning: Same here, can you check that count is not more than max int?

src/main/java/tools/jackson/core/json/ReaderBasedJsonParser.java
cowtowncoder: UGGH. This is insanely inefficient; definitely not the way to do it.

Instead should only check at boundary, when reading more content.

src/main/java/tools/jackson/core/json/ReaderBasedJsonParser.java
```diff
@@ -2268,6 +2301,55 @@ protected final void _skipString() throws JacksonException
...
+        final int maxStringLen = _streamReadConstraints.getMaxStringLength();
+
+        while (true) {
+            if (inPtr >= inLen) {
+                _inputPtr = inPtr;
+                if (!_loadMore()) {
+                    _reportInvalidEOF(": was expecting closing quote for a string value",
+                            JsonToken.VALUE_STRING);
+                }
+                inPtr = _inputPtr;
```
cowtowncoder, line 2333: Bounds checks should only be done here... not for every char.

src/main/java/tools/jackson/core/json/ReaderBasedJsonParser.java
```diff
@@ -2268,6 +2301,55 @@ protected final void _skipString() throws JacksonException
...
+                        _inputPtr = inPtr;
+                        break;
+                    }
+                    if (i < INT_SPACE) {
+                        _inputPtr = inPtr;
+                        _throwUnquotedSpace(i, "string value");
+                    }
+                }
+            }
+            writer.write(c);
```
cowtowncoder: Ummm, no, write call for every character wasteful; should read into intermediate buffer, bulk write from that.

sri-adarsh-kumar: Used an intermediate buffer as per suggestion

src/main/java/tools/jackson/core/json/ReaderBasedJsonParser.java
sri-adarsh-kumar: Implemented this

src/main/java/tools/jackson/core/json/UTF8StreamJsonParser.java
sri-adarsh-kumar: Added check

src/main/java/tools/jackson/core/json/ReaderBasedJsonParser.java
sri-adarsh-kumar: Added check at boundary

src/main/java/tools/jackson/core/JsonParser.java
```diff
@@ -896,6 +896,51 @@ public Boolean nextBooleanValue() throws JacksonException {
...
+     *         strings exceeding {@code Integer.MAX_VALUE})
+     *
+     * @throws JacksonIOException for low-level read issues, or failed write using {@link Writer}
+     * @throws tools.jackson.core.exc.StreamReadException for decoding problems
+     * @throws tools.jackson.core.exc.StreamConstraintsException if string length exceeds
+     *         {@link tools.jackson.core.StreamReadConstraints#maxStringLength()}
+     *
+     * @since 3.1
+     */
+    public abstract long readString(Writer writer) throws JacksonException;
```
cowtowncoder: Just realized that due to backwards-compatibility reasons, we can NOT leave this `abstract` -- will demote implementation from `ParserMinimalBase` here.

cowtowncoder:
@sri-adarsh-kumar Good job with the PR! I think I'm happy with it now, with some extended, Claude-generated tests & bit of refactoring.

cowtowncoder (approved):

PR #1550, opened 2026-02-19
Add test for nesting for `DataInput`-backed `JsonParser`

pjfanning:
I tried this against 2.18 but ran into issues with missing snapshots in central.sonatype.com.

We have already discussed the 3.x version of this which turns out to be different. This test passes as is in 2.21.

### DIFF
--- a/src/test/java/com/fasterxml/jackson/core/constraints/NestingDepthBypassDataInputTest.java
+++ b/src/test/java/com/fasterxml/jackson/core/constraints/NestingDepthBypassDataInputTest.java
@@ -6,7 +6,6 @@ import org.junit.jupiter.api.Test;
 
 import com.fasterxml.jackson.core.*;
 import com.fasterxml.jackson.core.exc.StreamConstraintsException;
-import com.fasterxml.jackson.core.json.JsonFactory;
 
 import static org.junit.jupiter.api.Assertions.*;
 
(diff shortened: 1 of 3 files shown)


cowtowncoder:
@pjfanning Ok, Sonatype is pruning SNAPSHOT versions rather aggressively; CI publishes these but only when there are changes.

But FWTW, what I do when this happen is clone `jackson-bom`, check out 2.18 (or whichever is needed) and do:

    ./mvnw clean install deploy

to re-publish latest SNAPSHOT.

So now 2.18 should build again (it worked locally as I had cached snapshots locally)

src/test/java/com/fasterxml/jackson/core/constraints/NestingDepthBypassDataInputTest.java
```diff
@@ -0,0 +1,46 @@
...
+        try (JsonParser p = factory.createParser(di)) {
+            int maxDepth = 0;
+            while (p.nextToken() != null) {
+                if (p.currentToken() == JsonToken.START_ARRAY) {
+                    maxDepth++;
+                }
+            }
+            fail("DataInput parser must reject nesting depth " + TEST_NESTING_DEPTH);
+        } catch (StreamConstraintsException e) {
+            //expected
```
cowtowncoder: Maybe verify part of wording to ensure its depth-based constraint? (just in case)

pjfanning: updated test

cowtowncoder (approved):

PR #1551, opened 2026-02-19
enable DeeplyNestedContentReadTest

pjfanning:
test missing `@Test` annotation


cowtowncoder:
Interesting. Must have occurred during JUnit 4 -> 5 conversion? But even if so, why is it ok in 2.x...

cowtowncoder (approved):

PR #1552, opened 2026-02-19
Re-enable failing (only) in 3.x DataInput test

cowtowncoder:


cowtowncoder:
@pjfanning Your test reproduces the issue in 3.x, as per this PR.

Odd things is; I added another `DataInput`-backed parser test in `DeeplyNestedContentReadTest` which does not fail. :-/

cowtowncoder:
Fixed via #1554, no longer needed.

PR #1555, opened 2026-02-21
Enforce `StreamReadConstraints.maxNumberLength` for non-blocking (async) parser

pjfanning:
validate number length against constraints in async parser


cowtowncoder (approved):
LGTM, will merge first thing tomorrow.

cebarks:
Hi there @cowtowncoder @pjfanning,
 
I work in Product Security over at Red Hat. We had a reporter reach out to us concerned that this should be considered a security vulnerability. If you don't mind, I was going to assign a CVE id to track this. Is this alright with you?

If so, let me know if anyone would like credit for the discovery and/or if @pjfanning want's credit for the fix.

Thanks!

pjfanning:
@cebarks we have https://github.com/FasterXML/jackson-core/security/advisories/GHSA-72hv-8253-57qq

We don't view this as CVE worthy. The effect of numbers with multi thousand digits is non-linear CPU time parsing the number. We use Java Runtime to do this parsing. If Jackson has an issue, then so does the Java Runtime.

cebarks:
Gotcha, totally reasonable. I'll defer to your judgement and skip assigning a CVE.

Thanks for your quick response!

Issue #1576, opened 2026-03-12
Use Java 17 intrinstics for long multiplication


xtonik:
Hi folks,

I have noticed, that new version of Java version 17 was adopted by Jackson, which offers several performance improvements:

1. `Long.multiplyHigh(long, long)` was intrinsified by [JDK-8187964](https://bugs.java.com/bugdatabase/JDK-8187964) and should be preferred over custom implementation, it is present in [MathUtils](https://github.com/FasterXML/jackson-core/blob/c6f47e4d748d5c38efaac4fdf639418826366f4c/src/main/java/tools/jackson/core/io/schubfach/MathUtils.java#L189) now.
 
2. Having intrinsified method `Long.multiplyHigh(long, long)`, it could be used for more performant division by multiply and shift (yes, my evergreen as #1203), according to my benchmarks results, long number divided by 10^9 should be 30% faster via multiply and shift, the code:
```
private static final int MAGIC_SHIFT = 28;
private static final long MAGIC_MULT = BigInteger.ONE.shiftLeft(Long.SIZE + MAGIC_SHIFT).divide(BigInteger.TEN.pow(9)).longValue() + 1; // ~ 2^(64+28) / 10^9
public static long divBy10to9(long n) {
    return Math.multiplyHigh(n, MAGIC_MULT) >>> 28;
}
```
A example of correctness verification:
```
long max = Long.MAX_VALUE / 1_000_000_000L + 1;
for (long i = 1; i < max; i++) {
    long n = 1_000_000_000L * i;
    long div1 = divBy10to9(n - 1);
    long div2 = divBy10to9(n);
    if (div1 != i - 1 || div2 != i) {
       throw new IllegalStateException(i + " " + div1 + " " + div2);
    }
}
if (divBy10to9(Long.MAX_VALUE) != Long.MAX_VALUE / 1_000_000_000L) {
      throw new IllegalStateException(Long.MAX_VALUE);
}
```
3. Base64 performance could be improved by standard JDK implementation, because:
a) It is faster than yours implementation according to #182.
b) This job could be done faster by custom Base64, I have achieved much better results with my solution (not published anywhere), even without use Unsafe or VarHandle.
[...]

PR #1577, opened 2026-03-17
Use intrinsified Math.multiplyHigh()

xtonik:
Prefer use of intrinsified method `Math.multiplyHigh()` over custom implementation as reported in #1576. Even built-in JDK method is for signed values and thus signed result, it works well for positive value also. Possible fallback to not intrinsified JDK version does not affect negatively performance as demonstrated in this benchmark (not provided) results:

  | (multiplier) | Mode | Cnt | Score | Error | Units
-- | -- | -- | -- | -- | -- | --
MultiplyHighBenchmark.multiplyHigh | SIGNED_INTRINSIFIED | thrpt | 10 | 0.991 | ± 0.443 | ops/us
MultiplyHighBenchmark.multiplyHigh | SIGNED | thrpt | 10 | 0.386 | ± 0.014 | ops/us
MultiplyHighBenchmark.multiplyHigh | UNSIGNED_JACKSON | thrpt | 10 | 0.388 | ± 0.012  | ops/us
MultiplyHighBenchmark.multiplyHigh | UNSIGNED_INTRINSIFIED  |  thrpt   |  10    |  0.999 | ±  0.048  |  ops/us

### DIFF
--- a/release-notes/CREDITS
+++ b/release-notes/CREDITS
@@ -40,3 +40,7 @@ Rohan Nagendra (@rohan-repos)
  * Reported #1553: Max Depth validation not working for `DataInput`-backed `JsonParser`s
    in 3.0
   (3.1.0)
+
+Antonin Janec (@xtonik)
+ * Contributed #1576: Use Java 17 intrinstics for long multiplication
+  (3.2.0)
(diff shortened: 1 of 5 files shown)


cowtowncoder (approved):
LGTM

PR #1580, opened 2026-03-24
Add `JsonParser.willInternPropertyNames()` for property name intern introspection (#1211)

### LINKED ISSUE #1211: Ability to determine whether field name interning is enabled for JsonParser
I have custom deserialization logic of protocol buffers and I took advantage of field names being interned to improve performance.

https://github.com/curioswitch/protobuf-jackson/blob/main/src/main/java/org/curioswitch/common/protobuf/json/ParseSupport.java#L458

Actually, at the time I didn't know it is possible to disable interning. I was able to create a failing test for it

https://github.com/curioswitch/protobuf-jackson/blob/main/src/test/java/org/curioswitch/common/protobuf/json/JacksonInteropTest.java#L20

It would be great to be able to know if the parser is interning to be able to handle this case correctly while also skipping non-reference equality check when it's fine to do so. Currently `JsonParser` only exposes parser features:

https://fasterxml.github.io/jackson-core/javadoc/2.8/com/fasterxml/jackson/core/JsonParser.html#getFeatureMask()

But there is no access to the factory features. Naturally `JsonParser` is not the right level for that but I wonder would it be possible to expose factory features from `ParserBase`? Or it could just be the three standard parser implementations individually exposing it. If we could check the factory features, or more specifically whether the symbol finder is interning strings, then it would be possible to skip `.equals` for name comparisons.

I looked at databind, and while I'm not completely sure I understand the code, I think this optimization could be applied there too

https://github.com/FasterXML/jackson-databind/blob/2.17/src/main/java/com/fasterxml/jackson/databind/deser/impl/BeanPropertyMap.java#L575

The `.equals` could be skipped when it's known the fields are interned.

maxpaulus43:
Add ability to determine whether field name interning is enabled for a `JsonParser` instance, as suggested in [GitHub issue #1211](https://github.com/FasterXML/jackson-core/issues/1211).

## Changes

- Add `JsonParser.willInternPropertyNames()` with default returning `false`
- Add `isInternStrings()` accessor to `ByteQuadsCanonicalizer` and `CharsToNameCanonicalizer` symbol tables
- Override in `UTF8StreamJsonParser`, `ReaderBasedJsonParser`, `UTF8DataInputJsonParser`, and `NonBlockingJsonParserBase` to delegate to their respective symbol tables
- Override in `JsonParserDelegate` to delegate to wrapped parser
- Add comprehensive unit tests (`InternFieldNamesTest`) covering all parser types with various feature configurations
- Update `DelegatesTest` to verify delegation of new method
- Add release notes entry under 3.2.0

Closes #1211

### DIFF
--- a/release-notes/CREDITS
+++ b/release-notes/CREDITS
@@ -41,6 +41,11 @@ Rohan Nagendra (@rohan-repos)
    in 3.0
   (3.1.0)
 
+Max Paulus (@maxpaulus43)
+ * Contributed #1211: Add `JsonParser.willInternPropertyNames()` to check whether
+   property name interning is enabled
+  (3.2.0)
+
 Antonin Janec (@xtonik)
  * Contributed #1576: Use Java 17 intrinstics for long multiplication
   (3.2.0)
--- a/release-notes/VERSION
+++ b/release-notes/VERSION
@@ -28,6 +28,8 @@ JSON library.
  (implemented by @pjfanning)
 #1576: Use Java 17 intrinstics for long multiplication
  (contributed by @xtonik)
+#1211: Add `JsonParser.willInternPropertyNames()` to check whether
+  property name interning is enabled
 
 3.1.1 (not yet released)
 
--- a/release-notes/VERSION
+++ b/release-notes/VERSION
@@ -19,6 +19,9 @@ JSON library.
 
 #679: Number parsing should fail for trailing dot (period)
  (fix by @cowtowncoder, w/ Claude code)
+#1211: Add `JsonParser.willInternPropertyNames()` to check whether
+  property name interning is enabled
+ (contributed by Max P)
 #1544: Validate `read()` parameters in `MergedStream` and `UTF32Reader`
  (implemented by @pjfanning)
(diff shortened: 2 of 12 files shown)


cowtowncoder:
Quick note on naming: instead of "Field" code uses "Property" for keys of JSON Objects in 3.0. So let's rename method(s).

**EDIT**: maybe rename as `willInternPropertyNames()`

cowtowncoder:
Aside from naming (apologies, I should have updated the issue earlier), looks good.

One thing we'll need before merging is CLA (from https://github.com/FasterXML/jackson/blob/main/CLA-jackson-2026.pdf) if not yet sent (only needs to be done once, before merging the first contribution; one is good for all future contributions).
The easiest way is usually to print, sign, scan/photo, email to `cla` at fasterxml dot com.

Looking forward to merging this, thank you!

maxpaulus43:
@cowtowncoder signed CLA and updated naming

cowtowncoder:
Received CLA; hoping to merge later today.

Issue #1596, opened 2026-04-17
Incorrect license in META-INF


this-will-do:
When acquiring jackson-core from Maven, the META-INF in the jar includes `FastDoubleParser-License`. This file lists an Apache License, however FastDoubleParser is licensed under MIT, as seen here: https://github.com/wrandelshofer/FastDoubleParser/blob/main/LICENSE.

pjfanning:
I'm not near a computer right now but we have worked with the author of this lib on getting the license right.

https://github.com/FasterXML/jackson-core/blob/3.x/pom.xml#L194

There is a relocation of the license from the fastdoubleparser license from its jar in here.

I will have to look later if this is still working.

this-will-do:
I've checked the following versions from Maven which all appear to exhibit the issue : 
2.17.2
2.21.2
3.1.2

The `FastDoubleParser-NOTICE` relocation looks to work correctly in all versions. Only the `FastDoubleParser-LICENSE` appears to grab the wrong file.

pjfanning:
https://github.com/FasterXML/jackson-core/issues/1002 is an earlier issue.

The FastDoubleParser has some BSD and maybe some other extra licensing.

Apache, MIT and BSD licenses are all permissive and regarded as compatible with each other.

pjfanning:
Also see https://github.com/FasterXML/jackson-core/issues/577#issuecomment-1086600093 which allows Jackson to use the Fastdoubleparser code under the Apache license.

this-will-do:
Appreciate the extra context. I will close this issue.

PR #1597, opened 2026-04-17
Rework 3rd party licenses in jar

pjfanning:
* based on #1596 but also noticed that we could do with listing the Schubfach license for code that we have for writing numbers
* I had a look at the license info in the fastdoubleparser jar and I don't think it is fully correct, so I think it is best if we maintain our own copies of the license info
* This needs to go back to the 2.18 and 2.21 branches but I would prefer to get the 3.1 changes agreed first


cowtowncoder (approved):
No strong opinions but sounds reasonable.

src/main/resources/META-INF/FastDoubleParser-ThirdParty-LICENSE
```diff
@@ -0,0 +1,91 @@
...
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+------
+Third-party license for
+bigint, Copyright 2020 Tim Buktu. 2-clause BSD License.
```
cowtowncoder, line 61: I assume this is what we got but that name def looks made up :)
(https://en.wikipedia.org/wiki/Timbuktu)

cowtowncoder (approved):

src/main/resources/META-INF/FastDoubleParser-ThirdParty-LICENSE
pjfanning: Seems legit though - https://github.com/tbuktu/bigint

PR #1599, opened 2026-04-20
Handle issue with releasing buffers (fix #1598)

### LINKED ISSUE #1598: Possible TOCTOU timing bug in `BufferRecycler`
https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use

releaseByteBuffer / releaseCharBuffer had a non-atomic read-then-conditional-write that could let a concurrent access silently downgrade to a smaller buffer. Fixed with a CAS retry loop.

https://github.com/pjfanning/jackson-core/pull/13 has a suggested fix - ignore the changes there for the other classes, it just saves me Copilot tokens to get a few issues solved in the generated PR

I'm raising this issue to focus on the BufferRecycler suggested change and we can look at the parser changes later.

The [BufferRecycler](https://github.com/pjfanning/jackson-core/pull/13/changes#diff-1c86614f53dac6312f50486efcbae0904dca4a99fd8033b98a0d5d5297834449) change uses a VarHandle and will probably not compile with Java 8 meaning this is unlikely to backportable to 2.x.

pjfanning:
Fixes #1598 (potential problem)


pjfanning:
@cowtowncoder Animal Sniffer check for Android doesn't like the VarHandle usage. I'll look into the options.

pjfanning:
@cowtowncoder I updated the code so that it should now be ok for Animal Sniffer

cowtowncoder (approved):

PR #1601, opened 2026-04-21
performance enhancement in UTF-8 generator

pjfanning:
This was suggested by Copilot.
If there any existing JMH benchmarks for UTF-8 generator, I can try this out.
These new loops can be optimised by the JIT compiler to process 4 bytes or chars at a time instead of one byte/char at a time. 

## UTF8JsonGenerator — 4-char loop unrolling in ASCII fast path

`_writeStringSegment(char[], …)` and `_writeStringSegment(String, …)` now process 4 characters per iteration. A single bit-OR of all four values serves as the non-ASCII guard, giving the JIT a wider straight-line block to work with:

```java
while ((offset + 4) <= len) {
    int ch0 = cbuf[offset], ch1 = cbuf[offset+1],
        ch2 = cbuf[offset+2], ch3 = cbuf[offset+3];
    if (((ch0 | ch1 | ch2 | ch3) > 0x7F)
            || escCodes[ch0] != 0 || escCodes[ch1] != 0
            || escCodes[ch2] != 0 || escCodes[ch3] != 0) break;
    outputBuffer[outputPtr]   = (byte) ch0;  outputBuffer[outputPtr+1] = (byte) ch1;
    outputBuffer[outputPtr+2] = (byte) ch2;  outputBuffer[outputPtr+3] = (byte) ch3;
    outputPtr += 4; offset += 4;
}
// tail handled by the existing single-char loop
```

### DIFF
--- a/src/main/java/tools/jackson/core/json/UTF8JsonGenerator.java
+++ b/src/main/java/tools/jackson/core/json/UTF8JsonGenerator.java
@@ -1485,6 +1485,7 @@ public class UTF8JsonGenerator
             if (((ch0 | ch1 | ch2 | ch3) > 0x7F)
                     || escCodes[ch0] != 0 || escCodes[ch1] != 0
                     || escCodes[ch2] != 0 || escCodes[ch3] != 0) {
+                // handle the remaining bytes using the one at a time loop below
                 break;
             }
             outputBuffer[outputPtr++] = (byte) ch0;
@@ -1535,6 +1536,7 @@ public class UTF8JsonGenerator
             if (((ch0 | ch1 | ch2 | ch3) > 0x7F)
                     || escCodes[ch0] != 0 || escCodes[ch1] != 0
                     || escCodes[ch2] != 0 || escCodes[ch3] != 0) {
+                // handle the remaining chars using the one at a time loop below
                 break;
             }
             outputBuffer[outputPtr++] = (byte) ch0;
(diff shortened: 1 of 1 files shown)


cowtowncoder:
https://github.com/FasterXML/jackson-benchmarks has write tests. If I find time maybe I could try testing too.

src/main/java/tools/jackson/core/json/UTF8JsonGenerator.java
```diff
@@ -1471,6 +1471,30 @@ private final void _writeStringSegment(char[] cbuf, int offset, int len)
         final byte[] outputBuffer = _outputBuffer;
         final int[] escCodes = _outputEscapes;
 
+        // Unrolled 4-char inner loop: reduces loop overhead and lets the JIT
+        // see a wider window for optimisation (auto-vectorisation, etc.).
+        // We use bitwise OR of all four char values: if the OR result is > 0x7F
+        // then at least one char needs multi-byte UTF-8 encoding; then we check
+        // the escape table for each char individually.
+        while ((offset + 4) <= len) {
```
cowtowncoder, line 1479: This could be changed to calculate end (len-4) as constant to compare to.

pjfanning: It is my understanding and Gen AI backed it up that JIT compiler will optimise this as is and also if we have the extra var and this code as is is a bit simpler avoiding need for an extra variable, etc.

pjfanning:
@cowtowncoder I have a basic JMH Benchmark suite and so far, the new code appears worse than the existing code.
So for now, this might be worth leaving as is.

https://github.com/pjfanning/jackson-bench/blob/main/src/jmh/java/org/example/jackson/UTF8GenBench.java

pjfanning:
I'll close this.

Issue #784, opened 2022-06-28
Optional leading plus sign not included in textual value of any integral nor all floating-point numbers


cowtowncoder:
Implementation of #774 is mostly complete, but there is one minor inconsistency: whereas leading plus sign -- if (and only if) enabled as non-standard feature -- is retained for some floating-point numbers  when accessing token with `JsonParser.getText()` (and related), it is not retained for any of integral numbers (nor all FPs).

The underlying problem to solve is that of assumptions based on there only ever being a leading minus sign in `TextBuffer` (or shared input buffer): something that made sense before allowing non-standard leading plus.

cowtowncoder:
Come to think of this, one option is to remove leading plus sign from all cases, instead of trying to add it.
This is not ideal but is possible.

cowtowncoder:
Ok, so: currently non-standard plus (`+`) is never included in text returned. This is different from other deviations, partly since they are omissions (lack of leading `0`, or trailing digit after decimal comma).
I guess comparison could be to leading zeroes, handling of which  (in non-standard) mode I don't actually know.
But if they are not exposed we could even consider current behavior as the expected way (leave out non-standard characters). Have to think more about this -- then again it will take some effort to retain plus sign anyway so there is no urgency.

cowtowncoder:
Current thinking: `getText()` should faithfully reproduce input token -- so leading plus, even if non-standard, should be included.

PR #1613, opened 2026-05-26
Add JsonReadFeature.ALLOW_HEXADECIMAL_NUMBERS for JSON5-style hex literals (#707)

### LINKED ISSUE #707: JsonReadFeature: Introduce missing features for JSON5 compatibility

seonwooj0810:
## Summary

Implements JSON5 hexadecimal integer literals (`0x` / `0X` prefix, optional sign, `[0-9a-fA-F]+`) across all four JSON parsers, gated behind a new opt-in `JsonReadFeature.ALLOW_HEXADECIMAL_NUMBERS`. Closes #707 (hex part; whitespace remains).

Per @pjfanning's [review](https://github.com/FasterXML/jackson-core/issues/707#issuecomment-4543767243):

- **JSON5 grammar followed verbatim** — `0x|0X` + `[0-9a-fA-F]+`, optional leading sign, no fraction / no exponent / no `_` separators.
- **Independent of `ALLOW_LEADING_ZEROS_FOR_NUMBERS`** — leading zeros within hex digits (e.g. `0x007F`) are always permitted when this feature is enabled, regardless of the leading-zero feature. The hex check fires **before** `_verifyNoLeadingZeroes()` in every parser, and the leading-zero feature is never consulted on the hex path. Covered by `unsignedHexWithLeadingZeros`.

The `+` variant additionally requires `ALLOW_LEADING_PLUS_SIGN_FOR_NUMBERS` (matching how `+` is treated for decimals, per #774). Covered by `plusSignHexRequiresLeadingPlusFeature` and `plusSignHexWithPlusFeature`.

Implementation per parser:

| Parser | Strategy |
| --- | --- |
| `ReaderBasedJsonParser` | Hex branch in `_parseNumber2`, before `_verifyNoLeadingZeroes`. |
| `UTF8StreamJsonParser` | Same pattern, byte-oriented; peek before `_verifyNoLeadingZeroes` in both `_parseUnsignedNumber` and `_parseSignedNumber`. |
| `UTF8DataInputJsonParser` | Intercept the existing `'x'`/`'X'` arm of `_handleLeadingZeroes` (no peek needed — DataInput already consumed it). |
| `NonBlockingUtf8JsonParserBase` | New minor states `MINOR_NUMBER_HEX_PREFIX` and `MINOR_NUMBER_HEX_DIGITS` so suspension works at every byte boundary; helpers `_startHexNumber` / `_startHexNumberWithSign` / `_finishHexDigits` drive the state machine. |

Shared decoding lives in `JsonParserBase._parseHexInt`:
- `int` for ≤ 7 hex digits (always fits in signed 32-bit),
- `long` for ≤ 16 hex digits with a top-nibble check at 16,
- `BigInteger` otherwise — eagerly decoded because the lazy `_numberString` path expects base-10.

`getString()` returns the original literal (sign + `0x`/`0X` + digits), matching the precedent set for `+` by #784. `getIntValue()` / `getLongValue()` / `getBigIntegerValue()` return the decoded value.

The bookkeeping additions on `JsonParserBase` are limited:
- `_numberIsHex` field (cleared by `resetInt` / `resetFloat` / `resetAsNaN`),
- `resetIntHex(neg, hexDigitLen)` helper alongside `resetInt`.

I put these on `JsonParserBase` only (since only JSON consumes it today).

## Test plan

- [x] `mvn clean verify` — 1697 tests, 0 failures, 0 regressions.
- [x] `NonStandardHexNumbers707Test` (12 cases) — unsigned, uppercase `X`, leading zeros, negative, `+` with/without plus-feature, long range (`0x80000000`), BigInteger range (`0x1ffffffffffffffff`), in-array, feature-off rejection, empty-prefix rejection. Runs across `ALL_MODES` (Reader, UTF8 stream, throttled, DataInput).
- [x] `AsyncHexNumbers707Test` (10 cases) — drives the non-blocking parser with 1-byte and 3-byte feeds to exercise suspension/resumption at every position; covers the same shapes plus a `plainZeroStillWorks` smoke test to verify the bare-`0` path didn't regress.

### DIFF
--- a/src/main/java/tools/jackson/core/base/ParserBase.java
+++ b/src/main/java/tools/jackson/core/base/ParserBase.java
@@ -169,19 +169,6 @@ public abstract class ParserBase extends ParserMinimalBase
      */
     protected boolean _numberIsNaN;
 
-    /**
-     * Marker for integer values read using JSON5 hexadecimal notation
-     * ({@code 0x} / {@code 0X} prefix), enabled via
-     * {@link tools.jackson.core.json.JsonReadFeature#ALLOW_HEXADECIMAL_NUMBERS}.
-     * When {@code true}, the textual representation buffered for the current
-     * token is the original hex literal (including any sign and the
-     * {@code 0x}/{@code 0X} prefix) and {@link #_intLength} records the
-     * number of hexadecimal digits (excluding sign and prefix).
-     *
-     * @since 3.2
-     */
-    protected boolean _numberIsHex;
-
     // And then other information about value itself
 
     /**
@@ -411,40 +398,14 @@ public abstract class ParserBase extends ParserMinimalBase
         return resetFloat(negative, intLen, fractLen, expLen);
     }
 
-    protected final JsonToken resetInt(boolean negative, int intLen)
+    protected JsonToken resetInt(boolean negative, int intLen)
         throws JacksonException
     {
         // May throw StreamConstraintsException:
         _streamReadConstraints.validateIntegerLength(intLen);
         _numberNegative = negative;
         _numberIsNaN = false;
-        _numberIsHex = false;
         _intLength = intLen;
-        _fractLength = 0;
-        _expLength = 0;
-        _numTypesValid = NR_UNKNOWN; // to force decoding
-        _numberString = null;
-        return JsonToken.VALUE_NUMBER_INT;
-    }
-
-    /**
-     * Variant of {@link #resetInt} used for integer values read in JSON5
-     * hexadecimal notation ({@code 0x...}). {@code hexDigitLen} is the
-     * number of hexadecimal digits (excluding sign and {@code 0x}/{@code 0X}
-     * prefix); the textual representation buffered by the caller is expected
-     * to contain the original literal including sign and prefix.
-     *
-     * @since 3.2
-     */
-    protected final JsonToken resetIntHex(boolean negative, int hexDigitLen)
-        throws JacksonException
-    {
-        // May throw StreamConstraintsException:
-        _streamReadConstraints.validateIntegerLength(hexDigitLen);
-        _numberNegative = negative;
-        _numberIsNaN = false;
-        _numberIsHex = true;
-        _intLength = hexDigitLen;
         _fractLength = 0;
         _expLength = 0;
         _numTypesValid = NR_UNKNOWN; // to force decoding
@@ -459,7 +420,6 @@ public abstract class ParserBase extends ParserMinimalBase
         _streamReadConstraints.validateFPLength(intLen + fractLen + expLen);
         _numberNegative = negative;
         _numberIsNaN = false;
-        _numberIsHex = false;
         _intLength = intLen;
         _fractLength = fractLen;
         _expLength = expLen;
@@ -474,7 +434,6 @@ public abstract class ParserBase extends ParserMinimalBase
         _numberDouble = value;
         _numTypesValid = NR_DOUBLE;
         _numberIsNaN = true;
-        _numberIsHex = false;
         _numberString = null;
         return JsonToken.VALUE_NUMBER_FLOAT;
     }
--- a/src/main/java/tools/jackson/core/base/ParserBase.java
+++ b/src/main/java/tools/jackson/core/base/ParserBase.java
@@ -169,6 +169,19 @@ public abstract class ParserBase extends ParserMinimalBase
      */
     protected boolean _numberIsNaN;
 
+    /**
+     * Marker for integer values read using JSON5 hexadecimal notation
+     * ({@code 0x} / {@code 0X} prefix), enabled via
+     * {@link tools.jackson.core.json.JsonReadFeature#ALLOW_HEXADECIMAL_NUMBERS}.
+     * When {@code true}, the textual representation buffered for the current
+     * token is the original hex literal (including any sign and the
+     * {@code 0x}/{@code 0X} prefix) and {@link #_intLength} records the
+     * number of hexadecimal digits (excluding sign and prefix).
+     *
+     * @since 3.2
+     */
+    protected boolean _numberIsHex;
+
     // And then other information about value itself
 
     /**
@@ -405,7 +418,33 @@ public abstract class ParserBase extends ParserMinimalBase
         _streamReadConstraints.validateIntegerLength(intLen);
         _numberNegative = negative;
         _numberIsNaN = false;
+        _numberIsHex = false;
         _intLength = intLen;
+        _fractLength = 0;
+        _expLength = 0;
+        _numTypesValid = NR_UNKNOWN; // to force decoding
+        _numberString = null;
+        return JsonToken.VALUE_NUMBER_INT;
+    }
+
+    /**
+     * Variant of {@link #resetInt} used for integer values read in JSON5
+     * hexadecimal notation ({@code 0x...}). {@code hexDigitLen} is the
+     * number of hexadecimal digits (excluding sign and {@code 0x}/{@code 0X}
+     * prefix); the textual representation buffered by the caller is expected
+     * to contain the original literal including sign and prefix.
+     *
+     * @since 3.2
+     */
+    protected final JsonToken resetIntHex(boolean negative, int hexDigitLen)
+        throws JacksonException
+    {
+        // May throw StreamConstraintsException:
+        _streamReadConstraints.validateIntegerLength(hexDigitLen);
+        _numberNegative = negative;
+        _numberIsNaN = false;
+        _numberIsHex = true;
+        _intLength = hexDigitLen;
         _fractLength = 0;
         _expLength = 0;
         _numTypesValid = NR_UNKNOWN; // to force decoding
@@ -420,6 +459,7 @@ public abstract class ParserBase extends ParserMinimalBase
         _streamReadConstraints.validateFPLength(intLen + fractLen + expLen);
         _numberNegative = negative;
         _numberIsNaN = false;
+        _numberIsHex = false;
         _intLength = intLen;
         _fractLength = fractLen;
         _expLength = expLen;
@@ -434,6 +474,7 @@ public abstract class ParserBase extends ParserMinimalBase
         _numberDouble = value;
         _numTypesValid = NR_DOUBLE;
         _numberIsNaN = true;
+        _numberIsHex = false;
         _numberString = null;
         return JsonToken.VALUE_NUMBER_FLOAT;
     }
--- a/src/main/java/tools/jackson/core/base/ParserBase.java
+++ b/src/main/java/tools/jackson/core/base/ParserBase.java
@@ -398,6 +398,9 @@ public abstract class ParserBase extends ParserMinimalBase
         return resetFloat(negative, intLen, fractLen, expLen);
     }
 
+    // NOTE: was `final` before 3.2; relaxed so that `JsonParserBase` can
+    // override to clear hex-specific state on integer reset (the sibling
+    // `resetFloat` / `resetAsNaN` remain `final`).
     protected JsonToken resetInt(boolean negative, int intLen)
         throws JacksonException
     {
--- a/src/main/java/tools/jackson/core/io/BigIntegerParser.java
+++ b/src/main/java/tools/jackson/core/io/BigIntegerParser.java
@@ -36,4 +36,18 @@ public final class BigIntegerParser
                     ", reason: " + nfe.getMessage());
         }
     }
+
+    public static BigInteger parseWithFastParser(final char[] ch, final int offset,
+            final int length, final int radix) {
+        try {
+            return JavaBigIntegerParser.parseBigInteger(ch, offset, length, radix);
+        } catch (NumberFormatException nfe) {
+            final String reportNum = length <= MAX_CHARS_TO_REPORT
+                    ? new String(ch, offset, length)
+                    : new String(ch, offset, MAX_CHARS_TO_REPORT) + " [truncated]";
+            throw new NumberFormatException("Value \"" + reportNum
+                    + "\" cannot be represented as `java.math.BigInteger` with radix " + radix +
+                    ", reason: " + nfe.getMessage());
+        }
+    }
 }
--- a/src/main/java/tools/jackson/core/io/BigIntegerParser.java
+++ b/src/main/java/tools/jackson/core/io/BigIntegerParser.java
@@ -37,6 +37,9 @@ public final class BigIntegerParser
         }
     }
 
+    /**
+     * @since 3.2
+     */
     public static BigInteger parseWithFastParser(final char[] ch, final int offset,
             final int length, final int radix) {
         try {
--- a/src/main/java/tools/jackson/core/json/JsonParserBase.java
+++ b/src/main/java/tools/jackson/core/json/JsonParserBase.java
@@ -1,5 +1,7 @@
 package tools.jackson.core.json;
 
+import java.math.BigInteger;
+
 import tools.jackson.core.*;
 import tools.jackson.core.base.ParserBase;
 import tools.jackson.core.exc.InputCoercionException;
@@ -192,6 +194,10 @@ public abstract class JsonParserBase
     {
         // Int or float?
         if (_currToken == JsonToken.VALUE_NUMBER_INT) {
+            if (_numberIsHex) {
+                _parseHexInt(expType);
+                return;
+            }
             int len = _intLength;
             // First: optimization for simple int
             if (len <= 9) {
@@ -250,7 +256,9 @@ public abstract class JsonParserBase
     {
         // Inlined variant of: _parseNumericValue(NR_INT)
         if (_currToken == JsonToken.VALUE_NUMBER_INT) {
-            if (_intLength <= 9) {
+            // Hex integers go through the generic path so the base-16 decode is
+            // applied (the base-10 fast path below would mis-read the literal):
+            if (!_numberIsHex && _intLength <= 9) {
                 int i = _textBuffer.contentsAsInt(_numberNegative);
                 _numberInt = i;
                 _numTypesValid = NR_INT;
@@ -295,6 +303,85 @@ public abstract class JsonParserBase
             _numberString = _textBuffer.contentsAsString();
             _numTypesValid = NR_DOUBLE;
         }
+    }
+
+    /**
+     * Decode a JSON5 hexadecimal integer that was buffered as the original
+     * textual literal (sign + {@code 0x}/{@code 0X} prefix + hex digits).
+     * {@link #_intLength} holds the count of hex digits.
+     *
+     * @since 3.2
+     */
+    private void _parseHexInt(int expType) throws JacksonException
+    {
+        final int hexLen = _intLength;
+        final char[] buf = _textBuffer.getTextBuffer();
+        // Locate the first hex digit: skip optional sign and "0x" / "0X" prefix
+        int idx = _textBuffer.getTextOffset();
+        final char first = buf[idx];
+        if (first == '-' || first == '+') {
+            ++idx;
+        }
+        idx += 2; // skip "0x" / "0X"
+
+        // Up to 7 hex digits always fit in a positive signed int (<= 0x0FFFFFFF).
+        // 8 hex digits may overflow signed int (e.g. 0x80000000), so we defer to
+        // the long path which handles range checks uniformly.
+        if (hexLen <= 7) {
+            int v = 0;
+            for (int i = 0; i < hexLen; ++i) {
+                v = (v << 4) | _hexDigit(buf[idx + i]);
+            }
+            _numberInt = _numberNegative ? -v : v;
+            _numTypesValid = NR_INT;
+            return;
+        }
+        // 9..15 hex digits always fit in a positive long (63 bits used at most)
+        if (hexLen <= 15) {
+            long v = 0L;
+            for (int i = 0; i < hexLen; ++i) {
+                v = (v << 4) | _hexDigit(buf[idx + i]);
+            }
+            _numberLong = _numberNegative ? -v : v;
+            _numTypesValid = NR_LONG;
+            return;
+        }
+        // 16 hex digits: may or may not fit in signed long, depending on top bit
+        if (hexLen == 16) {
+            int topNibble = _hexDigit(buf[idx]);
+            if (topNibble < 0x8) { // fits in positive signed long
+                long v = topNibble;
+                for (int i = 1; i < 16; ++i) {
+                    v = (v << 4) | _hexDigit(buf[idx + i]);
+                }
+                _numberLong = _numberNegative ? -v : v;
+                _numTypesValid = NR_LONG;
+                return;
+            }
+            // else fall through to BigInteger path
+        }
+        // Larger values -> BigInteger. We must eagerly decode here (the lazy
+        // base-10 path via _numberString would mis-read hex digits).
+        String digits = new String(buf, idx, hexLen);
+        BigInteger bi = new BigInteger(digits, 16);
+        if (_numberNegative) {
+            bi = bi.negate();
+        }
+        _numberBigInt = bi;
+        _numberString = null;
+        _numTypesValid = NR_BIGINT;
+        if ((expType == NR_INT) || (expType == NR_LONG)) {
+            // Force the overflow path to surface a meaningful error
+            _reportTooLongIntegral(expType, _textBuffer.contentsAsString());
+        }
+    }
+
+    private static int _hexDigit(char c) {
+        if (c <= '9') {
+            return c - '0';
+        }
+        // 'A'..'F' -> 10..15, 'a'..'f' -> 10..15
+        return (c & 0x1F) + 9;
     }
 
     private void _parseSlowInt(int expType) throws JacksonException
--- a/src/main/java/tools/jackson/core/json/JsonParserBase.java
+++ b/src/main/java/tools/jackson/core/json/JsonParserBase.java
@@ -51,6 +51,19 @@ public abstract class JsonParserBase
      */
     protected JsonToken _nextToken;
 
+    /**
+     * Marker for integer values read using JSON5 hexadecimal notation
+     * ({@code 0x} / {@code 0X} prefix), enabled via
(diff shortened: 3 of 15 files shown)


src/main/java/tools/jackson/core/json/JsonParserBase.java
```diff
@@ -297,6 +305,85 @@ private void _parseSlowFloat(int expType) throws JacksonException
...
+                _numberLong = _numberNegative ? -v : v;
+                _numTypesValid = NR_LONG;
+                return;
+            }
+            // else fall through to BigInteger path
+        }
+        // Larger values -> BigInteger. We must eagerly decode here (the lazy
+        // base-10 path via _numberString would mis-read hex digits).
+        String digits = new String(buf, idx, hexLen);
+        BigInteger bi = new BigInteger(digits, 16);
```
pjfanning: BigInteger is very slow. Can you consider adding a util method to NumberInput.java? NumberInput methods support fastdoubleparser and I'm pretty sure that it should have a parser method for BigInteger with a radix=16. So the choice of fastdoubleparser or java.math.BigInteger should be similar to how other methods in NumberInput work.

pjfanning: https://github.com/wrandelshofer/FastDoubleParser/blob/add7f778b480f3640afea4c3e33d10e2bc99cf1c/fastdoubleparser-dev/src/main/java/ch.randelshofer.fastdoubleparser/ch/randelshofer/fastdoubleparser/JavaBigIntegerFromCharArray.java

Pass the char array or byte array (there is a separate class for parsing byte arrays and another one for char sequences, including strings). Avoid constructing an intermediate string to save on the overhead.

seonwooj0810: Addressed in 45e02b38.

- Added `NumberInput.parseBigIntegerWithRadix(char[], offset, length, radix, useFastParser)` and a matching `BigIntegerParser.parseWithFastParser(char[], int, int, int)` overload, so the `useFastParser` toggle mirrors the existing String-based methods.
- `_parseHexInt` now hands the char[] slice straight to `JavaBigIntegerParser` (no intermediate `String`) when `USE_FAST_BIG_NUMBER_PARSER` is enabled.
- Added unit tests for the new char[] overload in `NumberInputTest` and `BigIntegerParserTest`.

seonwooj0810:
@pjfanning Addressed your review feedback in 45e02b38 (moved BigInteger parsing into `NumberInput.parseBigIntegerWithRadix(char[], ...)` and pass the char slice directly to FastDoubleParser, avoiding the intermediate String). Could you take another look when you have a moment?

src/test/java/tools/jackson/core/unittest/io/NumberInputTest.java
```diff
@@ -64,6 +64,17 @@ void bigIntegerWithRadix()
...
     }
 
+    @Test
+    void bigIntegerWithRadixFromCharArray()
+    {
+        // Embed the digits inside a larger buffer so offset/length are exercised.
+        char[] buf = ("xx" + "1ABCDEF" + "yy").toCharArray();
+        final int radix = 16;
+        BigInteger expected = new BigInteger("1ABCDEF", radix);
+        assertEquals(expected, NumberInput.parseBigIntegerWithRadix(buf, 2, 7, radix, true));
```
pjfanning: this is not complete testing - add tests with 0 offset too and and arrays where you then use a length that ignores the end chars

pjfanning: negative numbers too

src/test/java/tools/jackson/core/unittest/read/NonStandardHexNumbers707Test.java
```diff
@@ -0,0 +1,162 @@
...
+import tools.jackson.core.json.JsonReadFeature;
+import tools.jackson.core.unittest.JacksonCoreTestBase;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * Tests for [core#707]: JSON5-style hexadecimal integer literals enabled via
+ * {@link JsonReadFeature#ALLOW_HEXADECIMAL_NUMBERS}.
+ */
+class NonStandardHexNumbers707Test extends JacksonCoreTestBase
```
pjfanning, line 22: add negative tests - maybe to an existing test class - that tests that hex numbers fail when the new feature is not enabled

src/test/java/tools/jackson/core/unittest/io/NumberInputTest.java
seonwooj0810: Addressed in 8ea2cd5a. The char[] tests now cover:

- `offset=0` with `length` spanning the full array,
- `offset=0` with `length` deliberately shorter than the array (trailing chars ignored),
- `offset>0` with both leading and trailing chars sliced away,
- a separate `bigIntegerWithRadixFromCharArrayNegative` test exercising the same three shape variants with a leading `-` sign.

All run against both `useFastParser=true` and `useFastParser=false`.

cowtowncoder:
Hmmmh. This is HUGE. Not a big fan based on size alone... :-(

Will give some more thought.

seonwooj0810:
Quick note on the "fail when feature is not enabled" request — that scenario was actually already covered from the initial commit by `hexRejectedWhenFeatureDisabled` in both `NonStandardHexNumbers707Test` and `AsyncHexNumbers707Test` (each runs through all the hex prefix shapes with `ALLOW_HEXADECIMAL_NUMBERS` disabled and asserts they fail). I should have called that out explicitly in my reply — apologies for the noise.

Also pushed 4e8e0609 to add invalid-hex digit coverage (e.g. `0xZZ`) to the new `bigIntegerWithRadixFromCharArray*` tests, in case that was your intent rather than feature-disabled. Happy to add more if I'm still missing something.

seonwooj0810:
@cowtowncoder Totally hear you on the size — it's something I've been chewing on too. Let me break down what's in the +940 to make it easier to evaluate, and offer a couple of options.

**Where the lines actually go:**

| Bucket | Lines | What it is |
|---|---|---|
| Tests | ~383 (40%) | `NonStandardHexNumbers707Test`, `AsyncHexNumbers707Test`, plus the `NumberInput` / `BigIntegerParser` char[]-overload coverage @pjfanning asked for |
| Feature definition + utils | ~70 (7%) | `JsonReadFeature` constant, `NumberInput.parseBigIntegerWithRadix(char[],...)`, `BigIntegerParser.parseWithFastParser(char[],...)` |
| Parser plumbing | ~484 (51%) | Hex parsing wired into 7 parser variants |

**Why parser plumbing is 51%:** the feature has to land in each concrete parser independently because the input sources (Reader, byte stream, DataInput, async buffer) don't share a unified read primitive. I tried to keep as much as possible in `JsonParserBase` (+89 for shared state and entry points), but the per-variant additions are doing input-source-specific work — Reader uses `_loadMore()`, DataInput uses `readUnsignedByte()` with IOException, async needs MINOR state restart points, etc.

The one outlier is `NonBlockingUtf8JsonParserBase` at +123. That's not duplication — async needs `MINOR_NUMBER_HEX_PREFIX`/`MINOR_NUMBER_HEX_DIGITS` states and the `_finishHexDigits(requireFirst)` / `_completeHexNumber()` restart-able split that sync parsers don't need. Without it, hex parsing would silently break on partial buffers.

**What I can trim (good faith, ~15–20 lines):** `_isHexDigit` is defined four times across parsers, and the "prefix must be followed by at least one hex digit" message is duplicated. I can hoist the digit check to `ParserBase` (with a `char→int` widening for `ReaderBasedJsonParser`) and extract the message as a constant. Happy to do this if you'd like — it's a small win at the cost of adding one method to the base class.

**Suggested review path** if it helps to chunk it:
1. `JsonReadFeature.java` — feature intent (1 constant)
2. `NumberInput.java` + `BigIntegerParser.java` — the new char[] overloads (used by all parsers)
3. `JsonParserBase.java` — shared state / entry points
4. One of the concrete parsers (`ReaderBasedJsonParser.java` is the simplest) — to understand the per-variant pattern
5. The remaining parsers are mechanical applications of the same pattern, just adapted to their input source

**Alternative if size is still the blocker:** I could split into a prelim PR that lands just the utils (`NumberInput`/`BigIntegerParser` char[] overloads + `JsonReadFeature` constant) ahead of this one. It wouldn't change net diff but would give you two ~half-size reviews instead of one large one. Let me know if that's worth doing or if you'd rather keep it as one atomic change.

Whichever direction you'd like to take, let me know and I'll make it happen.

src/main/java/tools/jackson/core/base/ParserBase.java
```diff
@@ -169,6 +169,19 @@ public abstract class ParserBase extends ParserMinimalBase
...
+     * ({@code 0x} / {@code 0X} prefix), enabled via
+     * {@link tools.jackson.core.json.JsonReadFeature#ALLOW_HEXADECIMAL_NUMBERS}.
+     * When {@code true}, the textual representation buffered for the current
+     * token is the original hex literal (including any sign and the
+     * {@code 0x}/{@code 0X} prefix) and {@link #_intLength} records the
+     * number of hexadecimal digits (excluding sign and prefix).
+     *
+     * @since 3.2
+     */
+    protected boolean _numberIsHex;
```
cowtowncoder: Ideally all of these would instead go in `JsonParserBase` which is only used by JSON parsers -- `ParserBase` is shared with other format backends too.
Would it be possible to promote functionality there?

src/main/java/tools/jackson/core/json/async/NonBlockingUtf8JsonParserBase.java
```diff
@@ -1691,6 +1727,92 @@ protected JsonToken _finishNumberLeadingPosNegZeroes(final boolean negative) thr
...
+        char[] outBuf = _textBuffer.getBufferWithoutReset();
+        int outPtr = _textBuffer.getCurrentSegmentSize();
+        final boolean hasSignChar = (outBuf[0] == '-' || outBuf[0] == '+');
+        final int prefixLen = hasSignChar ? 3 : 2;
+        final int hexLen = outPtr - prefixLen;
+        resetIntHex(_numberNegative, hexLen);
+        return _valueComplete(JsonToken.VALUE_NUMBER_INT);
+    }
+
+    private static boolean _isHexDigit(int c) {
```
cowtowncoder: Could/should go in `JsonParserBase` as `protected` for reuse?

src/main/java/tools/jackson/core/json/UTF8DataInputJsonParser.java
```diff
@@ -1160,6 +1171,56 @@ private final JsonToken _parseSignedNumber(boolean negative) throws IOException
...
+        }
+        _textBuffer.setCurrentLength(outPtr);
+        _nextByte = c;
+        if (_streamReadContext.inRoot()) {
+            _verifyRootSpace();
+        }
+        return resetIntHex(neg, hexLen);
+    }
+
+    private static boolean _isHexDigit(int c) {
```
cowtowncoder: Too many copies, share via `JsonParserBase`.

Or better yet, check out what `CharTypes.java` already has; maybe `charToHex()` would be useful (gives `-1` for non-hex digit)

seonwooj0810:
@cowtowncoder Pushed `a4bab461` and `52e7a897` to address your three inline comments:

**`a4bab461` — Move hex parser state to `JsonParserBase`, dedupe per-parser helpers**
- [`ParserBase.java:183`](https://github.com/FasterXML/jackson-core/pull/1613#discussion_r3307833616): `_numberIsHex` field and `resetIntHex(...)` promoted to `JsonParserBase`. `ParserBase` no longer references hex at all. `resetInt` loses `final` so `JsonParserBase` can override it to clear `_numberIsHex`; `resetFloat`/`resetAsNaN` keep `final` (the flag is only read in the `VALUE_NUMBER_INT` branch, both reached only via `resetInt`/`resetIntHex`, so a stale value after a float reset is benign).
- [`NonBlockingUtf8JsonParserBase.java:1810`](https://github.com/FasterXML/jackson-core/pull/1613#discussion_r3307840258): the four duplicated `_isHexDigit` definitions consolidated as one `protected static` on `JsonParserBase` (then superseded entirely by the next commit — see below).
- The four duplicated hex-prefix error messages extracted to `JsonParserBase._hexPrefixNotFollowedMessage(char)`.

**`52e7a897` — Use `CharTypes.charToHex` for JSON5 hex literal scan/decode**
- [`UTF8DataInputJsonParser.java:1218`](https://github.com/FasterXML/jackson-core/pull/1613#discussion_r3307848784): great suggestion — `CharTypes.charToHex(int)` is a single masked array lookup that returns the digit value (0..15) or `-1`, so it subsumes **both** helpers introduced for this feature:
    - `_isHexDigit(int)` → `CharTypes.charToHex(c) >= 0`
    - `_hexDigit(char)` → `CharTypes.charToHex(c)` (no separate decode needed)
- Both helpers removed; the four concrete parser scan loops and `JsonParserBase._parseHexInt` now go through `CharTypes.charToHex` directly, matching the pattern already used elsewhere in these parsers for `\uXXXX` escape decoding.

**Net diff** is now **+923 / -4** across 16 files (was +940 / -3). Modest line reduction but materially cleaner structure: zero hex references in cross-format `ParserBase`, zero per-parser duplication, and the digit-validation/digit-decode pair fused into one well-established primitive.

Full test suite: **1702 pass / 2 pre-existing skips**, including `NonStandardHexNumbers707Test` (12 tests) and `AsyncHexNumbers707Test` (10 tests). Happy to keep iterating.

src/main/java/tools/jackson/core/base/ParserBase.java
```diff
@@ -398,7 +398,7 @@ protected final JsonToken reset(boolean negative, int intLen, int fractLen, int
         return resetFloat(negative, intLen, fractLen, expLen);
     }
 
-    protected final JsonToken resetInt(boolean negative, int intLen)
+    protected JsonToken resetInt(boolean negative, int intLen)
```
cowtowncoder, line 404: Worth adding comment that this was `final` before 3.2.

src/main/java/tools/jackson/core/io/BigIntegerParser.java
```diff
@@ -36,4 +36,18 @@ public static BigInteger parseWithFastParser(final String valueStr, final int ra
                     ", reason: " + nfe.getMessage());
         }
     }
+
+    public static BigInteger parseWithFastParser(final char[] ch, final int offset,
```
cowtowncoder, line 43: Add

```
/**
  * @since 3.2
  */
```

src/main/java/tools/jackson/core/json/JsonParserBase.java
```diff
@@ -186,12 +202,52 @@ protected void createChildObjectContext(final int lineNr, final int colNr) throw
...
+     * prefix); the textual representation buffered by the caller is expected
+     * to contain the original literal including sign and prefix.
+     *
+     * @since 3.2
+     */
+    protected final JsonToken resetIntHex(boolean negative, int hexDigitLen)
+        throws JacksonException
+    {
+        // May throw StreamConstraintsException:
+        _streamReadConstraints.validateIntegerLength(hexDigitLen);
```
cowtowncoder, line 229: Although this works, ideally should validate intermediate length on buffer boundaries (when accumulating digits), so input with, say, 10 million digits would be caught at ~4k or whatever first boundary is (assuming default max length of 1000 digits).

src/main/java/tools/jackson/core/json/JsonParserBase.java
```diff
@@ -297,6 +355,89 @@ private void _parseSlowFloat(int expType) throws JacksonException
...
+     * Decode a JSON5 hexadecimal integer that was buffered as the original
+     * textual literal (sign + {@code 0x}/{@code 0X} prefix + hex digits).
+     * {@link #_intLength} holds the count of hex digits.
+     *
+     * @since 3.2
+     */
+    private void _parseHexInt(int expType) throws JacksonException
+    {
+        final int hexLen = _intLength;
+        final char[] buf = _textBuffer.getTextBuffer();
```
cowtowncoder, line 368: Instead of decoding here, would it make more sense to do it in `TextBuffer` itself?

src/main/java/tools/jackson/core/json/ReaderBasedJsonParser.java
```diff
@@ -1581,6 +1581,16 @@ private final JsonToken _parseNumber2(boolean neg, int startPtr) throws JacksonE
         char c = (_inputPtr < _inputEnd) ? _inputBuffer[_inputPtr++]
                 : getNextChar("No digit following sign", JsonToken.VALUE_NUMBER_INT);
         if (c == '0') {
+            // [core#707]: JSON5 hexadecimal literal ('0x' / '0X')?
+            // Must be checked BEFORE _verifyNoLeadingZeroes(): leading zeros are
+            // valid in hex regardless of ALLOW_LEADING_ZEROS_FOR_NUMBERS.
+            if (_inputPtr < _inputEnd || _loadMore()) {
+                char peek = _inputBuffer[_inputPtr];
+                if ((peek == 'x' || peek == 'X')
```
cowtowncoder: Idea: add a helper method ("_checkHexNumbersAllowed()" etc) in `JsonParserBase` called after "0x" or "0X" seen, passing next character -- if feature NOT enabled, can give specific error message including feature to enable to allow hexadecimal numbers: either returns (indicating feature enabled) or throws exception (is not).

Reduces duplication slightly but more importantly gives better exception in case feature disabled.

cowtowncoder:
@seonwooj0810 wrt:

    Drive-by fix (please call out if you'd prefer this split)

yes, please, that'd be worth fixing separately, can merge quickly.

cowtowncoder:
@seonwooj0810 Ok, aside from review (with which are making good progress), one process thing: unless we have asked for it before (and got it), we'd need CLA. It's from:

and is usually done by printing, filling & signing, scan/photo, email to `cla` at fasterxml dot com.

Only needs to be sent once and is good for all future contributions.

seonwooj0810:
@cowtowncoder Pushed commits `8e5a4b07`, `c759815d`, `0f5a076e`, `f306baa1` addressing the rest of your review (and split out the drive-by as **#1614** per your request). Summary by item:

**`8e5a4b07` — Align `MINOR_NUMBER_PLUSZERO` numbering with the split**
- The `+0...` async-resumption primitive (`MINOR_NUMBER_PLUSZERO`) now lives in #1614. Renumbered to `33` in this PR too (shifted `HEX_PREFIX`/`HEX_DIGITS` to `34`/`35`) so when #1614 lands and 3.x merges back into this branch, the constant lines up cleanly with no numeric collision.
- Note: this branch still defines `PLUSZERO` itself because the hex `+0xff` async path needs it (`positiveHexWithPlusSignAndSplit` fails without it). When #1614 merges, the duplicate definition resolves to a trivial dedupe.

**`c759815d` — Document `resetInt` relaxation and char[] BigInteger parser**
- [`ParserBase.resetInt`](https://github.com/FasterXML/jackson-core/pull/1613/files#diff-...): added the "was final before 3.2" comment you asked for.
- [`BigIntegerParser.parseWithFastParser(char[],...)`](https://github.com/FasterXML/jackson-core/pull/1613/files#diff-...): added `@since 3.2`.

**`0f5a076e` — `_checkHexNumbersAllowed` helper for actionable feature-disabled error**
- New helper in `JsonParserBase`; called after `'x'`/`'X'` is seen following a leading `0`. If `ALLOW_HEXADECIMAL_NUMBERS` is disabled, throws via `_reportUnexpectedChar` with the feature name embedded:
  ```
  Unexpected character ('x') (code 120): hexadecimal number literals
  require enabling `JsonReadFeature.ALLOW_HEXADECIMAL_NUMBERS`
  ```
- All eight call sites across the four parsers now route through the helper — no more duplicated `isEnabled(ALLOW_HEXADECIMAL_NUMBERS)` checks at parser-specific entry points.
- `hexRejectedWhenFeatureDisabled` (sync + async) tightened to assert the feature-name message in addition to the original "Unexpected character ('x'" prefix.

**`f306baa1` — Validate hex digit length at every segment boundary**
- Per your comment about the 10M-digit case: validation now fires at every `_textBuffer` segment boundary in the hex-digit accumulation loop of all four parsers, not just at final `resetIntHex`. Default first segment is 500 chars, so with `maxNumberLength=1000` a giant literal is caught long before it allocates.
- Added `hexNumberLengthConstraint` test (8000-digit literal with `maxNumberLength=100`) exercising all sync modes; async path covered by the same primitive inside `_finishHexDigits`.

**One item I want to flag rather than implement here**: your suggestion to move hex decoding into `TextBuffer` itself ([this comment](https://github.com/FasterXML/jackson-core/pull/1613#discussion_r3307...)) feels like it would meaningfully grow this PR — and since you've already noted the size concern, **I'd prefer to spin it out into a follow-up PR after this lands**. Happy to do it either way, just call out whether you'd like it here or separately.

**Net effect after this round**: +1003 / -22 across 18 files (was +923 / -4 across 16 files). The bump is roughly evenly split between the new helper + boundary test (+~80) and the same churn spread across the four parsers — but the cross-parser duplication is now down (one feature check, one boundary check primitive each).

**Full test suite**: 1703 pass / 2 pre-existing skips (was 1702 — added `hexNumberLengthConstraint`).

**Process note**: CLA will follow by email shortly; same signature covers both this PR and #1614.

src/main/java/tools/jackson/core/base/ParserBase.java
seonwooj0810: Done in 7acfb5cf — added a comment noting `resetInt` was `final` before 3.2 and why `JsonParserBase` needs to override it to clear `_numberIsHex`.

src/main/java/tools/jackson/core/io/BigIntegerParser.java
seonwooj0810: Done in 7acfb5cf — added `@since 3.2` to the new `parseWithFastParser(char[], int, int, int)` overload.

src/main/java/tools/jackson/core/json/JsonParserBase.java
seonwooj0810: Done in d1bf5c22. Max-length check now runs inside the hex-digit accumulation loop, firing at every `_textBuffer` segment boundary (first segment ≈500 chars). With default `maxNumberLength=1000`, a 10M-digit hex literal is rejected at ~4 KiB rather than after full buffering. Added `hexNumberLengthConstraint` to `NonStandardHexNumbers707Test` exercising 8000 digits against `maxNumberLength=100` across all sync backends; the async path exercises the same `_finishHexDigits` validation primitive.

src/main/java/tools/jackson/core/json/JsonParserBase.java
seonwooj0810: Considered this but held off for the initial PR. `TextBuffer` is shared across every format backend and currently has no number-format awareness — adding hex-aware decoding would couple a foundational utility to JSON5-specific semantics, and the int/long fast paths also want direct access to `_numberInt` / `_numberLong` / `_numberNegative` / `_numTypesValid` on the parser. Given the size concern raised earlier on this PR, leaning toward keeping this focused and revisiting the `TextBuffer` move as a follow-up once #1613 lands. Happy to file a tracking issue if you'd like, or sketch a shape if you have one in mind.

src/main/java/tools/jackson/core/json/ReaderBasedJsonParser.java
seonwooj0810: Done in 4e5945f3. Added `JsonParserBase._checkHexNumbersAllowed(int)` that throws `StreamReadException` naming the feature when `JsonReadFeature.ALLOW_HEXADECIMAL_NUMBERS` is disabled. All eight call sites across the four JSON parsers (Reader, UTF8 stream, UTF8 DataInput, non-blocking UTF8) now go through it, eliminating duplicated `isEnabled(...)` checks. Existing feature-disabled tests tightened to assert the error message includes the feature name.

seonwooj0810:
Quick update batch on this PR:

- **CLA**: signed and sent — should be on file shortly (if not already).
- **Rebased onto 3.x** post-#1614 merge. The `MINOR_NUMBER_PLUSZERO=33` ordinal coordination you flagged is now consistent with upstream (`HEX_PREFIX=34`, `HEX_DIGITS=35`); the renumber commit (`8e5a4b07`) auto-dropped during rebase as its patch was already upstream. Local `mvn test` is clean (1706 tests, 0 failures).
- **Review items from your 03:15–03:28 round**:
  - `ParserBase:401` (`final` comment), `BigIntegerParser:40` (`@since 3.2`) — addressed in 7acfb5cf
  - `JsonParserBase:229` (buffer-boundary length validation) — addressed in d1bf5c22
  - `ReaderBasedJsonParser:1589` (`_checkHexNumbersAllowed` helper) — addressed in 4e5945f3
  - `JsonParserBase:368` (decode in `TextBuffer`) — left a per-thread reply leaning toward keeping this scoped out and revisiting as a follow-up given the PR-size concern; happy to file a tracking issue if you'd prefer.

Ready for another look when you have time.

cowtowncoder:
CLA received.

src/main/java/tools/jackson/core/json/JsonParserBase.java
cowtowncoder: Fair enough.

cowtowncoder (approved):
LGTM: planning to merge later tonight.

seonwooj0810:
Thanks @cowtowncoder and @pjfanning for the thorough review. Moving the hex bookkeeping into `JsonParserBase`, reusing `CharTypes.charToHex()`, and the buffer-boundary length check all made the PR meaningfully better. Really appreciate the time you both put in.

cowtowncoder:
@seonwooj0810 Great collaboration, will get this merged tonight.

PR #1614, opened 2026-05-27
Preserve explicit "+" sign across buffer boundaries for "+0..." on async parser

seonwooj0810:
## Summary

Splits out the drive-by async-parser fix from #1613 per @cowtowncoder's request, so the hex feature PR can stay focused.

The non-blocking parser reused `MINOR_NUMBER_ZERO` for both the bare-zero (`0…`) and explicit-plus-zero (`+0…`) suspension paths, so the leading `+` was lost when input arrived byte-by-byte. The sibling tests in `AsyncNonStandardNumberParsingTest` already disagreed with each other:

- `leadingPlusSignInDecimalEnabled` (`+123`) — asserted `"+123"` ✅
- `leadingPlusSignInDecimalEnabled2` (`+0.123`) — asserted `"0.123"` ❌ (lost the `+`)

This change adds a dedicated `MINOR_NUMBER_PLUSZERO` state alongside `MINOR_NUMBER_ZERO` / `MINOR_NUMBER_MINUSZERO`. `_finishNumberLeadingPosNegZeroes` now stores `MINOR_NUMBER_PLUSZERO` when `negative=false`, so the buffered text on resume retains the leading `+`, matching the blocking parsers and the sibling `+123` path.

`AsyncNonStandardNumberParsingTest#leadingPlusSignInDecimalEnabled2` is updated to assert the now-preserved `"+0.123"` output. The previous `"0.123"` assertion was masked by the bug.

## Background

Discovered while implementing the JSON5 hex feature (#1613) — async tests with 1-byte feeds surfaced the divergence. Per [maintainer comment](https://github.com/FasterXML/jackson-core/pull/1613#issuecomment-4550973163) ("yes, please, that'd be worth fixing separately, can merge quickly"), splitting it out here. Unrelated to the hex feature itself, so easy to review in isolation.

## Test plan

- [x] `mvn test` — 1675 tests, 0 failures, 0 regressions (2 pre-existing skips).
- [x] `AsyncNonStandardNumberParsingTest` — 18 tests pass, including the updated `leadingPlusSignInDecimalEnabled2`.

## Notes

- Diff is +18 / -2 across 3 files.
- CLA: signed copy will follow by email (covers this and #1613).

### DIFF
--- a/src/main/java/tools/jackson/core/json/async/NonBlockingJsonParserBase.java
+++ b/src/main/java/tools/jackson/core/json/async/NonBlockingJsonParserBase.java
@@ -110,6 +110,12 @@ public abstract class NonBlockingJsonParserBase
     protected final static int MINOR_NUMBER_EXPONENT_MARKER = 31;
     protected final static int MINOR_NUMBER_EXPONENT_DIGITS = 32;
 
+    // Resumption state for an explicit '+' followed by '0' (and possibly more
+    // zeros). Companion to {@link #MINOR_NUMBER_ZERO} and
+    // {@link #MINOR_NUMBER_MINUSZERO}; needed so that the buffered text on
+    // resume correctly retains the leading '+' character.
+    protected final static int MINOR_NUMBER_PLUSZERO = 33;
+
     protected final static int MINOR_VALUE_STRING = 40;
     protected final static int MINOR_VALUE_STRING_ESCAPE = 41;
     protected final static int MINOR_VALUE_STRING_UTF8_2 = 42;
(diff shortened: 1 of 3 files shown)


cowtowncoder:
Nice clean split — easy to review in isolation. One small thing worth calling out in the PR description (not a blocker):

`_finishTokenWithEOF` still groups `MINOR_NUMBER_PLUSZERO` with `MINOR_NUMBER_ZERO` / `MINOR_NUMBER_MINUSZERO` and returns `_valueCompleteInt(0, "0")`, so the input `+0<EOF>` resumes as `"0"` (sign lost) — diverging from the blocking parser the same way `+0.123` did before this fix. Looks intentional (bug-compatible with the pre-existing `-0<EOF>` behavior, and the existing inline comment already flags it as deferrable), I'll add a comment.

cowtowncoder (commented):
Will add fix for the `+0<EOF>` / `-0<EOF>` text-loss cases.

PR #1642, opened 2026-07-28
Fix maxDocumentLength bypass in async parser single-feedInput() case [GHSA-2c4j-63jj-9fqr]

revanthmeesala:
Recreated as a normal PR per request in the private advisory fork (GHSA-2c4j-63jj-9fqr / PR FasterXML/jackson-core-ghsa-2c4j-63jj-9fqr#1) — GitHub doesn't support converting/transferring a private security-advisory fork PR into a regular one, so this is the same commit re-opened directly against `2.18`.

## Bug

`NonBlockingJsonParser.feedInput()` and `NonBlockingByteBufferJsonParser.feedInput()` validated document length against `_currInputProcessed` **before** adding the length of the buffer just supplied — `_origBufferLen` still held the previous call's length (`0` on the first call). So the first, and typically only, `feedInput()` call always validated against `0` regardless of the actual buffer size, silently bypassing `maxDocumentLength` whenever a caller feeds an entire pre-buffered document in a single call — the idiomatic pattern once the full payload is already in memory (e.g. right after HTTP body aggregation in reactive-stack integrations).

## Fix

Validate against `_currInputProcessed + (end - start)` instead, so the check reflects the buffer actually being fed — mirroring the eager upfront validation #1549 already added for the synchronous fixed-buffer constructor path.

## Tests

- `largeNameWithSmallLimitAsyncSingleFeed` — feeds the whole over-limit document in one `feedInput()` call for both the `byte[]` and `ByteBuffer` backends; confirms the bypass is closed.
- `largeNameWithSmallLimitAsyncSingleFeedAtBoundary` — feeds a document of exactly `maxDocumentLength` bytes in a single call and asserts it still parses successfully, since `validateDocumentLength` is `len > _maxDocLen` (limit is inclusive). Verified this test actually pins the boundary (not just passing vacuously) by temporarily flipping the check to `len >= maxDocumentLength`, confirming the new test fails against that off-by-one, then reverting.
- Existing `largeNameWithSmallLimitAsync` (multi-call incremental feeding) is unaffected.
- Full suite: 1442/1442 passing, no regressions.

CLA (Individual) signed and sent to cla@fasterxml.com.


cowtowncoder (approved):
LGTM! Thank you @revanthmeesala!!!

PR #1611, opened 2026-05-20
Apply number-length validator on streaming integer path of async parser [GHSA-r7wm-3cxj-wff9]

tonghuaroot:
Aligns the async non-blocking parser's integer-path `NOT_AVAILABLE` exits with the length-validation idiom already used by the sync paths and the async fraction path, so `maxNumberLength` is enforced consistently across sync and async parsers.

The change reuses the existing `_setIntLength` helper at each of the integer-path streaming-suspension points in `NonBlockingUtf8JsonParserBase` (`_startPositiveNumber(int)`, `_startNegativeNumber()`, `_startPositiveNumber()` no-arg, `_finishNumberIntegralPart()`).

### Tests

- New `AsyncNumberLengthConsistencyTest` pins both the integer- and fraction-path streaming behavior under digit-only chunks.
- Full `mvn test`: 1437 tests run, 0 failures, 0 errors, 2 skipped (pre-existing).

Happy to reword / split / amend per maintainer preference.

PR #1643, opened 2026-07-28
Enforce maxNameLength incrementally in ReaderBasedJsonParser [CVE-2026-68498]

tinyb0y:
Fixes GHSA-649p-m576-vr99.

`ReaderBasedJsonParser._parseName2()` only validated accumulated
property-name length against `maxStringLength` while buffering, and
against `maxNameLength` once at the very end (via `findSymbol`).
Reader/String/char[] input could therefore buffer a property name up
to `maxStringLength` before `maxNameLength` ever fired, unlike
byte-based input, which already validates incrementally via
`ParserBase._growNameDecodeBuffer()`.

This tracks accumulated length locally in `_parseName2()` and
validates on each buffer-segment fill, matching the existing
byte-parser idiom (same pattern as #1611 for the analogous
`maxNumberLength` async-parser gap).

### Tests

- New `LargeNameReadTest.largeNameWithSmallLimitCharsFailsFast` pins
  the fix: fails on unpatched code (reports the full buffered length),
  passes on patched code (rejects within one buffer-segment fill).
- Full `mvn test`: 1441 tests run, 0 failures, 0 errors, 2 skipped
  (pre-existing).

### DIFF
--- a/release-notes/CREDITS-2.x
+++ b/release-notes/CREDITS-2.x
@@ -485,7 +485,7 @@ Revanth Meesala (@revanthmeesala)
    case [GHSA-2c4j-63jj-9fqr]
   (2.18.10)
 
-@tinyb0y:
+@tinyb0y
  * Contributed #1643: Enforce maxNameLength incrementally in ReaderBasedJsonParser
    [GHSA-649p-m576-vr99]
   (2.18.10)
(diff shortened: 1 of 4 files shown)


cowtowncoder:
@tinyb0y Sounds good, I will review.

One last (?) thing, if not done already (apologies if I missed it, if so just LMK): we need CLA:

which is usually done by printing, filling/signing, scan/photo, email to `cla` at fasterxml dot com.
Only needed once; once we have it can merge any and all PRs.

Looking forward to merging!

tinyb0y:
@cowtowncoder I've shared the signed CLA via email.

cowtowncoder:
Wow that was fast, @tinyb0y ! Thank you.

PR #1648, opened 2026-08-08
Fix #1640: validate numeric value separator in non-root context for non-blocking parser

### LINKED ISSUE #1640: Non-root number separator validation not applied to non-blocking (async) parser (follow-up to #1557)
Follow-up to #1557 / #1615, as discussed in https://github.com/FasterXML/jackson-core/pull/1615#issuecomment (with @cowtowncoder agreeing async warrants a separate change).

### Background

#1615 generalized the root-only `_verifyRootSpace` into `_verifyNumberSeparator` and now eagerly validates the separator after **non-root** number values, but only in the three **blocking** parsers (`ReaderBasedJsonParser`, `UTF8StreamJsonParser`, `UTF8DataInputJsonParser`).

The **non-blocking / async** parser path (`NonBlockingJsonParserBase` and its subclasses `NonBlockingJsonParser` / `NonBlockingByteArrayJsonParser`) was intentionally left out of that PR and still fails **lazily** — the malformed content is only detected when the next token is read, one token too late.

### Repro

The same inputs from #1557 that are now caught eagerly by the blocking parsers still slip through on the async path:

```
[ 123true ]   // -> reports number token 123, fails only on next token
[ 100k ]      // -> 100
[ 100/ ]      // -> 100
[ 1.5x ]      // -> 1.5
[ 1.5false ]  // -> 1.5
```

### Scope

Apply the equivalent trailing-separator check at the number-completion sites in the non-blocking parser so async parsing reports the error at the offending character, matching the blocking parsers' behavior and location. Async introduces `NEED_MORE_INPUT` boundaries at end-of-chunk, so the check has to tolerate a not-yet-available trailing byte (defer the decision to the next feed rather than reject), which is why this is a separate change from #1615.

Happy to pick this up.

seonwooj0810:
## Summary

Follow-up to #1615 (which fixed the three blocking parsers) and per @cowtowncoder's comment in that PR confirming a separate fix for the non-blocking path is warranted.

The non-blocking (`async`) parser did not verify that numeric values are followed by a valid separator (whitespace, comma, `]`, `}`) when parsed inside an Array or Object context, silently accepting malformed inputs that blocking parsers rejected:

```java
// Blocking parsers (fixed in #1615): correctly throw
// Async parser (this PR): previously returned VALUE_NUMBER_INT
JsonParser p = factory.createNonBlockingByteArrayParser(...);
feeder.feedInput(new byte[]{'[','1','2','3','t','r','u','e',']'}, 0, 9);
feeder.endOfInput();
p.nextToken(); // START_ARRAY
p.nextToken(); // was VALUE_NUMBER_INT, should throw StreamReadException
```

## Root cause

Five integer-completion sites and three float-completion sites in `NonBlockingUtf8JsonParserBase` called `_valueComplete(JsonToken.VALUE_NUMBER_*)` without a separator check:

- `_startPositiveNumber(int ch)` — inline digit loop
- `_startNegativeNumber()` — inline digit loop
- `_startPositiveNumber()` — inline digit loop (+ sign variant)
- `_finishNumberIntegralPart(...)` — resumption path (chunk boundary)
- Hex number completion (via `resetIntHex`)
- `_startFloat(...)` — inline fraction/exponent loop
- `_finishFloatFraction()` — resumption path (fraction digits)
- `_finishFloatExponent(...)` — resumption path (exponent digits)

## Fix

Adds `_verifyNumberSeparator(int ch)` to `NonBlockingUtf8JsonParserBase`, mirroring the same method added to the three blocking parsers in #1615. For root context it delegates to the existing `_verifyRootSpace(ch)`; for non-root context it validates the separator is one of ` \t\r\n,]}` (or a comment-start character if the respective feature is enabled), consistent with the blocking-parser behavior.

All 8 completion sites now call `_verifyNumberSeparator(ch)` before `_valueComplete(...)`.

## Testing

- New test class `AsyncNumberSeparator1640Test` covering int, float-fraction, and float-exponent mangled cases across `bytesPerRead` values of 90, 3, and 1 (to exercise both inline and resumption paths).
- `AsyncTokenBranchNumberErrorTest.mangledNonRootInts` in `tofix/` now passes; removed its `@JacksonTestFailureExpected` annotation. The `mangledNonRootFloats` test remains in `tofix/` (the 'f'/'d' suffix case still uses a dedicated code path with a different error message).
- Full test suite: 1813 tests, 0 failures, 0 errors.

Fixes #1640

### DIFF
--- a/release-notes/VERSION
+++ b/release-notes/VERSION
@@ -17,6 +17,10 @@ JSON library.
 
 3.3.0 (not yet released)
 
+#1640: Non-blocking (async) parser does not validate numeric value separator
+  in non-root context (Array/Object), unlike blocking parsers
+ (reported by @seonwooj0810)
+ (fix by @seonwooj0810)
 #1637: Add `JsonPointer.startsWith(JsonPointer)` to check prefix match
  (contributed by @scottslewis)
 #1646: Non-blocking ByteBuffer parser `releaseBuffered()` can write
--- a/src/main/java/tools/jackson/core/json/async/NonBlockingUtf8JsonParserBase.java
+++ b/src/main/java/tools/jackson/core/json/async/NonBlockingUtf8JsonParserBase.java
@@ -1355,6 +1355,8 @@ public abstract class NonBlockingUtf8JsonParserBase
         }
         _setIntLength(outPtr);
         _textBuffer.setCurrentLength(outPtr);
+        // 08-Aug-2026, tatu: [core#1640] Verify separator in both root and non-root context
+        _verifyNumberSeparator(ch);
         return _valueComplete(JsonToken.VALUE_NUMBER_INT);
     }
 
@@ -1422,6 +1424,8 @@ public abstract class NonBlockingUtf8JsonParserBase
         }
         _setIntLength(outPtr-1);
         _textBuffer.setCurrentLength(outPtr);
+        // 08-Aug-2026, tatu: [core#1640] Verify separator in both root and non-root context
+        _verifyNumberSeparator(ch & 0xFF);
         return _valueComplete(JsonToken.VALUE_NUMBER_INT);
     }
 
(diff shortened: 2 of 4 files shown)


seonwooj0810:
Circling back on this — it's the scoped non-blocking-parser follow-up we agreed to split out on #1615/#1640. Whenever you get a chance to take a look, happy to adjust anything.

cowtowncoder:
@seonwooj0810 Unfortunately I am severely behind so this will take some time. Thank you!

seonwooj0810:
No worries at all, thanks for the heads-up @cowtowncoder! Happy to wait, and just let me know if anything needs adjusting.

PR #1651, opened 2026-08-10
Fix object context handling for buffered `INCLUDE_NON_NULL` tokens

Dongnyoung:
## Summary

Fixes a `FilteringParserDelegate` buffering path that handled a `START_OBJECT`
token by creating an array context when using `TokenFilter.Inclusion.INCLUDE_NON_NULL`.

This could leave the exposed parser context inconsistent with the current token:
the parser returned `START_OBJECT`, but `streamReadContext()` reported an array
context. In nested object cases this could also lead to an internal buffering
error.

## Changes

- Add regression coverage for `INCLUDE_NON_NULL` with a nested object.
- Change the buffered `START_OBJECT` / `INCLUDE_NON_NULL` path to create an
  object context instead of an array context.

## Testing

```powershell
.\mvnw.cmd "-Dtest=BasicParserFilteringTest,ParserFilterEmpty1418Test,ParserFilterEmpty708Test,ParserFiltering700Test" "-Dsurefire.useModulePath=false" test
Result: 59 tests passed.

### DIFF
--- a/release-notes/CREDITS
+++ b/release-notes/CREDITS
@@ -54,3 +54,7 @@ Revanth Meesala (@revanthmeesala)
  * Contributed #1643: Enforce maxNameLength incrementally in ReaderBasedJsonParser
    [GHSA-649p-m576-vr99]
   (3.1.6)
+
+Lee Dongnyoung (@DongNyoung)
+ * Contributed #1651: Fix object context handling for buffered `INCLUDE_NON_NULL` tokens
+  (3.1.6)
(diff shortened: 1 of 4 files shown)


cowtowncoder:
Makes sense -- but could you change target branch to be 3.1?

Dongnyoung:
@cowtowncoder Thanks! I’ve updated the target branch to 3.1 and rebased the changes accordingly.

cowtowncoder:
Fixed in 3.1, backported in 2.21; will be in:

* 2.21.6, 2.22.2
* 3.1.6, 3.2.2

PR #1655, opened 2026-08-12
Prevent stale symbol table children from replacing newer root tables

### LINKED ISSUE #1654: Stale symbol table child can replace a newer root table
### Description

A stale child symbol table can replace a newer, larger root symbol table when parsers are closed out of creation order.

For example, if one parser is created and reads one field name, and another parser created from the same `JsonFactory` reads additional field names and closes first, the root symbol table is updated with the larger child table. When the older parser closes later, however, its smaller child table can replace the newer root table.

As a result, the shared symbol table can lose symbols learned by a newer parser.

### Reproduction

The issue applies to both byte-backed and char-backed parsers:

1. Create parser A from a shared `JsonFactory`.
2. Read one field name with parser A.
3. Create parser B from the same factory.
4. Read additional field names with parser B.
5. Close parser B.
6. Close parser A.

After parser B closes, the root symbol table contains the larger symbol set. After parser A closes, the root table can regress to parser A's stale, smaller child table.

### Expected behavior

An older child symbol table should not replace a newer, larger root table produced by another child.

### Actual behavior

The older child can merge after the newer child and replace the root table with its stale state.

### Notes

This affects both `CharsToNameCanonicalizer` and `ByteQuadsCanonicalizer`.

This is separate from the intentional reset behavior for oversized symbol tables. The issue here is specifically an older child state replacing a newer root state during out-of-order parser release.

Dongnyoung:
Fixes #1654 
### Summary

Prevent stale child symbol tables from replacing newer, larger root symbol tables when parsers are closed out of creation order.

Previously, `mergeChild()` compared the child state against the root state observed at release time. This made the update atomic, but did not detect that the child itself could have been created from an older parent state. As a result, an older, smaller child could replace a root table that had already been updated by another child.

Child symbol tables now retain the parent table state they were created from. During release, a stale child no longer replaces the root if another child has already advanced it to a newer, larger state.

This applies to both `CharsToNameCanonicalizer` and `ByteQuadsCanonicalizer`. The existing reset behavior for oversized symbol tables is preserved.

### Tests

Added coverage to `SymbolTableMergingTest` for out-of-order stale child release with:

* byte-backed parsers
* char-backed parsers

Ran:

* `.\mvnw.cmd -Dtest=SymbolTableMergingTest '-Dsurefire.useModulePath=false' test`
* `.\mvnw.cmd '-Dtest=SymbolTableMergingTest,TestSymbolTables,TestByteBasedSymbols,TestHashCollisionChars,PlaceholderSymbolTableTest,TestSymbolsWithMediaItem' '-Dsurefire.useModulePath=false' test`

### DIFF
--- a/release-notes/CREDITS
+++ b/release-notes/CREDITS
@@ -67,6 +67,8 @@ DongNyoung Lee (@Dongnyoung)
  * Reported, fixed #1664: UTF-8 byte-backed parsers report raw byte instead of
    decoded character for invalid root-value separator
   (3.3.0)
+ * Reported #1654: Stale symbol table child can replace a newer root table
+  (3.3.0)
 
 seonwoojung (@seonwooj0810)
  * Contributed #707: Add `JsonReadFeature.ALLOW_HEXADECIMAL_NUMBERS`
--- a/release-notes/VERSION
+++ b/release-notes/VERSION
@@ -22,6 +22,8 @@ JSON library.
 #1646: Non-blocking ByteBuffer parser `releaseBuffered()` can write
   already-consumed bytes
  (reported, fix by DongNyoung L)
+#1654: Stale symbol table child can replace a newer root table
+ (reported by DongNyoung L)
 #1664: UTF-8 byte-backed parsers report raw byte instead of decoded character
   for invalid root-value separator
  (reported, fix by DongNyoung L)
--- a/src/main/java/tools/jackson/core/sym/ByteQuadsCanonicalizer.java
+++ b/src/main/java/tools/jackson/core/sym/ByteQuadsCanonicalizer.java
@@ -76,7 +76,7 @@ public class ByteQuadsCanonicalizer
      */
     protected final AtomicReference<TableInfo> _tableInfo;
 
-    private final TableInfo _parentTableInfo;
+    private final Object _parentTableInfoId;
 
     /**
      * Seed value we use as the base to make hash codes non-static between
@@ -225,7 +225,7 @@ public class ByteQuadsCanonicalizer
     {
         // Settings to distinguish parent table: no parent
         _parent = null;
-        _parentTableInfo = null;
+        _parentTableInfoId = null;
         _count = 0;
 
         // and mark as shared just in case to prevent modifications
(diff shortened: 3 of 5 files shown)


cowtowncoder:
I don't know about this one -- is there an actual real problem resolved here?

Dongnyoung:
@cowtowncoder 
Fair question. I measured this as a symbol-table cache/performance regression rather than a parsing correctness issue.

The regression is reproducible through normal `JsonParser` usage: parser A reads one field name, parser B created from the same `JsonFactory` reads a larger set and closes first, and then A closes. Before the fix, A's stale child can replace the newer root table.

With a synthetic document containing 1,024 unique field names, I observed:
```
pre-fix:
bytes: rootSize 1024 -> 1
chars: rootSize 1024 -> 1

fixed:
bytes: rootSize remains 1024
chars: rootSize remains 1024
```
I also ran a small JMH benchmark with `-prof gc`, measuring the next parse of the same document after restoring either the baseline or regressed root state before each parse:
```
pre-fix:
bytes baseline: 18.5 us/op,     664 B/op
bytes stale:    51.7 us/op, 211,681 B/op

chars baseline: 24.1 us/op,     568 B/op
chars stale:    35.4 us/op,  87,072 B/op

fixed:
bytes baseline: 18.6 us/op, 664 B/op
bytes stale:    18.0 us/op, 664 B/op

chars baseline: 24.4 us/op, 576 B/op
chars stale:    24.4 us/op, 576 B/op
```
So in this synthetic workload, the stale root causes the following parse to re-canonicalize field names that had already been learned, with a clear allocation and parsing-time cost. The fix prevents that backwards transition.

I wouldn't characterize this as a parsing correctness bug, but the cache regression is measurable and can result from normal parser lifecycle.

cowtowncoder:
@Dongnyoung  Isn't there concern that `TableInfo _parentTableInfo` might lead to unintentional memory retention via parent chain?

Dongnyoung:
@cowtowncoder Good point. There isn't a `TableInfo` parent chain since `TableInfo` itself doesn't reference previous states, but retaining `_parentTableInfo` did keep the old table's backing arrays and symbols alive for the lifetime of the child/parser.

I changed this to retain only a lightweight identity token for the parent state instead. Each `TableInfo` now has a unique token, and the child keeps only that token. This preserves the stale-state identity check without retaining the full parent `TableInfo`.

The size/count comparisons and oversized-table reset behavior remain unchanged.

cowtowncoder (approved):

Issue #1652, opened 2026-08-11
No-allocation DoubleToDecimal / FloatToDecimal


westinrm:
OpenJDK25 added implementations of `DoubleToDecimal` and `FloatToDecimal` that encode directly to Latin1 bytes without any allocations: https://github.com/openjdk/jdk/commit/9d20b58f40275002afa0348d94d5592a26894e88

If this was public API, it could be used pretty easily within UTF8JsonGenerator for some fairly massive performance benefits when serializing large numbers of floating point numbers -- something that's very common with GeoJSON.
```java
@Override
public void writeNumber(double d) throws IOException {
    if (this._outputTail + DoubleToDecimal.MAX_CHARS >= this._outputEnd) {
        this._flushBuffer();
    }
    this._outputTail = DoubleToDecimal.LATIN1.putDecimal(this._outputBuffer, this._outputTail, d);
}
```

A simple benchmark that tests this change by writing a 10k element float/double array to a null output stream shows that this approach is nearly 3x as fast, even when the relevant internal utility is copied out of the JDK (removing the requirement of `--add-opens`). 
```
Benchmark                                         (type)        (mode)   Mode  Cnt        Score     Error   Units
WriteFloatingPointArray.write                      FLOAT       VANILLA  thrpt    4     1934.454 ±  84.062   ops/s
WriteFloatingPointArray.write:gc.alloc.rate.norm   FLOAT       VANILLA  thrpt    4   880296.792 ±   0.825    B/op
WriteFloatingPointArray.write                      FLOAT  JACKSON_FAST  thrpt    4     2116.445 ±  52.042   ops/s
WriteFloatingPointArray.write:gc.alloc.rate.norm   FLOAT  JACKSON_FAST  thrpt    4   880263.591 ± 248.715    B/op
WriteFloatingPointArray.write                      FLOAT  JDK_INTERNAL  thrpt    4     3708.587 ± 491.341   ops/s
WriteFloatingPointArray.write:gc.alloc.rate.norm   FLOAT  JDK_INTERNAL  thrpt    4      383.432 ±  12.262    B/op
WriteFloatingPointArray.write                      FLOAT      JDK_COPY  thrpt    4     3623.090 ± 240.293   ops/s
WriteFloatingPointArray.write:gc.alloc.rate.norm   FLOAT      JDK_COPY  thrpt    4      383.387 ±  13.224    B/op
[...]

PR #1656, opened 2026-08-12
Use XJB algorithm for writing Floats/Doubles

pjfanning:
* See #1652 
* Aimed as a replacement for the Schubfach code
* One of the main benefits is that the XJBWriter can avoid String allocations if you provide the byte or char array to write to
* benchmarked in https://github.com/pjfanning/double-reader-writer - similar toString as Schubfach but the fact that you can get it to write directly to a byte array makes it a very useful addition for the Utf8JsonGenerator - this saves on String and array allocations and makes a big contribution to the performance improvement
* has been enhanced to support writing directly to char arrays too
* Android doesn't support VarHandles so we end up with some messy Reflection. For me the better solution is to take advantage of the fact that jackson-core is already a Multi-Release Jar because FastDoubleParser embedded code needs that. Android ignores the classes in META-INF/versions, so we could put the VarHandle supporting code in a class in META-INF/versions/17. JVM users will get to use that version while Android users will use the fallback version that is in the main code.


pjfanning:
@plokhotnyuk feel free to ignore this but if you time, it would be nice to get your input on the XJBWriter. I added support for writing to char arrays as well as byte arrays because this is useful in some Jackson code paths. The char stream code may not be as heavily optimised as the byte stream path.

plokhotnyuk:
> @plokhotnyuk feel free to ignore this but if you time, it would be nice to get your input on the XJBWriter. I added support for writing to char arrays as well as byte arrays because this is useful in some Jackson code paths. The char stream code may not be as heavily optimized as the byte stream path.

Great work! Have you used AI agents or translated it manually to Java?

The next step could be using James Anhalt's algorithm for faster writing of mantissas: https://jk-jeon.github.io/posts/2022/02/jeaiii-algorithm/

Or you can leverage 3 digit LUT table to reduce cycles as it is done for integer numbers.

pjfanning:
@plokhotnyuk mainly AI agent work but it was pretty iterative so possibly not a great time saver in this case.

pjfanning:
@cowtowncoder if this does get merged in time for 3.3.0, the CSV module could also benefit by uptaking the new NumberOutput APIs that this PR adds.

cowtowncoder:
@pjfanning Hoping to get to this next.

cowtowncoder:
Hmmmh. I have some concerns based on Claude-assisted code review.

First is LICENSE:

```
  1. License and NOTICE
     - META-INF/jackson-core-NOTICE still describes Schubfach, and Schubfach-LICENSE is still bundled even though that code is removed.
     - There is no license text or NOTICE entry for jsoniter-scala (MIT, Andriy Plokhotnyuk), which the code is ported from. MIT requires shipping its copyright and permission notice.
     - If any of it traces back to the xjb reference code (Apache 2.0), that license and notice need to ship too.
```
 
which may be easy to address. Will add other notes too, but first things first.

cowtowncoder:
Second: does XJB require use of VarHandles? If not, maybe first version could just use added `ByteArrayUtil` access methods, and we could consider move to VarHandles (and class-loading issues) separately?

HOWEVER, also: Claude claims:

```
 Tests pass and the speed regression is gone with direct VarHandle usage, but animal-sniffer flags the
  signature-polymorphic set/get calls as a false positive since VarHandle itself is available in Android 33+.
  I'll add a narrow ignore rule scoped just to java.lang.invoke.VarHandle, without excluding source or ignoring
  MethodHandle. 

...

 VarHandle is part of the Android
    API from 33, but animal-sniffer can't resolve calls like VarHandle.set(byte[], int, int) and reports them as missing.
```

pjfanning:
@cowtowncoder AnimalSniffer failures are what sent me down the route of using indirection for the VarHandles. The docs that I found suggest that Android should support VarHandles at the version we support but AnimalSniffer doesn't appear to agree.

pjfanning:
> Hmmmh. I have some concerns based on Claude-assisted code review.
> 
> First is LICENSE:
> 
> ```
>   1. License and NOTICE
>      - META-INF/jackson-core-NOTICE still describes Schubfach, and Schubfach-LICENSE is still bundled even though that code is removed.
>      - There is no license text or NOTICE entry for jsoniter-scala (MIT, Andriy Plokhotnyuk), which the code is ported from. MIT requires shipping its copyright and permission notice.
>      - If any of it traces back to the xjb reference code (Apache 2.0), that license and notice need to ship too.
> ```
> 
> which may be easy to address. Will add other notes too, but first things first.

@cowtowncoder I updated the licenses in META-INF - c72991178504bba5df9581c54695553ed432fa4d

pjfanning:
@cowtowncoder I'd prefer to keep the VarHandles as they really help the performance. For instance, we make use of them in the binary dataformats now.
I updated the PR to use ByteArrayUtil for the fallback when VarHandles are not available in the JRE.
b5d5d611d48ad5bfbb9fc87c739aa2432aa981b9

cowtowncoder:
@pjfanning Ok. Yes, I know about binary (and some textual?) backends wrt recent PRs.

One last concern code review mentioned: some of serializations change for numbers -- not in wrong way (they still round-trip), but differ from what JDK produces.
Specifically for something called "smallest subnormals".

```
 With the feature enabled, comparing XJB against the Schubfach code it replaces, I found 17 differing values
  out of 40M random plus targeted ranges. All of them are the smallest subnormals:

  ┌──────────────────┬──────────────────────────────────────────┬──────────────────────────────────────────┐
  │       Bits       │                XJB (new)                 │             Schubfach (3.2)              │
  ├──────────────────┼──────────────────────────────────────────┼──────────────────────────────────────────┤
  │ double 0x1 …     │ 5.0E-324, 1.0E-323, 5.0E-323, 6.0E-323,  │ 4.9E-324, 9.9E-324, 4.9E-323, 5.9E-323,  │
  │ 0x14 (8 values)  │ 7.0E-323, 8.0E-323, 9.0E-323, 1.0E-322   │ 6.9E-323, 7.9E-323, 8.9E-323, 9.9E-323   │
  ├──────────────────┼──────────────────────────────────────────┼──────────────────────────────────────────┤
  │ float 0x1 … 0x47 │ 1.0E-45, 3.0E-45, 4.0E-45, 6.0E-45,      │ 1.4E-45, 2.8E-45, 4.2E-45, 5.6E-45,      │
  │  (9 values)      │ 8.0E-45, 1.0E-44, 3.0E-44, 4.0E-44,      │ 8.4E-45, 9.8E-45, 2.9E-44, 4.1E-44,      │
  │                  │ 1.0E-43                                  │ 9.9E-44                                  │
  └──────────────────┴──────────────────────────────────────────┴──────────────────────────────────────────┘

  The cause is that XJB emits the shortest decimal that round-trips, while the JDK and Schubfach add a second
  digit when the shortest has only one.

```

This may be fine, just wanted to mention; at least it can trigger test failures downstream.

pjfanning:
@cowtowncoder The writer is opt-in but I understand the worry about the different output for some numbers.

@plokhotnyuk what do you think about the issue described in https://github.com/FasterXML/jackson-core/pull/1656#issuecomment-5641077116 ?

plokhotnyuk:
> @plokhotnyuk what do you think about the issue described in [#1656 (comment)](https://github.com/FasterXML/jackson-core/pull/1656#issuecomment-5641077116) ?

Originally XJB formats without dot when mantissa is just one digit (for instance 1e-45 - the shortest representation). 

But for Java there is a formatting rule that dot and at least one digit after are required, so that why 1.0E-45 reappears like on 17 JDK.

cowtowncoder:
@pjfanning Yes, opt-in is probably sufficient to allow change.

PR #1657, opened 2026-08-12
UTF8JsonGenerator: writeNumber fast path that avoids extra String allocation

pjfanning:
Partially related to #1656 
The new XJBWriter has support for writing the number to a pre-existing byte array and avoid String allocations.
It's possible that 1656 will get nowhere but I thought that it might be worth considering this change as a completely separate change.
I modified the existing Schubfach code to allow String allocation to be skipped when writing numbers in UTF8JsonGenerator.
And if we do switch to XJB, the NumberOutput class can be modified to use XJB instead of Schubfach but this PR tests UTF8JsonGenerator in isolation.

This change only kicks in if you enable StreamWriteFeature.USE_FAST_DOUBLE_WRITER

### DIFF
--- a/release-notes/VERSION
+++ b/release-notes/VERSION
@@ -28,6 +28,8 @@ JSON library.
  (fix by @tinyb0y)
 #1649: Optimize `NumberInput.looksLikeValidNumber`
  (fix by @cowtowncoder, w/ Claude code)
+#1657: UTF8JsonGenerator: writeNumber fast path that avoids extra String allocation
+ (implemented by @pjfanning)
 
 3.2.1 (10-Jul-2026)
 
--- a/src/main/java/tools/jackson/core/io/NumberOutput.java
+++ b/src/main/java/tools/jackson/core/io/NumberOutput.java
@@ -286,6 +286,40 @@ public final class NumberOutput
         return useFastWriter ? FloatToDecimal.toString(v) : Float.toString(v);
     }
 
+    /**
+     * Direct-to-buffer write for {@code float} values, bypassing String allocation.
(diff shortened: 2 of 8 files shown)


pjfanning:
@cowtowncoder this change gives quite a boost when using Utf8JsonGenerator with USE_FAST_DOUBLE_WRITER enabled. I'll put the numbers in https://github.com/pjfanning/double-reader-writer readme.
But basically,
```
Benchmark                                         Mode  Cnt    Score    Error   Units
JsonGeneratorBenchmark.baselineWriteDoubleArray  thrpt    3   93.082 ± 38.279  ops/ms
JsonGeneratorBenchmark.pr1657WriteDoubleArray    thrpt    3  146.721 ± 78.125  ops/ms
JsonGeneratorBenchmark.xjbWriteDoubleArray       thrpt    3  189.703 ± 59.990  ops/ms
```
Where baselineWriteDoubleArray is existing jackson-core, pr1657WriteDoubleArray is with this change and xjbWriteDoubleArray is with this change plus the XJB Writer from #1656.

pjfanning:
@cowtowncoder I think this change is small enough that it might be worth considering including it in the next 3.2 release or even the nest 3.1 LTS release. There is an argument that the related PR1656 is too big to add until 3.3.0.

cowtowncoder:
@pjfanning If you can rebase against 3.2, sure -- I am now releasing 3.1.6 so maybe not 3.1.

cowtowncoder:
Needs `release-notes/VERSION` entry.

**EDIT**: added.

pjfanning:
Retargeted PR to 3.2 branch

cowtowncoder (approved):

PR #1659, opened 2026-08-13
modernise code - Java 17 support

pjfanning:
Now that we have Java 17 as a min supported version

### DIFF
--- a/src/main/java/tools/jackson/core/JacksonException.java
+++ b/src/main/java/tools/jackson/core/JacksonException.java
@@ -332,8 +332,8 @@ public class JacksonException
         if (t instanceof JacksonException jex) {
             return jex.getOriginalMessage();
         }
-        if (t instanceof InvocationTargetException && t.getCause() != null) {
-            return t.getCause().getMessage();
+        if (t instanceof InvocationTargetException ite && ite.getCause() != null) {
+            return ite.getCause().getMessage();
         }
         return t.getMessage();
     }
@@ -387,7 +387,7 @@ public class JacksonException
     public List<Reference> getPath()
     {
         if (_path == null) {
-            return Collections.emptyList();
+            return List.of();
         }
         return Collections.unmodifiableList(_path);
     }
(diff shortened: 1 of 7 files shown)


src/main/java/tools/jackson/core/JacksonException.java
```diff
@@ -332,8 +332,8 @@ public static String _exceptionMessage(Throwable t) {
         if (t instanceof JacksonException jex) {
             return jex.getOriginalMessage();
         }
-        if (t instanceof InvocationTargetException && t.getCause() != null) {
-            return t.getCause().getMessage();
+        if (t instanceof InvocationTargetException ite && ite.getCause() != null) {
```
cowtowncoder: Why? There is no benefit to casting here.
(auto-generated I'm sure)

src/main/java/tools/jackson/core/JacksonException.java
```diff
@@ -387,7 +387,7 @@ public JacksonException prependPath(Reference r)
     public List<Reference> getPath()
     {
         if (_path == null) {
-            return Collections.emptyList();
+            return List.of();
```
cowtowncoder: .... why?

src/main/java/tools/jackson/core/JsonGenerator.java
```diff
@@ -2365,8 +2365,8 @@ protected void _copyCurrentFloatValueExact(JsonParser p) throws JacksonException
         Number n = p.getNumberValueExact();
         if (n instanceof BigDecimal bigDecimal) {
             writeNumber(bigDecimal);
-        } else if (n instanceof Double) {
-            writeNumber(n.doubleValue());
+        } else if (n instanceof Double d) {
```
cowtowncoder: Not needed/useful

cowtowncoder:
Half of casts not useful, otherwise good.

src/main/java/tools/jackson/core/JacksonException.java
pjfanning: autogenerated - I'm neutral about this, AI suggested it had no cost and that it was easier to read - I can revert if preferred

cowtowncoder: UNderstood. I like minimalism so yeah let's revert.

cowtowncoder (approved):

pjfanning:
@cowtowncoder I reverted all the changes other than the switch changes

PR #1662, opened 2026-08-14
tests NaN writing support when USE_FAST_DOUBLE_WRITER is enabled

pjfanning:
Copies a test used with default setup but I think it is important to see if USE_FAST_DOUBLE_WRITER has an effect.

### DIFF
--- a/src/test/java/tools/jackson/core/unittest/json/StreamWriteFeaturesTest.java
+++ b/src/test/java/tools/jackson/core/unittest/json/StreamWriteFeaturesTest.java
@@ -59,19 +59,7 @@ class GeneratorFeaturesTest
     @Test
     void nonNumericQuoting() throws IOException
     {
-        JsonFactory f = new JsonFactory();
-        // by default, quoting should be enabled
-        assertTrue(f.isEnabled(JsonWriteFeature.WRITE_NAN_AS_STRINGS));
-        _testNonNumericQuoting(f, true);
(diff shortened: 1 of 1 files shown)


cowtowncoder (approved):

src/test/java/tools/jackson/core/unittest/json/StreamWriteFeaturesTest.java
```diff
@@ -59,19 +59,16 @@ void fieldNameQuoting() throws IOException
...
-        assertTrue(f.isEnabled(JsonWriteFeature.WRITE_NAN_AS_STRINGS));
-        _testNonNumericQuoting(f, true);
-        // can disable it
-        f = f.rebuild().disable(JsonWriteFeature.WRITE_NAN_AS_STRINGS)
-                .build();
-        _testNonNumericQuoting(f, false);
-        // and (re)enable:
-        f = f.rebuild()
-                .enable(JsonWriteFeature.WRITE_NAN_AS_STRINGS)
+        _testNonNumericQuoting(new JsonFactory());
```
cowtowncoder: Should probably not rely on default settings but explicitly disable setting here too?

pjfanning: changed

cowtowncoder (approved):

PR #1665, opened 2026-08-15
Fix UTF-8 decoding for root separator error reporting

### LINKED ISSUE #1664: UTF-8 byte-backed parsers report raw byte instead of decoded character for invalid root-value separator
When a UTF-8 byte-backed parser reports an invalid character after a root-level numeric value, the error message may describe the first UTF-8 byte instead of the decoded Unicode code point.

For example, input equivalent to:

```java
byte[] json = new byte[] { '1', (byte) 0xC2, (byte) 0xA0 };
```

contains `U+00A0` after the root number. This should be reported as code `160`, but the `InputStream` parser currently reports code `194`, which is the first byte of the UTF-8 sequence.

This seems to happen in root-value separator validation. `UTF8StreamJsonParser` already has `_decodeCharForError(...)`, but `_verifyRootSpace(...)` does not use it before reporting the unexpected character. The same path should be checked for `UTF8DataInputJsonParser`.

**Expected:** The unexpected character should be reported using the decoded UTF-8 code point, consistently with other UTF-8 error-reporting paths.

**Actual:** The error may report the raw leading byte of the UTF-8 sequence.

Tested with:

```text
.\mvnw.cmd -q '-Dtest=tools.jackson.core.unittest.read.NonStandardNumberParsingTest' '-Dsurefire.useModulePath=false' test
```

Dongnyoung:
Fixes: #1664 
Fixes root-value separator error reporting for UTF-8 byte-backed parsers.

When an invalid non-ASCII UTF-8 character appears after a root-level number, `UTF8StreamJsonParser` reported the leading UTF-8 byte instead of the decoded character. `UTF8DataInputJsonParser` had the same root-space validation path.

This change decodes non-ASCII bytes with `_decodeCharForError(...)` before reporting the missing root whitespace error, and adds coverage for `InputStream`, throttled `InputStream`, and `DataInput` modes.

Tested with:

```bash
.\mvnw.cmd -q '-Dtest=tools.jackson.core.unittest.read.NonStandardNumberParsingTest' '-Dsurefire.useModulePath=false' test

### DIFF
--- a/src/main/java/tools/jackson/core/json/UTF8DataInputJsonParser.java
+++ b/src/main/java/tools/jackson/core/json/UTF8DataInputJsonParser.java
@@ -1332,7 +1332,7 @@ public class UTF8DataInputJsonParser
      * If we did want, we could rearrange things to require space before
      * next read, but initially let's just do nothing.
      */
-    private final void _verifyRootSpace() throws IOException
+    private final void _verifyRootSpace() throws JacksonException
     {
         int ch = _nextByte;
         if (ch <= INT_SPACE) {
@@ -1343,7 +1343,11 @@ public class UTF8DataInputJsonParser
             return;
         }
         if (ch > 0x7F) {
-            ch = _decodeCharForError(ch);
+            try {
+                ch = _decodeCharForError(ch);
+            } catch (IOException e) {
+                throw _wrapIOFailure(e);
+            }
         }
         _reportMissingRootWS(ch);
     }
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/main/java/tools/jackson/core/json/UTF8DataInputJsonParser.java [hunk 1]
```diff
@@ -1332,7 +1332,7 @@ private final JsonToken _parseFloat(char[] outBuf, int outPtr, int c,
      * If we did want, we could rearrange things to require space before
      * next read, but initially let's just do nothing.
      */
-    private final void _verifyRootSpace() throws JacksonException
+    private final void _verifyRootSpace() throws IOException
```
cowtowncoder: No, why should we change that ?!

Dongnyoung: Oh! Good point. I reverted the signature change and kept `_verifyRootSpace()` throwing `JacksonException`. I wrapped the `IOException` from UTF-8 decoding via `_wrapIOFailure(e)` instead.

PR #1667, opened 2026-08-18
Fix `willInternPropertyNames()` when canonicalization is disabled

Dongnyoung:
### Problem

`CharsToNameCanonicalizer.willInternStrings()` currently checks only
`INTERN_PROPERTY_NAMES`.

When `INTERN_PROPERTY_NAMES` is enabled but
`CANONICALIZE_PROPERTY_NAMES` is disabled, `findSymbol()` bypasses
`_addSymbol()` and returns `new String(buffer, start, len)` directly.
That means property names are not actually interned, even though
`JsonParser.willInternPropertyNames()` reports `true`.

This is also consistent with the feature documentation:
`INTERN_PROPERTY_NAMES` only has effect when
`CANONICALIZE_PROPERTY_NAMES` is enabled.

### Fix

Include `_canonicalize` in `CharsToNameCanonicalizer.willInternStrings()`.

Also update `InternPropertyNamesTest` to cover the
`intern=true, canonicalize=false` case.

### Tests

```powershell
.\mvnw.cmd "-Dtest=InternPropertyNamesTest,SimpleParserTest,JsonFactoryTest" "-Dsurefire.useModulePath=false" test


cowtowncoder:
Sorry, I broke the PR, trying to get Claude to target 3.1; will need to re-create

cowtowncoder:
Re-created as #1670, so closing this one in favor of that.

PR #1669, opened 2026-08-18
Fix WriterBasedJsonGenerator AIOOBE for zero-length custom escape

### LINKED ISSUE #1668: `WriterBasedJsonGenerator` throws `ArrayIndexOutOfBoundsException` when a zero-length custom escape ends exactly at the output buffer boundary
### Description
When a CharacterEscapes implementation returns a zero-length escape sequence, `WriterBasedJsonGenerator.writeString(String)` throws:
```
java.lang.ArrayIndexOutOfBoundsException: Index 4000 out of bounds for length 4000
        at tools.jackson.core.json.WriterBasedJsonGenerator.writeString(WriterBasedJsonGenerator.java:461)
```
It happens only when the custom-escaped character is the last character of the string being written.

### Reproduction
```
  import tools.jackson.core.*;                                                                                                                                                                                   
  import tools.jackson.core.io.CharacterEscapes;                                                                                                                                                                 
  import tools.jackson.core.io.SerializedString;                                                                                                                                                                 
  import tools.jackson.core.json.JsonFactory;                                                                                                                                                                    
  import java.io.StringWriter;                                                                                                                                                                                   
                                                                                                                                                                                                                 
  public class Repro {                                                                                                                                                                                           
      static class ZeroLengthNulEscape extends CharacterEscapes {                                                                                                                                                
          private static final SerializableString EMPTY = new SerializedString("");                                                                                                                              
          private final int[] escapes;                                                                                                                                                                           
          ZeroLengthNulEscape() {                                                                                                                                                                                
              escapes = standardAsciiEscapesForJSON();                                                                                                                                                           
              escapes[0] = ESCAPE_CUSTOM;                                                                                                                                                                        
          }                                                                                                                                                                                                      
          @Override public int[] getEscapeCodesForAscii() { return escapes; }                                                                                                                                    
          @Override public SerializableString getEscapeSequence(int ch) { return ch == 0 ? EMPTY : null; }                                                                                                       
      }                                                                                                                                                                                                          
                                                                                                                                                                                                                 
      public static void main(String[] args) throws Exception {                                                                                                                                                  
          JsonFactory factory = JsonFactory.builder()                                                                                                                                                            
                  .characterEscapes(new ZeroLengthNulEscape())                                                                                                                                                   
                  .build();                                                                                                                                                                                      
          String value = "Istio LFS144" + '\0';   // last char is the custom-escaped one                                                                                                                         
          for (int pad = 3900; pad <= 4100; pad++) {                                                                                                                                                             
              StringWriter w = new StringWriter();                                                                                                                                                               
              try (JsonGenerator gen = factory.createGenerator(ObjectWriteContext.empty(), w)) {                                                                                                                 
                  gen.writeStartArray();                                                                                                                                                                         
                  gen.writeString("x".repeat(pad));                                                                                                                                                              
                  gen.writeString(value);                                                                                                                                                                        
                  gen.writeEndArray();                                                                                                                                                                           
              } catch (Exception e) {                                                                                                                                                                            
                  System.out.println("FAILED at pad=" + pad + ": " + e);                                                                                                                                         
              }                                                                                                                                                                                                  
          }                                                                                                                                                                                                      
      }                                                                                                                                                                                                          
  }
```

### Expected behavior
The string serializes successfully with the escaped character omitted, as it does at every other alignment.

kalayciburak:
## Summary

Fixes #1668.

`WriterBasedJsonGenerator.writeString(String)` throws
`ArrayIndexOutOfBoundsException` when a `CharacterEscapes` implementation
returns a zero-length sequence and that escaped character is the last
character of the string, aligned with the output buffer end.

`_writeStringCustom` plus `_prependOrWriteCharacterEscape` left
`_outputHead == _outputTail` at the buffer end. The closing quote then
wrote at index 4000 of a 4000-char buffer.

## Changes

- Empty custom escape now appends nothing (skip the write) instead of
  changing `_flushBuffer`.
- After the custom-escape scan, recycle the buffer if it is empty at
  the end so the next write does not need a no-op flush.
- Add `ZeroLengthCustomEscape1668Test` covering Writer and UTF-8
  generators, plus `writeString(String)` and `writeString(char[])`.

## Testing

```
./mvnw -B -ff -ntp test -Dtest=ZeroLengthCustomEscape1668Test,TestCustomEscaping -Dsurefire.useModulePath=false
```

Result: 9/9 GREEN.

```
./mvnw -B -ff -ntp test -Dsurefire.useModulePath=false
```

Result: 1836 tests, 0 failures, 2 skipped.

The same empty-escape case exists on 2.21 if a backport is wanted.

### DIFF
--- a/release-notes/CREDITS
+++ b/release-notes/CREDITS
@@ -82,3 +82,13 @@ DongNyoung Lee (@Dongnyoung)
  * Reported, fixed #1646: Non-blocking ByteBuffer parser `releaseBuffered()` can write
    already-consumed bytes
   (3.3.0)
+
+@hdimitrieski
+ * Reported #1668: WriterBasedJsonGenerator AIOOBE when a zero-length custom
+   escape lands at the output buffer boundary
+  (3.3.0)
+
+Burak KALAYCI (@kalayciburak)
+ * Contributed #1668: WriterBasedJsonGenerator AIOOBE when a zero-length custom
+   escape lands at the output buffer boundary
+  (3.3.0)
--- a/src/main/java/tools/jackson/core/json/WriterBasedJsonGenerator.java
+++ b/src/main/java/tools/jackson/core/json/WriterBasedJsonGenerator.java
@@ -1552,6 +1552,11 @@ public class WriterBasedJsonGenerator
             ++_outputTail;
             _prependOrWriteCharacterEscape(c, escCode);
         }
+        // [core#1668] empty custom escape can leave head==tail at the buffer
+        // end. Recycle after the scan so the next write does not call flush.
+        if (_outputHead == _outputTail && _outputTail >= _outputEnd) {
+            _outputHead = _outputTail = 0;
+        }
     }
 
     private void _writeSegmentCustom(int end)
@@ -1980,6 +1985,13 @@ public class WriterBasedJsonGenerator
             _currentEscape = null;
         }
         int len = escape.length();
+        // 18-Aug-2026, [core#1668]: empty custom escape appends nothing.
+        // Prefix is already flushed; drop the source char. Do not recycle
+        // the buffer here: callers still walk toward a precomputed end.
+        if (len == 0) {
+            _outputHead = _outputTail;
+            return;
+        }
         if (_outputTail >= len) { // fits in, prepend
             int ptr = _outputTail - len;
             _outputHead = ptr;
@@ -2079,6 +2091,9 @@ public class WriterBasedJsonGenerator
             _currentEscape = null;
         }
         int len = escape.length();
+        if (len == 0) { // [core#1668] empty custom escape: append nothing
+            return ptr;
+        }
         if (ptr >= len && ptr < end) { // fits in, prepend
             ptr -= len;
             escape.getChars(0, len, buffer, ptr);
@@ -2139,6 +2154,9 @@ public class WriterBasedJsonGenerator
             _currentEscape = null;
         }
         int len = escape.length();
+        if (len == 0) { // [core#1668] empty custom escape: append nothing
+            return;
+        }
         if ((_outputTail + len) > _outputEnd) {
             _flushBuffer();
             if (len > _outputEnd) { // very very long escape; unlikely but theoretically possible
--- a/release-notes/VERSION
+++ b/release-notes/VERSION
@@ -22,6 +22,10 @@ JSON library.
 #1646: Non-blocking ByteBuffer parser `releaseBuffered()` can write
   already-consumed bytes
  (reported, fix by DongNyoung L)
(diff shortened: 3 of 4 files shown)


release-notes/CREDITS
```diff
@@ -82,3 +82,13 @@ DongNyoung Lee (@Dongnyoung)
  * Reported, fixed #1646: Non-blocking ByteBuffer parser `releaseBuffered()` can write
    already-consumed bytes
   (3.3.0)
+
+@hdimitrieski
+ * Reported #1668: WriterBasedJsonGenerator AIOOBE when a zero-length custom
+   escape lands at the output buffer boundary
+  (3.3.0)
```
pjfanning, line 93: this is fine for the moment but is likely to be added to an earlier release since this is a bug fix - we can give you the details when we decide on which release to target. It may end up needing to be a 2.18 fix and we will forward merge it

pjfanning:
@kalayciburak If you haven't submitted a CLA, we will need one from you.

which is usually done by printing, filling/signing, scan/photo, email to cla at fasterxml dot com.

release-notes/CREDITS
cowtowncoder: I'd be ok with 2.21 backport (2.18 only for secvuln at this point, will be closed by end of 2026)

src/main/java/tools/jackson/core/json/WriterBasedJsonGenerator.java
```diff
@@ -2189,6 +2189,11 @@ protected void _flushBuffer() throws JacksonException
             } catch (IOException e) {
                 throw _wrapIOFailure(e);
             }
+        } else if (_outputTail >= _outputEnd) {
+            // 18-Aug-2026, [core#1668]: a zero-length custom escape can leave
```
cowtowncoder: This really should not be necessary.... Seems like wrong place for fix (side-effect)

cowtowncoder:
Thank you for reporthing the issue and contributing this @kalayciburak. I am bit confused by fix -- 0-length should append nothing, so why change "flushBuffer()" and not prevent call.

src/main/java/tools/jackson/core/json/WriterBasedJsonGenerator.java
kalayciburak: fair point, flushBuffer was the wrong place. empty escape now just skips the write so we dont leave tail at the buffer end. pushed

cowtowncoder:
Ok; good news, got the CLA!

I also realized that maybe change in `_flushBuffer()` makes sense, so will re-create 2.21 targeted PR (#1672), merge via that, crediting you @kalayciburak .

Thank you!

PR #1673, opened 2026-08-23
Fix AIOOBE in `readString(Writer)` at output buffer boundary

pjfanning:
`JsonParser.readString(Writer)` throws `ArrayIndexOutOfBoundsException` on the two byte-based parsers when a string's plain-ASCII prefix is an exact multiple of the 1024-char streaming buffer and the next character is an escape or a multi-byte UTF-8 char.

`_streamString(Writer)` in `UTF8StreamJsonParser` and `UTF8DataInputJsonParser` only checks `outPtr >= outBuf.length` *inside* the ASCII loop, before appending. On exit via `break ascii_loop`, `outPtr` can equal the buffer length, and the `case 1/2/3/4:` decode arms then write `outBuf[outPtr++]` unchecked.

```
"aaa...a\n"   (1024 'a's)  -> AIOOBE: Index 1024 out of bounds for length 1024
"aaa...aä"    (1024 'a's)  -> AIOOBE: Index 1024 out of bounds for length 1024
```

Recurs at every multiple of 1024. `ReaderBasedJsonParser` checks at the top of each iteration and is unaffected.

Issue appears to be related to #1288 and only affects 3.x releases.

## Fix

Add the flush check between the closing-quote test and the `switch`, covering cases 1-3 and the high surrogate of case 4 (case 4's existing mid-pair check still handles the low surrogate). The flush carries the `maxStringLength` validation so constraint enforcement stays on the same cadence as the other flush points.

## Tests

Two tests added to `ReadStringStreamingTest`:

- `escapeAtFullOutputBuffer` — `\n` escape after an ASCII prefix of exactly 1024 and 2048 chars, across `ALL_MODES`
- `multiByteAtFullOutputBuffer` — same boundary via 2-, 3- and 4-byte UTF-8 characters, across `ALL_BINARY_MODES`

The file's existing boundary tests used `OUT_BUF_SIZE - 1`, one char short of the failing case, which is why this slipped through. Both new tests fail without the fix and pass with it; full suite is 1714 tests, 0 failures.

No release-notes entry yet — no issue number filed for this.

Forward merge to 3.2 / 3.x left to a human.

https://claude.ai/code/session_0161vnbY6rGfPFAYBMnbv64c

### DIFF
--- a/release-notes/VERSION
+++ b/release-notes/VERSION
@@ -21,6 +21,9 @@ JSON library.
   lands at the output buffer boundary
  (reported by @hdimitrieski)
  (fix by @kalayciburak)
+#1673: `readString(Writer)` AIOOBE when an escape or multi-byte character
+  lands at the output buffer boundary
+ (fix by @pjfanning)
 
 3.1.6 (14-Aug-2026)
 
--- a/src/main/java/tools/jackson/core/json/UTF8DataInputJsonParser.java
+++ b/src/main/java/tools/jackson/core/json/UTF8DataInputJsonParser.java
@@ -2199,11 +2199,9 @@ public class UTF8DataInputJsonParser
 
             ascii_loop:
             while (true) {
-                c = _inputData.readUnsignedByte();
-                if (codes[c] != 0) {
-                    break ascii_loop;
-                }
-                // Flush intermediate buffer when full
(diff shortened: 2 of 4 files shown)


cowtowncoder (approved):
LGTM

Issue #1677, opened 2026-08-26
Validate `resetWithCopy()` known length before initial TextBuffer allocation


Dongnyoung:
### Problem

`ReadConstrainedTextBuffer` enforces `StreamReadConstraints.maxStringLength`
through the `TextBuffer.validateStringLength()` hook.

I understand that `TextBuffer.append()` intentionally performs deferred validation only when the current segment needs to expand (#923). This keeps `maxStringLength` approximate and avoids adding a validation check to every append operation.

However, `resetWithCopy(..., len)` appears to have a different case because the complete input length is already known before the initial buffer allocation.

For a fresh `TextBuffer`, the current flow is:

    resetWithCopy(..., len)
      -> buf(len)
      -> allocate char[] large enough for len
      -> append(..., len)
      -> current segment already has enough capacity
      -> append fast path
      -> validateAppend() is not called

For example, with:

    maxStringLength = 500
    len = 2000
    _currentSegment = null

`buf(2000)` allocates a segment large enough for the input before any `maxStringLength` validation. The subsequent `append()` then takes its fast
path because the newly allocated segment already has enough capacity.

So the concern here is not that `maxStringLength` may be exceeded slightly due to buffer-granularity validation. The concern is that a known oversized bulk-copy input can cause allocation and copy proportional to `len` before the constraint is consulted.

### Possible change

Would it make sense for the known-size `resetWithCopy()` paths to call the existing `validateStringLength(len)` hook before allocating the initial
segment?

Conceptually:

    resetWithCopy(..., len)
      -> validateStringLength(len)
      -> buf(len)
      -> append(..., len)

This would leave the existing `append()` fast path and the deferred validation policy from #923 unchanged.

The same pattern appears to apply to both:

- `resetWithCopy(char[], int, int)`
- `resetWithCopy(String, int, int)`

### Reproduction

A regression test can reproduce this using a `ReadConstrainedTextBuffer` with `maxStringLength = 500` and `len = 2000`.

For example, the expected constraint behavior can be expressed as:

```java
try (IOContext ioContext = makeConstrainedContext(500)) {
    TextBuffer buffer = ioContext.constructReadConstrainedTextBuffer();

    char[] chars = new char[2000];
    Arrays.fill(chars, 'A');

    assertThrows(StreamConstraintsException.class,
            () -> buffer.resetWithCopy(chars, 0, chars.length));
}
```

[...]

PR #1678, opened 2026-08-31
Add failing `ReaderBasedJsonParser` reproduction for #1677

Dongnyoung:
### Problem

This PR adds a parser-level failing reproduction for #1677, as discussed in the issue.

The reproduction uses the normal `JsonFactory.createParser(..., Reader)` API rather than calling `TextBuffer` directly. The custom `Reader` only controls read boundaries: it returns an over-limit string prefix in one read and the closing quote in the next read, which is a normal condition for streaming `Reader` input.

When the first read ends before the closing quote, `ReaderBasedJsonParser._finishString()` reaches:

`TextBuffer.resetWithCopy(char[], int, int)`

with the known prefix length already exceeding `maxStringLength`.

With `maxStringLength = 500` and a 2,000-character prefix, the current implementation does not throw `StreamConstraintsException`; `getStringLength()` completes normally. Therefore the added test intentionally fails.

### Reproduction path

`JsonFactory.createParser(..., Reader)` → `ReaderBasedJsonParser` → `getStringLength()` → `_finishString()` → `TextBuffer.resetWithCopy(...)`

This PR intentionally contains only the failing parser-level reproduction and does not include a production fix yet.

### DIFF
--- a/src/test/java/tools/jackson/core/unittest/read/SimpleParserTest.java
+++ b/src/test/java/tools/jackson/core/unittest/read/SimpleParserTest.java
@@ -745,6 +745,27 @@ class SimpleParserTest extends JacksonCoreTestBase
     }
 
     @Test
+    void readerBasedStringLengthEnforcesMaxStringLengthAtInputBoundary() throws Exception
+    {
+        final int maxLen = 500;
+        final String longText = "x".repeat(2_000);
+
+        JsonFactory factory = JsonFactory.builder()
+                .streamReadConstraints(StreamReadConstraints.builder().maxStringLength(maxLen).build())
+                .build();
+
+        // Keep the closing quote in the second read so _finishString() must
+        // copy the over-limit prefix into TextBuffer before loading more input.
+        try (JsonParser parser = factory.createParser(ObjectReadContext.empty(),
+                new ChunkedReader("[\"" + longText, "\"]"))) {
+            assertToken(JsonToken.START_ARRAY, parser.nextToken());
+            assertToken(JsonToken.VALUE_STRING, parser.nextToken());
+
+            assertThrows(StreamConstraintsException.class, parser::getStringLength);
+        }
+    }
+
+    @Test
     void readStringWithIncreasedLimit() throws Exception
     {
         for (int mode : ALL_MODES) {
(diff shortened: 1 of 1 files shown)


src/test/java/tools/jackson/core/unittest/read/SimpleParserTest.java
```diff
@@ -744,6 +744,27 @@ private void _testReadStringEnforcesMaxStringLength(int mode) throws Exception
         parser.close();
     }
 
+    @Test
+    void readerBasedStringLengthEnforcesMaxStringLengthAtInputBoundary() throws Exception
+    {
+        final int maxLen = 500;
```
cowtowncoder, line 750: Let's use something more sizable, given that constraints are meant for "hundred Kb" level.

So max len of say 100kB, feeding 200kB.

Dongnyoung: @cowtowncoder I tried the 100k / 200k case with the default reader buffer as well. In that setup, the initial `resetWithCopy()` only sees about 4k characters, and the existing segment-expansion validation eventually throws `StreamConstraintsException`.

So the issue is reproducible with the default parser when `maxStringLength` is smaller than the default token buffer size, for example with a 500-character limit and a 2,000-character prefix before the closing quote. Reproducing the same `resetWithCopy()` path at the 100k / 200k scale requires configuring a larger token buffer through the public `recyclerPool()` API.

Would you prefer keeping the reproduction focused on the default-buffer case, or using the larger public buffer configuration for the 100k / 200k case?

cowtowncoder: My point is that if there is no real problem to fix, let's not bother.

Is there a realistic way user would be affected by limited constraints checking being fixed by related PR?
I am not looking for artificial tests to tease out theoretical concerns but rather reproduction of realistic use cases.

Dongnyoung: I checked this with normal `StringReader` and `BufferedReader` usage, without controlling the read boundaries.

With `maxStringLength = 100_000` and a ~200,000-character string, `resetWithCopy(...)` only received about 4,000 characters, bounded by the parser's input buffer. Subsequent growth went through the normal `TextBuffer` segment path and eventually triggered `StreamConstraintsException`.

So I couldn't reproduce the large pre-validation allocation/copy I was concerned about through normal Reader-based parsing. The current reproduction does appear to depend on an artificially controlled read boundary (or a very small constraint relative to the input buffer).

Given that, I agree this doesn't demonstrate a meaningful real-world problem worth changing the production code for. Thanks for pointing this out. :)

cowtowncoder: @Dongnyoung that happens sometimes :)

Thanks for investigating!

PR #1681, opened 2026-09-01
Fix #1680: drop the per-character escape-table load in the ASCII copy loop

### LINKED ISSUE #1680: Avoid per-character escape-table load in _writeStringSegment ASCII loop
`_writeStringSegment()`'s ASCII fast loop reads `_outputEscapes[ch]` for every character:

```java
while (offset < len) {
    int ch = text.charAt(offset);
    if (ch > 0x7F || escCodes[ch] != 0) {
        break;
    }
    ++offset;
}
```

That is a second, character-dependent load on the critical path of every character copied.

### Suggested fix

For the standard escape table the same test is expressible with constants
(`ch < 0x20 || ch > 0x7F || ch == '"' || ch == '\\'`), so the loop can branch on a loop-invariant
flag:

```java
final boolean stdEsc = (escCodes == CharTypes.get7BitOutputEscapes());
```

C2 unswitches the loop on it and constant-folds the flag in each clone, so the standard-table clone
has no per-character table load. A custom quote character, `ESCAPE_FORWARD_SLASHES`, or a
`CharacterEscapes` instance all produce a different array, so the flag is false and the original
lookup runs unchanged.

### Measurements

JDK 25 / x86_64, one String property per object, 3 forks, 5x5 iterations, pinned. Removing the load
shortens the loop body enough for C2 to unroll it twice as far:

| | insns | `(%rsp)` operands | unroll | insns/char | stack ops/char |
|---|---|---|---|---|---|
| without | 31 | 4 | x2 | 15.5 | 2.0 |
| with | 55 | 4 | **x4** | **13.8** | **1.0** |

`SingleBench.serialize`: 356.5 -> 347.5 ns/op (3.1.5), 331.9 -> 318.6 ns/op (2.22.0).

Output is byte-identical: verified over all characters 0x00-0x100, quotes, backslashes, control
characters, forward slash, non-ASCII, and strings crossing the segment boundary.

### Important note on the size of the win

Measure this **after** https://github.com/FasterXML/jackson-databind/issues/6182. With that issue
still present, `String.charAt`'s branch profile is polluted, the copy loop is neither unrolled nor
range-check eliminated, and this change appears to be worth about -23%. Almost all of that is the
pollution rather than the removed table load. With the profile clean it is worth a few percent, as
above.

### Reproducer

https://github.com/franz1981/c2-writestring-spill-jmh -- `SingleBench.serialize`. The README
documents the `perfasm` loop measurements and the before/after numbers. `master` is Jackson 2.22.0;
the `jackson3-write-paths` branch is the same benchmark on 3.1.5.

franz1981:
_writeStringSegment() reads _outputEscapes[ch] per character. For the standard table the same test is constant-expressible, so the loop now branches on a loop-invariant flag that C2 unswitches and constant-folds; the standard clone has no table load. Custom quote char, escaped slashes and CharacterEscapes are unaffected -- different array, flag false, original lookup.

Shorter body lets C2 unroll 4 characters per iteration instead of 2: 15.5 ->
13.8 instructions and 2 -> 1 stack operands per character on JDK 25/x86_64. Writing one String property per object: 356.5 -> 347.5 ns/op (3.1.5).

Measure after databind#6182: with that present the loop is not unrolled at all and this change appears to be worth ~23%, which is mostly that issue and not the removed load.

Output verified byte-identical over chars 0x00-0x100, quotes, backslashes, control chars and strings crossing the segment boundary.

### DIFF
--- a/release-notes/CREDITS
+++ b/release-notes/CREDITS
@@ -68,3 +68,8 @@ Burak KALAYCI (@kalayciburak)
  * Contributed #1668: `WriterBasedJsonGenerator` AIOOBE when a zero-length custom
    escape lands at the output buffer boundary
   (3.1.7)
+
+Francesco Nigro (@franz1981)
+ * Reported, contributed fix for #1680: Avoid per-character escape-table load in
+   `_writeStringSegment` ASCII loop
+  (3.1.7)
--- a/release-notes/VERSION
+++ b/release-notes/VERSION
@@ -24,6 +24,9 @@ JSON library.
 #1673: `readString(Writer)` AIOOBE when an escape or multi-byte character
   lands at the output buffer boundary
  (fix by @pjfanning)
+#1680: Avoid per-character escape-table load in `_writeStringSegment` ASCII loop
+ (reported by @franz1981)
+ (fix by @franz1981)
 
 3.1.6 (14-Aug-2026)
 
--- a/src/main/java/tools/jackson/core/json/UTF8JsonGenerator.java
+++ b/src/main/java/tools/jackson/core/json/UTF8JsonGenerator.java
@@ -1470,11 +1470,17 @@ public class UTF8JsonGenerator
         int outputPtr = _outputTail;
         final byte[] outputBuffer = _outputBuffer;
         final int[] escCodes = _outputEscapes;
+        // 01-Sep-2026, franz1981: [core#1680] For the standard table the test is expressible
+        //   with constants; C2 unswitches on this flag so that clone has no per-character
+        //   table load. A custom quote char, escaped slashes or CharacterEscapes give a
+        //   different array and fall back to the lookup below.
+        final boolean stdEsc = (escCodes == CharTypes.get7BitOutputEscapes());
 
         while (offset < len) {
             int ch = cbuf[offset];
             // note: here we know that (ch > 0x7F) will cover case of escaping non-ASCII too:
-            if (ch > 0x7F || escCodes[ch] != 0) {
+            if (stdEsc ? (ch < 0x20 || ch > 0x7F || ch == '"' || ch == '\\')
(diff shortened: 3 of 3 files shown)


franz1981:
Thanks @pjfanning - I tried an alternative that avoids hardcoding anything: derive a 128-bit mask
(two `long`s) from whatever `_outputEscapes` is in effect, and test it with a shift. It would have
covered custom quote chars and `CharacterEscapes` too. Measured on the reproducer with real
jackson-core 3.1.5, `SingleBench.serialize`, JDK 25, 3 forks:

| | clean profile | `String.charAt` profile polluted |
|---|---|---|
| stock | 370.6 ± 42.7 | 444.9 ± 5.3 |
| this PR | **344.5 ± 11.5** | **345.2 ± 4.7** |
| mask | 381.6 ± 6.8 | 552.4 ± 20.8 |

So the mask is worse in both cases - its two extra live values add register pressure. Reading the
compiled loop explains why this PR helps: when the profile is polluted C2 cannot unroll, and then
the stock loop spills 13 stack operands per character and the mask 18, while this PR spills **zero**.
Dropping the table reference is what keeps the live set in registers. That is also why it is flat
across profiles while stock degrades 20%.

On your correctness point, I agree the constants must not silently drift from the table. Rather than
supporting or removing `_outputEscapes`, I can make the duplication self-verifying: compute a
`static final boolean` once by checking the predicate against `get7BitOutputEscapes()` over 0..127,
and require it in addition to the identity check, so if the default table ever changes the fast path
simply switches itself off instead of changing output. Plus a unit test asserting equivalence for
the standard table, escaped slashes, a custom quote char and `CharacterEscapes`.

Would that address the concern?

pjfanning:
@franz1981 have you seen #1682? There are probably a few different ways to do it.

PR #1583, opened 2026-03-27
Fix issue #1581 with surrogate chars in NonBlockingJsonParser

pjfanning:
backported #1582 
issue #1581 

Copilot suggests:

Now let me look at the actual source of the bug — the `_addName` method in `NonBlockingJsonParserBase` on the 2.21 branch, and the corresponding fix in the 3.x branch.Here's a thorough analysis of why the test in PR #1583 works on the `3.x` branch but fails on `2.21`, and what fix is needed.

## Root Cause

The `3.x` branch introduced a dedicated field `_pendingSurrogateInName` in `NonBlockingJsonParserBase` (since 3.1), along with a `_finishPropertyWithEscape()` method that explicitly handles surrogate pairs split across chunk boundaries:

```java name=NonBlockingJsonParserBase.java url=https://github.com/FasterXML/jackson-core/blob/c53454a55c2e8f6642459378befaf7fe28050f1d/src/main/java/tools/jackson/core/json/async/NonBlockingJsonParserBase.java#L117-L165
/**
 * High surrogate code point awaiting matching low surrogate during
 * property name parsing, or 0 if none pending.
 *
 * @since 3.1
 */
protected int _pendingSurrogateInName;
```

```java name=NonBlockingUtf8JsonParserBase.java url=https://github.com/FasterXML/jackson-core/blob/c53454a55c2e8f6642459378befaf7fe28050f1d/src/main/java/tools/jackson/core/json/async/NonBlockingUtf8JsonParserBase.java#L2505-L2565
protected final JsonToken _finishPropertyWithEscape() throws JacksonException
{
    int ch;

    // [jackson-core#1541]: Check if we have a pending high surrogate
    if (_pendingSurrogateInName != 0) {
        // We have a high surrogate saved, now need to decode the low surrogate escape
        if (_quotedDigits == -2) {
            // Need to read the backslash first
            if (_inputPtr >= _inputEnd) {
                return JsonToken.NOT_AVAILABLE;
            }
            int b = getNextUnsignedByteFromBuffer();
            if (b != INT_BACKSLASH) {
                _reportError("...");
            }
            _quotedDigits = -1;
            _quoted32 = 0;
        }
        ch = _decodeSplitEscaped(_quoted32, _quotedDigits);
        if (ch < 0) {
            _minorState = MINOR_PROPERTY_NAME_ESCAPE;
            return JsonToken.NOT_AVAILABLE;
        }
        ch = _decodeSurrogate(_pendingSurrogateInName, ch);
[...]

Issue #1541, opened 2026-02-06
Unexpected Illegal surrogate character when parsing field names


AfzalivE:
Reference: Discussion in https://github.com/FasterXML/jackson-core/pull/1494

Affects: 2.21.0/2.21.1 at least, possibly 3.0.x (but not 3.1.0+)

Seems like something in broken/missing in field name decoding with JSON escapes.

In the minimum repro unit tests below, `acceptJsonEscapedSurrogatePairInFieldName` is failing but `acceptJsonEscapedSurrogatePairInStringValue` passes. I don't know nearly enough about unicode but seems like either both should fail or both should pass (as evidenced by similar tests in `UTF8SurrogateValidation363Test`). Seems like the field name codepath is doing something different.

```java
    @Test
    void acceptJsonEscapedSurrogatePairInFieldName() throws Exception
    {
        // JSON: {"\ud83d\udc4d":"value"}
        byte[] doc = new byte[] {
            '{', '"',
            '\\', 'u', 'd', '8', '3', 'd',  // JSON escape: \ud83d (high surrogate)
            '\\', 'u', 'd', 'c', '4', 'd',  // JSON escape: \udc4d (low surrogate)
            '"', ':', '"', 'v', 'a', 'l', 'u', 'e', '"',
            '}'
        };

        try (JsonParser p = FACTORY.createParser(doc)) {
            assertToken(JsonToken.START_OBJECT, p.nextToken());
            assertToken(JsonToken.FIELD_NAME, p.nextToken());
            // The escaped surrogate pair should decode to U+1F44D (thumbs up emoji)
            assertEquals("\uD83D\uDC4D", p.currentName());
            assertToken(JsonToken.VALUE_STRING, p.nextToken());
            assertEquals("value", p.getText());
            assertToken(JsonToken.END_OBJECT, p.nextToken());
        }
    }

    /**
     * Test that JSON escape sequence \ud83d\udc4d in string value is accepted.
     *
     * JSON: {"key":"\ud83d\udc4d"}
     */
    @Test
    void acceptJsonEscapedSurrogatePairInStringValue() throws Exception
    {
        // JSON: {"key":"\ud83d\udc4d"}
        byte[] doc = new byte[] {
            '{', '"', 'k', 'e', 'y', '"', ':', '"',
            '\\', 'u', 'd', '8', '3', 'd',  // JSON escape: \ud83d (high surrogate)
[...]

PR #1684, opened 2026-09-06
Fix #1683: reject JSON-escaped lone surrogates in ReaderBasedJsonParser

### LINKED ISSUE #1683: JSON \uXXXX escape accepts lone surrogates in string values (follow-on to #1494 for #363)
**The problem**

`ReaderBasedJsonParser` accepts JSON string values and field names containing unpaired UTF-16 surrogate `\uXXXX` escapes and returns them as `char[]` with `isSurrogate(c) == true` orphans. `UTF8StreamJsonParser` was made strict for field names by #1541 (2.21.2). The Reader-based parser and the string-value path were not touched.

**The specs**

- [RFC 8259 §8.2](https://www.rfc-editor.org/rfc/rfc8259.html#section-8.2): "behavior of software that receives JSON texts containing such values is unpredictable; for example, implementations might return different values for the length of a string value or even suffer fatal runtime exceptions."
- [RFC 7493 §2.1](https://www.rfc-editor.org/rfc/rfc7493.html#section-2.1) (I-JSON): "Object member names, and string values in arrays and object members, MUST NOT include code points that identify Surrogates or Noncharacters as defined by [UNICODE]... `\uDEAD` is invalid because it is an unpaired surrogate."
- [RFC 8785 §3.2.2.2](https://www.rfc-editor.org/rfc/rfc8785.html#section-3.2.2.2) (JCS): "occurrences of such data MUST cause a compliant JCS implementation to terminate with an appropriate error."

**Reproducer: Reader vs UTF8Stream disagree (jackson-core 2.22.2, JDK 21)**

```java
import com.fasterxml.jackson.core.*;
import java.nio.charset.StandardCharsets;

public class Repro1683 {
    public static void main(String[] args) throws Exception {
        String doc = "{\"\\uD800\":1}";
        JsonFactory f = new JsonFactory();

        try (JsonParser p = f.createParser(doc)) {
            p.nextToken(); p.nextToken();
            String name = p.currentName();
            System.out.println("Reader    : ACCEPT length=" + name.length()
                + " char 0=0x" + Integer.toHexString(name.charAt(0)));
        } catch (JsonParseException e) {
            System.out.println("Reader    : REJECT " + e.getMessage().split("\n")[0]);
        }

        try (JsonParser p = f.createParser(doc.getBytes(StandardCharsets.UTF_8))) {
            p.nextToken(); p.nextToken();
            String name = p.currentName();
            System.out.println("UTF8Stream: ACCEPT length=" + name.length()
                + " char 0=0x" + Integer.toHexString(name.charAt(0)));
        } catch (JsonParseException e) {
            System.out.println("UTF8Stream: REJECT " + e.getMessage().split("\n")[0]);
        }
    }
}
```

Output:
```
Reader    : ACCEPT length=1 char 0=0xd800
UTF8Stream: REJECT Broken surrogate pair in field name: expected '\' to start low surrogate, got 0x22
```

Two authoritative jackson parsers, same JSON bytes, opposite verdicts. The Reader-based side is the site #1541 didn't cover.

PR #1684 mirrors #1541's shape on `ReaderBasedJsonParser`.

elang2:
Fixes #1683.

Mirror of #1541 for `ReaderBasedJsonParser`. The `\uXXXX` escape decoder now rejects lone / reversed surrogates in field-name and string-value positions, plus `_skipString` for consistency with `skipChildren()` callers. Error messages match #1541's wording; helpers are private and per-parser to keep duplication low across the four sites.

Not gated behind a `JsonReadFeature` — matches #1542/#1583. If your #1494 note about wanting this class of validation feature-gated has become the preferred shape since, happy to rebase behind a new `StreamReadFeature`.

Scope: PR 1 of a series covering `ReaderBasedJsonParser` alone. `UTF8StreamJsonParser` (string-value), `UTF8DataInputJsonParser` (never received #1541), and the async parser (string-value) will follow as separate PRs matching the #1541 → #1567 → #1583 progression.

### DIFF
--- a/release-notes/CREDITS-2.x
+++ b/release-notes/CREDITS-2.x
@@ -554,3 +554,9 @@ Burak KALAYCI (@kalayciburak)
  * Contributed #1668: `WriterBasedJsonGenerator` AIOOBE when a zero-length custom
    escape lands at the output buffer boundary
   (2.21.7)
+
+Elankumaran Srinivasan (@elang2)
+ * Contributed #1683: `ReaderBasedJsonParser` should reject JSON-escaped lone
+   surrogates in field names and string values (mirror of #1541 fix in
+   `UTF8StreamJsonParser`)
+  (2.23.0)
--- a/release-notes/VERSION-2.x
+++ b/release-notes/VERSION-2.x
@@ -20,6 +20,10 @@ a pure JSON library.
   based on supplied length
  (requested by @kilink)
  (contributed by @seonwooj0810)
+#1683: `ReaderBasedJsonParser` should reject JSON-escaped lone
+  surrogates in field names and string values (mirror of #1541
+  fix in `UTF8StreamJsonParser`)
+ (contributed by @elang2)
 
 2.22.3 (not yet released)
 
--- a/src/main/java/com/fasterxml/jackson/core/json/ReaderBasedJsonParser.java
+++ b/src/main/java/com/fasterxml/jackson/core/json/ReaderBasedJsonParser.java
@@ -1866,6 +1866,46 @@ public class ReaderBasedJsonParser
                      * For now let's assume it does not.
                      */
                     c = _decodeEscaped();
+                    // [jackson-core#1683]: Validate JSON-escaped surrogates in
+                    //   field name. Mirror of [jackson-core#1541] fix in
+                    //   UTF8StreamJsonParser.
+                    if (c >= 0xD800 && c <= 0xDFFF) {
+                        if (c < 0xDC00) { // high surrogate: must be followed by low surrogate escape
+                            char hi = c;
+                            if (_inputPtr >= _inputEnd) {
+                                if (!_loadMore()) {
+                                    _reportInvalidEOF(" in field name", JsonToken.FIELD_NAME);
+                                }
+                            }
+                            if (_inputBuffer[_inputPtr] != INT_BACKSLASH) {
+                                _reportUnexpectedCharAfterHighSurrogate(_inputBuffer[_inputPtr] & 0xFFFF, "field name");
+                            }
+                            ++_inputPtr;
+                            char lo = _decodeEscaped();
+                            if (lo < 0xDC00 || lo > 0xDFFF) {
+                                _reportBrokenSurrogatePair(lo, "field name");
+                            }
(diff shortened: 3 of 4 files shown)


Dongnyoung:
One clarification: `ReaderBasedJsonParser` can also receive raw UTF-16 surrogate code units directly from a `Reader`.

With this change, escaped pairs (`\uD83D\uDE00`) are validated, but mixed raw/escaped pairs are rejected, while fully raw pairs and raw lone surrogates are still accepted.

Is this intentional, with validation limited to JSON-escaped surrogates? Or would it make sense to validate surrogate pairs consistently regardless of whether each code unit is escaped or raw?

elang2:
Good catch @Dongnyoung. Yes, intentional. This PR only checks what `\uXXXX` escapes produce.

```
"\uD800" + "\uDC00"    accepted    escaped pair
U+D800 + U+DC00        accepted    raw pair
"\uD800" + U+DC00      rejected    escaped + raw
U+D800 + "\uDC00"      rejected    raw + escaped
"\uD800"               rejected    escaped lone
U+D800                 accepted    raw lone
```

Raw lone surrogates are accepted, like you said.

The mixed cases are deliberate rather than incidental; the tests here assert both directions
reject. My thinking was that an escaped half next to a raw half is more likely a producer bug than
intent. Whether that's the right call is fair to question. It's a policy question first, and for
the raw + escaped direction it also needs a lookbehind the parser doesn't currently keep.

Rejecting raw lone surrogates is a different matter -- it would change behaviour for input that
parses today. I'll file that one as a separate issue so it's tracked and out of this PR's scope.

Dongnyoung:
@elang2 Thanks, that makes sense. A separate issue sounds good — I think it would also help clarify the expected behavior for raw and mixed raw/escaped surrogate pairs.

elang2:
> @elang2 Thanks, that makes sense. A separate issue sounds good — I think it would also help clarify the expected behavior for raw and mixed raw/escaped surrogate pairs.

@Dongnyoung  here you go. please review https://github.com/FasterXML/jackson-core/issues/1705

Dongnyoung:
Thanks for opening #1705 and for looking into this! @elang2

PR #1687, opened 2026-09-07
Add StripedArrayPool and virtual-thread-aware HybridPool recycler pools

stevenschlansker:
Context: While upgrading Blackbird, we identified contention in the buffer pool recycling algorithm. We found that pollFirst / offerLast on a ConcurrentLinkedDeque has less than optimal performance:

- each queue element needs a Node wrapper
- threads tend to contend on the head and tail
- the FIFO ordering means the coldest array is always used.

So, I asked Claude for options on a better design. Through a number of iteration rounds, we settled on a hybrid approach: platform threads continue to use soft thread locals, while virtual threads probe a striped atomic reference array queue, avoiding contending on a single head or tail. The array is padded to avoid false sharing effects.

The hope is that this hybrid pool ends up performing equal or better than any current implementation, and eventually could become default.

PR designed with assistance from Claude Fable 5. 

**Claude's report follows:**

StripedArrayPoolBase recycles through a fixed array of atomic slots scanned from a thread-id home index: acquire is getAndSet(null), release is a CAS into the first empty slot, and a release with every slot occupied drops the instance, which bounds retention. No queue node is allocated on either path.

HybridPoolBase extends it: platform threads get the exact ThreadLocal leave-in behavior (SoftReference slot, unlinked, no-op release) and virtual threads share the striped slots, so their instances recycle across threads instead of dying with each thread. The virtual-thread check is a MethodHandle resolved once against Thread#isVirtual with an always-false fallback, so the code runs on the JDK 17 floor and the Android SDK 26 API baseline.

Concrete BufferRecycler pools and factory methods follow the existing JsonRecyclerPools patterns (GLOBAL shared instances, construct(), readResolve). defaultPool() is unchanged. Tests cover lifecycle, the retention bound, capacity rounding, clear()/pooledCount() conformance, double-lease stress with platform and virtual threads, JDK serialization round-trips, and both paths of the hybrid pool; virtual-thread tests skip below JDK 21.

Measured with JMH on a 485-byte POJO document through databind 3.3, as paired, order-alternating, calibrated rounds on linux-aarch64 (JDK 26) and linux-x86_64 (Temurin 25). Against the current default (ConcurrentDequePool), the hybrid pool improves 4-thread writes by +36% on aarch64 and +19% on x86_64, with every round positive on both architectures. Four-thread reads improve by +13% and +10%, and single-thread writes by +4% and +1%. Virtual-thread-per-task workloads are parity in throughput and in GC allocation: the shared slots recycle at 2,432 B/op, against 2,456 B/op for the deque and 2,968 B/op with no pooling. For comparison, threadLocalPool() on the same virtual-thread workload allocates 3,152 B/op, more than no pooling at all, because each short-lived thread abandons its recycler, SoftReference, and ThreadLocal map entry. The striped pool alone measures +22%/+20% on 4-thread writes and +8%/+8% on 4-thread reads. No benchmark cell regressed against the default on either architecture. The default  slot count (16) was checked by a capacity sweep from 4 to 64 slots: virtual-thread allocation is flat across the whole range and contended writes show no size trend, so drop-on-full does not engage at these concurrencies.

At wider concurrency the gap grows: on a 14-16 thread Apple Silicon host (native, single run of the PR implementations), 16-way contended writes measure 3,995k ops/s for the deque against 9,050k for the striped pool (2.3x) and 15,171k for the hybrid (3.8x), with contended reads at +46% and +111%. The deque's shared head/tail CAS pair is the degrading element as thread count grows; the striped slots scale, and the hybrid's platform path (no shared traffic at all) scales best. The in-repo throughput harness reproduces this run.


pjfanning:
Fails on an Animal Sniffer check for Android support. You can use MethodHandles but we need to have a failover that kicks in MethodHandles are unsupported in the JVM (older Android releases).

stevenschlansker:
Updated to have a fallback for Android.

franz1981:
@stevenschlansker this is very similar to the approach of Vertx we implemented with @mariofusco 

See https://github.com/eclipse-vertx/vert.x/pull/4920

Would be worth referencing it, benchmarking and comparing against it too?🙏

In this current pr, I see no protection from false sharing and other diff which could be worth to be aligned under heavy contention.

Another possible approach could be:
- striped slot are not a single cell but a Treiber intrusive stack, to have LIFO locality and save Node allocations
- presize the striped array, padded for false sharing, to be max available cores
- use an atomic length to track the current grown state
- perform the linear probe N times (3?) and eventually try growing the length value 
- each stripe slot can be lazily allocated 
- a release is not a probe but reuse the same slot index, which is now stable 
- the striped array length can start with 1

This should keep the acquired slot index stable and allow a release to hit the same stripe. This should help to keep the slots balanced, unless only a specific VT slot id is mostly utilized and then never again.

stevenschlansker:
Thank you for the suggestions! I was not aware of the vertx implementation - if you feel your implementation already covers these additional points well and would like to contribute it to Jackson, please feel free to open a PR and I will gladly close this one. Otherwise, I'll see which of the suggestions I can fit in and add a "inspired by" line.

stevenschlansker:
I see the vert.x implementation is Eclipse public license - I think that's okay for me to "read and be inspired by" and then contribute under Apache 2 license? Just want to clarify before I do anything "bad".

Edit: Looks like it's "or Apache 2", so no problem.

franz1981:
@stevenschlansker 
Thanks! I think that's it could be because of the training set of a modern LLM model, so is common that these models produced something similar or it has taken inspiration from the striped64 class in JDK  too.

franz1981:
Whatever we can allow us (Vertx/Quarkus) to remove code is welcome 😁 
WDYT @vietj ?

stevenschlansker:
Thanks for pointing out the false sharing, as you say the effect is significant. I updated this PR to pad the entries and now the striped pool performs approximately at parity with the ThreadLocal implementation.

PR #1688, opened 2026-09-07
Add `JsonParser.nextNameMatchAndToken()`: fused name match plus value-token advance

stevenschlansker:
Combined form of nextNameMatch() and nextToken(): on a non-negative match the current token is the property's value token; negative results leave the stream exactly as nextNameMatch() does. Default implementation is the two-call sequence; UTF8StreamJsonParser commits the value token it already classified during name matching, saving the second dispatch and the nextToken() prologue per property. Motivation: generated per-bean deserializers dispatch on the match index and then read the value; nextNameMatch (597 bytes) never inlines into such callers, so each property pays two out-of-line parser calls today.

This improvement is a wash with vanilla deserialization, as it's not measurable behind megamorphic dispatches into value deserializers.
However, upcoming Blackbird rewrite will use this feature when code-generating Jackson codecs, and in that work I measure a +5% to +13% performance boost

PR designed with help from Claude Fable 5.

### DIFF
--- a/release-notes/CREDITS
+++ b/release-notes/CREDITS
@@ -98,3 +98,8 @@ Burak KALAYCI (@kalayciburak)
  * Contributed #1668: `WriterBasedJsonGenerator` AIOOBE when a zero-length custom
    escape lands at the output buffer boundary
   (3.1.7)
+
+Steven Schlansker (@stevenschlansker)
+ * Contributed #1688: Add `JsonParser.nextNameMatchAndToken()`, fused property
+   name match plus value token advance
+  (3.3.0)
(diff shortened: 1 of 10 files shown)


cowtowncoder:
Quick note: new method must be delegated; added.

cowtowncoder:
@stevenschlansker I worked on this a bit and I think it might be ready for merging. WDYT? Just let me know and I can merge it when ready (mark as "Ready for review").

stevenschlansker:
Thank you for looking at this and for the improvements. I agree, good to ship. One optional documentation point: a one-line javadoc note on `JsonParserDelegate` (or the base method) saying "subclasses that override `nextToken` or `nextNameMatch` should also override `nextNameMatchAndToken`" would give any future subclass writers a hint to preserve correctness.

PR #1694, opened 2026-09-08
Release IOContext when parser/generator construction fails

pjfanning:
Second of three PRs from the resource-handling sweep that followed #1692 (see also #1693 for the file-descriptor leaks).

`TokenStreamFactory._createContext()` (`TokenStreamFactory.java:1218`) leases a `BufferRecycler` from the configured pool, and the only thing that returns it is `IOContext.close()` — which requires a parser or generator that was successfully constructed **and** later closed. So any failure between the two drops the lease:

- invalid `byte[]` / `char[]` offsets (`_checkRangeBoundsForByteArray`)
- a `maxDocumentLength` violation, validated up front for fixed buffers
- a missing file, or a throwing `InputDecorator` / `OutputDecorator`
- a backend's `_createParser` / `_createGenerator` rejecting a schema

The [core#763] guard in `JsonFactory._createParser(ObjectReadContext, IOContext, InputStream)` closes a managed stream but does not release the context.

```
pooled recyclers after 10 failed createParser(byte[]): 0     (before)
pooled recyclers after 10 failed createParser(byte[]): 1     (after)
```

Severity is modest — the recycler is garbage-collected and `BoundedPool` simply holds fewer entries, so this is lost recycling rather than exhaustion — but under a workload that rejects input regularly (constraint limits doing their job) the pool stays cold.

### Fix

- New package-private `DecorableTSFactory._releaseOnFailedConstruction(IOContext, RuntimeException)`: closes the context, attaching any secondary failure as suppressed. No public API change.
- Every `createParser` / `createGenerator` path in `TextualTSFactory` and `BinaryTSFactory` routes failures through it. Since these methods are inherited, the format backends get the fix too.

### Tests

New `unittest/json/FailedConstructionRecyclerTest` (4 tests) drives a bounded pool through the failure shapes above — constraint violation, invalid bounds, and a `JsonFactory` subclass that throws from `_createParser`/`_createGenerator` the way a schema-validating backend does — asserting `pooledCount()` afterwards. All four fail without the main-source change (`expected: <1> but was: <0>`).

Full `./mvnw -B -ff -ntp verify` passes (1718 tests).

### Note

Branched off `3.1`, and it wraps the same create methods as #1692 and #1693, so expect merge conflicts. The two helpers have distinct names and signatures (`_closeOnFailedConstruction` vs `_releaseOnFailedConstruction`), so resolution is keeping both calls in the shared `catch` blocks. Happy to rebase whichever lands later.

### DIFF
--- a/release-notes/VERSION
+++ b/release-notes/VERSION
@@ -24,6 +24,9 @@ JSON library.
 #1673: `readString(Writer)` AIOOBE when an escape or multi-byte character
   lands at the output buffer boundary
  (fix by @pjfanning)
+- Release `IOContext` (and `BufferRecycler` lease it holds) when construction of
+  parser or generator fails
+ (fix by @pjfanning)
 
 3.1.6 (14-Aug-2026)
 
--- a/release-notes/VERSION
+++ b/release-notes/VERSION
@@ -24,7 +24,7 @@ JSON library.
 #1673: `readString(Writer)` AIOOBE when an escape or multi-byte character
   lands at the output buffer boundary
  (fix by @pjfanning)
-- Release `IOContext` (and `BufferRecycler` lease it holds) when construction of
+#1694: Release `IOContext` (and `BufferRecycler` lease it holds) when construction of
   parser or generator fails
  (fix by @pjfanning)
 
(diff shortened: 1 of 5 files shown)


cowtowncoder (approved):

cowtowncoder:
This could have gone in 3.x, fwtw. But will merge in 3.1 since PR is for that.

Issue #1649, opened 2026-08-09
Optimize `NumberInput.looksLikeValidNumber` [CVE-2026-89407]


cowtowncoder:
Currently implementation of helper method `looksLikeValidNumber` in `tools.jackson.core.io.NumberInput` uses `java.util.regex.Pattern`s for matching.
This is functionally correct, but is inefficient for sizable content (potentially matching and some complicated non-matching cases). Due to relative simplicity of Patterns, it should be straight-forward to convert code to regular Java functions.

cowtowncoder:
Backported, so fix is or will be in:

* 2.18.11
* 2.21.7
* 2.22.3
* 3.1.7
* 3.2.2 (already released)

cowtowncoder:
Add CVE id: CVE-2026-89407

PR #1699, opened 2026-09-10
Replace regexp in NumberInput.looksLikeValidNumber() with single-pass scan

pjfanning:
### Problem

`NumberInput.looksLikeValidNumber()` validated with:

```java
PATTERN_FLOAT             = "[+-]?[0-9]*[\.]?[0-9]+([eE][+-]?[0-9]+)?"
PATTERN_FLOAT_TRAILING_DOT = "[+-]?[0-9]+[\.]"
```

`[0-9]*` and `[0-9]+` are adjacent, separated only by an optional dot, over the same character
class. Java's backtracking engine has no possessive quantifier or atomic group here, so on input
that does not match it must try every split point between the two groups — and for each split the
inner `[0-9]+` backtracks over its own run. That is quadratic. A failing match then runs the entire
string through `PATTERN_FLOAT_TRAILING_DOT` as well, doubling the constant factor.

Worst case is a long digit run whose only invalid character is at the very end:

| N (digits, then one non-digit) | old | new |
|---|---|---|
| 1,000 | 27 ms | 0.03 ms |
| 4,000 | 160 ms | 0.13 ms |
| 8,000 | 661 ms | 0.14 ms |
| 16,000 | 2,538 ms | 0.26 ms |
| 32,000 | 10,387 ms | 0.55 ms |

Doubling N quadruples the old time, as expected. Extrapolating the same curve to 20,000,000 — the
`maxStringLength` default that gates this path, rather than the `maxNumberLength` default of 1,000
that bounds numeric content elsewhere — puts a single call in the region of weeks. The new scan does
20M in 22 ms.

### Change

Replace both expressions with one left-to-right pass: optional sign, integer digits, optional dot
and fraction digits, optional exponent, then require the scan to have consumed the whole string. No
allocation, no backtracking, and the second pattern's trailing-dot case (`"12."`) folds into the same
walk.

### Behaviour is unchanged

Verified against the original expressions, not just re-asserted:

- **Exhaustive** over the alphabet `+-.09eEx` for every input up to length 6 — 299,593 strings, zero
  divergence.
- **Randomised**, digit-weighted, up to length 14 — 400,000 strings, zero divergence.

The new correctness tests in `NumberInputTest` were each checked to produce the same answer under the
old regexes as under the new scan, so the suite pins down preserved behaviour rather than newly
chosen behaviour. They cover sign placement, leading/trailing dots, multiple dots, exponent forms and
malformed exponents, non-ASCII digits (Arabic-Indic and full-width, which `[0-9]` never accepted),
untrimmed whitespace, and `null`.

There is also a regression test for the original complaint: a 1,000,000-character input under a 10s
`assertTimeoutPreemptively`. It fails against the old implementation (times out) and passes in ~1 ms
against the new one. Happy to drop that one if you would rather keep timing out of the suite.

Full build green: 1552 tests, 0 failures; `animal-sniffer:check` clean.

### DIFF
--- a/release-notes/VERSION-2.x
+++ b/release-notes/VERSION-2.x
@@ -20,6 +20,8 @@ a pure JSON library.
   lands at the output buffer boundary
  (reported by @hdimitrieski)
  (fix by @kalayciburak)
+#1699: Replace regexp in `NumberInput.looksLikeValidNumber()` with single-pass
+  scan, to avoid quadratic backtracking on long input
 
 2.21.6 (14-Aug-2026)
 
--- a/src/main/java/com/fasterxml/jackson/core/io/NumberInput.java
+++ b/src/main/java/com/fasterxml/jackson/core/io/NumberInput.java
@@ -2,7 +2,6 @@ package com.fasterxml.jackson.core.io;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
-import java.util.regex.Pattern;
 
 import ch.randelshofer.fastdoubleparser.JavaDoubleParser;
 import ch.randelshofer.fastdoubleparser.JavaFloatParser;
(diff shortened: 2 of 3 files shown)


cowtowncoder:
Duplicate of #1649;  backport #1650 that went in 3.2 ?

pjfanning:
Sure @cowtowncoder, can you backport #1649? The regex is very very slow on large text and it is applied before we check the number len constraint.

cowtowncoder:
@pjfanning I can do it if you want; coding agents make this simple enough. I assume this needs to actually go in 2.18, not just 2.21 (2.18.x will be open till end of 2026).

pjfanning:
> @pjfanning I can do it if you want; coding agents make this simple enough. I assume this needs to actually go in 2.18, not just 2.21 (2.18.x will be open till end of 2026).

2.18 makes sense - I thought 2.18 was already being wound down, I'll retarget my other PRs to 2.18

cowtowncoder:
@pjfanning Security fixes only to 2.18 (CVEs), other fixes can go to later (2.21, 2.x or only 3.*), depending.

PR #1706, opened 2026-09-12
Write number text directly as ASCII bytes in `UTF8JsonGenerator`

pjfanning:
`UTF8JsonGenerator.writeNumber()` for `double`/`float` (when `USE_FAST_DOUBLE_WRITER` is off), `BigInteger` and `BigDecimal` went through `writeRaw(String)`, which copies the text into `_charBuffer` via `getChars()` and then runs the general UTF-8 encoding loop (with 3x room reservation) over it. Number text is always ASCII, so this adds `_writeRawAscii()` / `_writeQuotedAscii()` helpers that copy straight into the output `byte[]`, chunking across buffer flushes for very long text.

`writeNumber(String)` is left as-is since its text is caller-supplied. `WriterBasedJsonGenerator` already writes with `getChars()` directly into its output `char[]`, so nothing to do there.

### Numbers

Hand-rolled harness, JDK 17, 5 JVM forks, 2000 values x 300 reps per round, `USE_FAST_DOUBLE_WRITER` off:

| case | base median | new median | base min | new min |
|---|---|---|---|---|
| BigDecimal (`toString` cached, isolates the copy path) | 31.7 ms | 15.9 ms | 23.7 | 15.2 |
| double | 276 | 246 | 229 | 229 |
| float | 128 | 111 | 101 | 105 |
| BigInteger (200-bit) | 994 | 916 | 831 | 840 |
| double, quoted | 266 | 248 | 228 | 228 |
| BigInteger, quoted | 999 | 938 | 849 | 889 |

The copy path itself is roughly halved; for `double`/`float`/`BigInteger` the `toString()` cost dominates so the gain there is modest. Tried the deprecated `String.getBytes(int,int,byte[],int)` instead of a `charAt` loop: no measurable difference. The loop was later changed to copy via `String.getChars()` into `_charBuffer` and narrow from there, avoiding `charAt()` branch-profile pollution from non-Latin1 Strings elsewhere (see #1680); numbers above were measured with the `charAt` version.

Adds `GeneratorBasicTest.longNumberTextAcrossBufferBoundary` covering 20,000-digit `BigInteger`/`BigDecimal` through both generators, plain and quoted.

### DIFF
--- a/release-notes/VERSION
+++ b/release-notes/VERSION
@@ -44,6 +44,8 @@ JSON library.
 #1704: Add `NumberOutput.outputDouble/outputFloat(char[])`, use in
   `WriterBasedJsonGenerator` with `USE_FAST_DOUBLE_WRITER`
  (contributed by @pjfanning, w/ Claude code)
+#1706: Write number text directly as ASCII bytes in `UTF8JsonGenerator`
+ (contributed by @pjfanning, w/ Claude code)
 
 3.2.3 (not yet released)
 
--- a/src/main/java/tools/jackson/core/json/UTF8JsonGenerator.java
+++ b/src/main/java/tools/jackson/core/json/UTF8JsonGenerator.java
@@ -1058,9 +1058,9 @@ public class UTF8JsonGenerator
         if (value == null) {
             _writeNull();
         } else if (_cfgNumbersAsStrings) {
-            _writeQuotedRaw(value.toString());
+            _writeQuotedAscii(value.toString());
         } else {
-            writeRaw(value.toString());
+            _writeRawAscii(value.toString());
         }
         return this;
     }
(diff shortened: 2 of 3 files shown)


cowtowncoder:
(note: Claude Code generated comment)
------

Re-ran benchmarks against current PR head (`46e2af291`, `getChars()` version) vs merge-base `57bd0c655`.

Setup: JDK 17.0.19, 5 JVM forks, 2000 values x 300 reps per round, `USE_FAST_DOUBLE_WRITER` off, 8 warmup + 12 measured rounds, separate JVM per case, base/new order alternated per fork, output to a null `OutputStream`. 60 measured rounds per cell; times in ms per round.

| case | base median | new median | change | base min | new min |
|---|---|---|---|---|---|
| BigDecimal (`toString` cached) | 21 | 18 | -14% | 19 | 14 |
| double | 260 | 253 | -3% | 258 | 251 |
| float | 91 | 91.5 | ±0 | 89 | 89 |
| BigInteger (200-bit) | 514 | 492 | -4% | 508 | 488 |
| double, quoted | 264 | 257 | -3% | 260 | 251 |
| BigInteger, quoted | 518 | 493 | -5% | 511 | 488 |

New is faster in every fork for BigDecimal, double and BigInteger; float is within noise. No regressions.

Gains are smaller than in the original table (BigDecimal copy path not halved here). Likely causes: short BigDecimal values in this harness (up to ~20 digits, so copying is a smaller share of the work), different hardware, and a newer base commit.

cowtowncoder (approved):

PR #1707, opened 2026-09-12
Copy `String` content via `getChars()` in `UTF8JsonGenerator`, drop `charAt()` scanning

pjfanning:
Follow-up to the `String.charAt` profile-pollution discussion in #1682 / #1680 (thanks @franz1981).

`UTF8JsonGenerator.writeString(String)` and `writeName(String)` scanned the input with `String.charAt()`. That method has a single JVM-wide branch profile: one call on a non-Latin1 `String` anywhere in the process (e.g. `StdDateFormat.<clinit>` in databind before FasterXML/jackson-databind#6183, or any `writeString` of non-ASCII text) leaves an out-of-line `StringUTF16.charAt` call in every ASCII `charAt` loop C2 compiles afterwards, which disables unrolling and forces spills.

This PR makes the generator immune to that by copying `String` content into `_charBuffer` via `String.getChars()` (an intrinsic, no branch profile) and encoding from the `char[]` — the same thing `_writePPName` and `WriterBasedJsonGenerator` already did. The `String`-scanning twins of the segment methods (`_writeStringSegment(String,..)`, `_writeStringSegment2(String,..)`, `_writeStringSegmentASCII2(String,..)`, `_writeCustomStringSegment2(String,..)`, `_writeStringSegments(String,int,int)`) are removed (all `private`); `_writeStringSegments(String, boolean)` now chunks through the copy buffer instead.

Behavioural notes:
- Output is unchanged. Surrogate pairs are not split across copy-buffer chunks (extends the #1473 guard to the new 4000-char boundary).
- `writeString(String)` now also takes the chunked path when `len > _charBufferLength`, so a custom `BufferRecycler` with a copy buffer smaller than the output segment size works (new test covers a 100-char copy buffer).
- Complementary to #1682's table-free escape check; the two address different costs.

Not yet benchmarked — needs runs in both clean and polluted `charAt` profile states (the `pollutedCharAtProfile` `@Param` from https://github.com/FasterXML/jackson-core/pull/1682#issuecomment-5648799037). `getChars` adds a copy pass, so short strings on a clean profile may lose a little; the expectation is the polluted case recovers most of the ~20% franz measured.

Draft until the numbers are in.


franz1981:
@pjfanning what if #1707 also used the static-final 256-entry table for the `char[]` fast loop (default escapes only, i.e. `escCodes == CharTypes.get7BitOutputEscapes()`)? I tried it on top of your branch, and the way the index is formed matters more than I expected.

The obvious form, `if (ch > 0xFF || tbl[ch] != 0) break;`, does not remove the bounds check on a `char[]`: `ch` comes from a 16-bit load typed `[0, 65535]` and C2 doesn't narrow it from the dominating compare, so the loop keeps both `cmpl $0xff` and `cmpl $0x100` per char and only unrolls x2. Making the index provably in range does the job:

```java
final int[] tbl = STD_ESC_256; // 256 entries: std escapes for 0..127, non-zero for 128..255
while (offset < len) {
    int ch = cbuf[offset];
    if ((tbl[ch & 0xFF] | (ch >>> 8)) != 0) {
        break;
    }
    ++offset;
}
```

The `& 0xFF` gives the table load a static `[0, 255]` index (no range check), the `>>> 8` folds the non-Latin-1 case into the same test, there's no branch on the char itself, and C2 unrolls it x4. It's not a single instruction (`movzwl` + `movzbl` + `shr` + `or`), but it's branch-free and profile-independent.

Clearly a 64K static final table would be faster than any option if you can accept the trade off of having a 64K size table which is not used but the 128 bytes - unless you start to use it to index the escaping data there too - which is not a terrible idea, wdyt?

I am experimenting with another idea too, will share it soon

franz1981:
It seems that tables approaches are not as good on aarch64 😢 the comparison chain just does better

And the reason is not fun, see https://bugs.openjdk.org/browse/JDK-8392394

ARM is duplicating the instructions required to materialize the lookup table (once static final) 😭

And regardless the getChars intrinsics can be an overkill with small strings too

PR #1709, opened 2026-09-13
Validate shared `TextBuffer` append length before unsharing

### LINKED ISSUE #1708: `TextBuffer` may unshare before validating shared append length
## Summary
`TextBuffer` can validate `maxStringLength` too late when appending to a shared buffer.

When the buffer is initialized with `resetWithShared(...)`, append paths call `unshare(...)` before checking whether the resulting logical text length exceeds `StreamReadConstraints.maxStringLength`.

#### For example:

- shared length == `maxStringLength`
- append one more char

In this case, `unshare(...)` can request a private text buffer and copy the shared contents even though the resulting logical length already exceeds the configured constraint.

It seems preferable to validate `shared length + appended length` before unsharing, so the constraint failure occurs without the unnecessary buffer allocation/copy.

This is not an API change. It looks like a small internal constraint-enforcement timing fix, and may be suitable for backporting if maintainers think it is worthwhile.

Dongnyoung:
Fixes #1708 

### Summary

This changes `TextBuffer` so that appends to a shared buffer validate the resulting logical string length before calling `unshare(...)`.

Previously, when `TextBuffer` was in `resetWithShared(...)` mode, append paths called `unshare(...)` first. If the resulting logical length exceeded `maxStringLength`, Jackson could request a private text buffer and copy the shared contents before the constraint was enforced.

### Changes

- Validate shared append length before unsharing in:
  - `append(char)`
  - `append(char[], int, int)`
  - `append(String, int, int)`
- Validate using the actual appended content length.
  - For `append(char)`, validate `_inputLen + 1` while retaining `unshare(16)` for spare capacity.
- Add regression tests verifying that rejected shared-buffer appends do not request a text buffer.
- Add an exact-limit test where `shared length + appended length == maxStringLength`.

### Verification

- `.\mvnw.cmd -q -DskipTests test-compile`
- Manual before/after verification with a tracking `BufferRecycler`:
  - before: the append requested a text buffer before the constraint failure
  - after: the append fails before unsharing, with the allocation count remaining `0`


cowtowncoder:
As with some earlier suggestions, is there a realistic case where this matters? I am not sure I can see one.

Dongnyoung:
@cowtowncoder Thanks, that made me look more closely at the actual parser paths.

I found a more realistic case in `ReaderBasedJsonParser`. For char-backed property names, `_parseName2()` / `_handleOddName2()` can call `resetWithShared(...)` followed by `getCurrentSegment()`. That means the shared prefix is unshared and copied into a `CHAR_TEXT_BUFFER` before `maxNameLength` is checked later.

I was able to reproduce this with actual parser input and `StreamReadConstraints.maxNameLength(100)`:

- With `Reader` / `StringReader` input, the allocation is bounded by the parser input buffer, around 4K chars in my test.
- With direct `String` input below the char-array cutoff, the shared prefix can be much larger. In my test it grew to around 32K chars before the exception.
- With direct `char[]` input, the copied prefix can grow with the caller-provided buffer prefix.

So I agree that the original `append(...)` case is not a particularly strong real-world example. The property-name path seems more relevant, because it is actual parser input and the allocation/copy happens before the configured name-length constraint is enforced.

Would it make more sense to address this at the parser level instead, with a parser-level regression test and validation of the shared name length before `getCurrentSegment()` unshares it?🤔

cowtowncoder:
Yes: since `TextBuffer` does not (and should not) know if we are decoding String or name, it can't really validate things -- parser would be the place.

But even then, with all changes going on, this seems bit like distraction to be honest.

Dongnyoung:
Makes sense, thanks for clarifying!😊 I'll close this for now.

PR #1712, opened 2026-09-15
Fix leaked, double-closed and truncated `File`/`Path` generator targets on failed construction

### LINKED ISSUE #1711: `File`/`Path` generator targets may be leaked, double-closed or truncated when construction fails
## Description

`File`/`Path` generator construction in `TextualTSFactory` and `BinaryTSFactory` does not
consistently clean up the resource it owns when construction fails.

There are four distinct problems. (1) and (2) were in the original report; (3) and (4) were found
while reviewing the fix in #1712.

### 1. Output stream is opened before `IOContext` creation

For example, `TextualTSFactory.createGenerator(File, JsonEncoding)` starts with:

```java
final OutputStream out = _fileOutputStream(f);
final IOContext ioCtxt = _createContext(_createContentReference(f), true, enc);

try {
    // generator construction
} catch (RuntimeException e) {
    _releaseOnFailedConstruction(ioCtxt, e);
    _closeOnFailedConstruction(out, e);
    throw e;
}
```

If `_fileOutputStream(f)` succeeds but `_createContext(...)` fails, the failure happens before the
guarded block is entered, so the already-opened stream is never closed:

```text
open OutputStream -> create IOContext -> failure -> cleanup block never entered -> stream leaks
```

This reproduces for both the `File` and `Path` paths. Creating the `IOContext` before opening the
stream, and acquiring the output inside the guarded block, avoids the gap.

### 2. Decorated output is not closed when later construction fails

The decorated stream is passed straight into generator construction:

```java
_createUTF8Generator(writeCtxt, ioCtxt, _decorate(ioCtxt, out))
```

If the `OutputDecorator` returns a wrapper and generator construction then fails, the handler closes
the original `out` but not the wrapper. For a decorator holding its own state or resources, closing
only the underlying stream bypasses its cleanup.

Once decoration succeeds the decorated output is what was passed downstream, so it is what should be
closed. If decoration itself fails, the factory-created output remains the resource to close.

### 3. Stream opened by Jackson is closed twice

Once cleanup tracks the outermost resource, the natural implementation closes that resource and then
closes the raw stream as a fallback:

```java
_closeOnFailedConstruction(outputToClose, e);
if (rawOut != outputToClose) {
    _closeOnFailedConstruction(rawOut, e);
}
```

In the non-UTF-8 path `outputToClose` is always a `Writer` wrapping `rawOut`, so the guard is
unconditionally true and `rawOut.close()` runs twice; same whenever an `OutputDecorator` is
configured. `_fileOutputStream()`/`_pathOutputStream()` are `protected` and are overridden by format
backends, so a stream that is not close-idempotent will misbehave or contribute a spurious
suppressed exception.

The raw stream should be closed as a fallback only when closing the outer resource threw.

### 4. Binary `File` target is truncated when `IOContext` creation fails

`BinaryTSFactory.createGenerator(File, JsonEncoding)` builds the content reference from the opened
`OutputStream`, which forces the file to be opened -- and therefore truncated -- before
`_createContext()` can fail. A subclass whose `_createContext()` throws destroys the existing
contents of the file, while the identical call with a `Path` leaves it untouched.

Referencing the `File` itself, as `TextualTSFactory` and the `Path` overload already do, removes the
truncation window and reports better locations.

### Affected paths

- `TextualTSFactory.createGenerator(File, JsonEncoding)`
- `TextualTSFactory.createGenerator(Path, JsonEncoding)`
- `BinaryTSFactory.createGenerator(File, JsonEncoding)`
- `BinaryTSFactory.createGenerator(Path, JsonEncoding)`

### Expected behavior

`File`/`Path` generator construction should clean up the resource it currently owns at every failure
point:

- If `IOContext` creation fails, no factory-created output stream is left open, and the target file
  is not truncated.
- If output stream creation fails after an `IOContext` was created, the context is released.
- If decoration fails, the factory-created output stream is closed.
- If decoration succeeds but generator construction later fails, the decorated output is closed --
  and the underlying stream is not closed a second time.
- If the generator is constructed but generator *decoration* fails, the generator itself is not
  closed (closing a partially constructed generator could write to the target); the output resource
  is closed and the `IOContext` released.
- `IOContext` instances are released on failed construction.

Dongnyoung:
Fixes #1711

## Description

Fixes cleanup of `File`/`Path` generator construction in `TextualTSFactory` and `BinaryTSFactory`,
which could leak, double-close or truncate the target when construction fails.

## Changes

Generator construction tracks the outermost successfully created `Closeable` as the current cleanup
target. Ownership progresses as construction succeeds:

```text
OutputStream -> decorated OutputStream -> Writer / decorated Writer
```

On failure the outermost tracked resource is closed, instead of always closing the original output
stream.

Note that the `JsonGenerator` itself is deliberately *not* tracked: closing a partially constructed
generator could write to the target. When generator decoration fails, the output resource is closed
and the `IOContext` released, but the generator is left unclosed.

Specifically:

- `IOContext` is created before the output stream is opened, so a context-creation failure can no
  longer leave a factory-created stream open.
- A new `_closeOnFailedConstruction(Closeable, Closeable, RuntimeException)` closes the outermost
  resource and falls back to the raw stream *only if that close threw*, so the stream Jackson opened
  is no longer closed twice. (Package-private: no new API surface in a patch release.)
- `BinaryTSFactory.createGenerator(File, ...)` now builds its `ContentReference` from the `File`
  rather than from the opened `OutputStream`, matching `TextualTSFactory` and the `Path` overload.
  This removes the truncation window: the target is no longer opened (and emptied) before
  `_createContext()` has had a chance to fail. It also improves reported locations, but is a visible
  change for binary backends that assert on generator error messages or locations.

## Tests

Added regression coverage for:

- `File` and `Path` output cleanup when `IOContext` creation fails
- decorated `OutputStream` cleanup when generator construction fails
- decorated `Writer` cleanup for non-UTF-8 generator construction
- `BinaryTSFactory` `File`/`Path` construction failures, including that a context-creation failure
  leaves existing file content intact
- generator decoration failure with `AUTO_CLOSE_TARGET` disabled
- that the stream Jackson opened is closed exactly once

The full `jackson-core` test suite passes with these changes.


pjfanning:
#1692 and #1693 are on the 3.1 branch so if those changes are wrong, any fixes should be targeted at the same branch.
At the moment, I think this PR has too many issues but these issues can be fixed.
I may have time tonight to look again at the issues.

A Claude Fable review gave these issues with this PR:

Blocking or should-fix before merge

- Raw file stream no longer closed unconditionally on failure. The catch now closes only the outermost wrapper and relies on it cascading down. With a gzip or deflate OutputDecorator, a failing finish() in the wrapper's close never reaches the underlying file stream, so the descriptor leaks. Pre-PR the raw stream was always closed. Fix by hoisting out and closing it after the wrapper. Sites: TextualTSFactory.java:320 and :353, BinaryTSFactory.java:223 and :247.
- Wrong target branch. The bug and the helper methods it uses (#1692 to #1694, @since 3.1.7) exist on 3.1. The repo convention is oldest applicable branch, merged forward. Landing on 3.x only leaves 3.1 and 3.2 users with the leak and creates a forward-merge conflict.
- Unannounced behaviour change for binary backends. Setting outputToClose = generator runs the full close() lifecycle on failure, which flushes buffered content. For Smile or CBOR, whose constructors write a header, a failing JsonGeneratorDecorator now produces a file containing a valid empty document rather than zero bytes. JSON tests cannot see this. Either restrict the close to the stream chain and release the context separately, or document it.
- Three new tests pass vacuously. The _verifyNoLeakedOutput helper accepts an empty list. The Textual File/Path and Binary Path overloads create the context before opening the stream, so the context-failure tests never open one and only assert that an exception propagated. Use assertEquals(1, size()) where a stream must open and isEmpty() where the point is that none does. See FailedGeneratorConstructionCloseTest.java:396.
- Missing release notes entry for #1711. CREDITS already has the contributor, so only the VERSION line is needed, under 3.1.7 once retargeted.

Worth addressing, lower severity

- Partial decorator chain leaks. With several JsonGeneratorDecorators, if an earlier one returns a resource-owning wrapper and a later one throws, only the base generator is closed. Tracking the latest successful result in _decorate would cover it. See DecorableTSFactory.java:242.
- Only RuntimeException is caught. An Error during construction still leaks the file stream and buffer lease. Pre-existing and consistent with the helpers, but this PR rewrote these exact blocks. A success-flag try/finally would close the gap.
- BinaryTSFactory.createGenerator(File) passes the stream, not the file, to _createContentReference. That is the only reason the method has a different shape from its three siblings, including the nullable context and null-guarded release. Using the File is behaviourally free and lets the special-case test scaffolding go. See BinaryTSFactory.java:216.
- Duplicated open-decorate-wrap ladder across the File and Path overloads of both factories, and the outputToClose = w assignment after encodingWriter() is dead since that call cannot throw. A private helper per class would collapse each public method to a few lines.
- Test scaffolding duplicates FailedParserConstructionCloseTest. The binary factory stub is now hand-rolled in three test classes, and the close trackers record a boolean rather than a count, so double-close cannot be detected.

Dongnyoung:
@pjfanning Thanks for the detailed review. These are good catches🧐

I'll retarget the PR to 3.1 and address the blocking issues first, especially preserving unconditional cleanup of the raw file stream, avoiding the binary generator close behavior change, and tightening the regression tests.

cowtowncoder (approved):

PR #1714, opened 2026-09-15
Fix #1713: UTF32Reader corrupts or drops the low surrogate on a split read

### LINKED ISSUE #1713: `UTF32Reader` corrupts or drops the low surrogate when a supplementary character splits across a read() boundary
### Describe the bug

`UTF32Reader` stores the wrong value when a supplementary character has to be split across a `read()` boundary, so the low surrogate is either corrupted or dropped.

`UTF32Reader.java` (2.21, `:164-169`):

```java
cbuf[outPtr++] = (char) (0xD800 + (ch >> 10));
lo = (0xDC00 | (ch & 0x03FF));      // the low surrogate
if (outPtr >= outEnd) {             // no room for it in this call
    _surrogate = (char) ch;         // stores the normalized code point instead
    break main_loop;
}
```

`ch` is `codepoint - 0x10000`, not the low surrogate. The non-split path two lines below writes `(char) lo`, so the same method emits two different things for the same character depending on whether the buffer had room.

### Version information

2.21 and later, including 3.x. The line is identical on `2.21`, `2.22`, `2.x`, `3.1`, `3.2` and `3.x`.

### To Reproduce

```java
// U+1F600, big-endian UTF-32
byte[] input = { 0x00, 0x01, (byte) 0xF6, 0x00 };
UTF32Reader r = new UTF32Reader(null, null, input, 0, input.length, true);
r.read();   // 0xD83D, correct
r.read();   // 0xF600 -- expected 0xDE00
```

`read()` with no arguments delegates to `read(_tmpBuf, 0, 1)`, so it always splits. Two flavours of damage:

- `U+1F600` comes back as `0xF600`, a private-use character, instead of `0xDE00`.
- `U+10000` normalizes to `0`, and `(char) 0` is the `NC` "no surrogate pending" sentinel, so the low half is **dropped entirely** and the reader reports EOF one character early.

### Expected behavior

The carried-over half should be the low surrogate that was just computed, `(char) lo`.

### Additional context

It also reaches the parser. `JsonFactory.createParser(byte[])` builds a `UTF32Reader` over the whole array with no `InputStream`, so a single `read(cbuf, 0, 4000)` can consume arbitrarily many bytes and a supplementary character can land exactly on the 4000-char boundary. Parsing `["<3997 chars>\uD83D\uDE00"]` encoded as UTF-32 then returns a string with the emoji corrupted.

With an `InputStream` source it is not reachable, because the 8000-byte read buffer runs out before the 4000-char output buffer fills.

I have a one-line fix and tests ready and will open a PR against `2.21` shortly.

youdie006:
Fixes #1713.

`UTF32Reader` stores the wrong value when a supplementary character has to be split across a `read()` boundary:

```java
lo = (0xDC00 | (ch & 0x03FF));      // the low surrogate
if (outPtr >= outEnd) {             // no room for it in this call
    _surrogate = (char) ch;         // stores the normalized code point instead
```

`ch` is `codepoint - 0x10000`. The non-split path two lines below writes `(char) lo`, so the same method emits two different things for the same character depending on whether the buffer had room.

`U+1F600` comes back as `0xF600` instead of `0xDE00`. `U+10000` normalizes to `0`, which is the `NC` "no surrogate pending" sentinel, so its low half is **dropped** and the reader reports EOF one character early.

Targeting `2.21` per the oldest-branch rule in `CLAUDE.md`; the line is identical on `2.21`, `2.22`, `2.x`, `3.1`, `3.2` and `3.x`.

<details>
<summary>Verification</summary>

Three tests in a new `UTF32ReaderSurrogateTest`: the two `read()` cases above, and a parser-level one. `createParser(byte[])` builds a `UTF32Reader` over the whole array with no `InputStream`, so one `read(cbuf, 0, 4000)` can consume arbitrarily many bytes and a supplementary character can land on the 4000-char boundary. With an `InputStream` source it is not reachable, because the 8000-byte read buffer runs out first.

| row | `UTF32Reader.java` md5 | result |
|---|---|---|
| master | `ff5f734e7a7d7e07040d90a67a49dee5` | **3 of 3 fail**: `56832` vs `62976`; `56320` vs **`-1`**; parser returns the string with the emoji corrupted |
| this change | `903320f60c1e10bba3b6bda45da51c16` | 3 of 3 pass |

Full suite on the change: `Tests run: 1557, Failures: 0, Errors: 0, Skipped: 2`, `BUILD SUCCESS`.

Mutation rows run on `3.x`, where the same three tests exist, to check the tests are not redundant:

| mutant | killed by |
|---|---|
| `(char) (ch & 0x03FF)` | all three |
| `(char) hi` | all three |
| `(char) (0xDC00 \| ch)` without the mask | the `U+1F600` test and the parser test only |
| `(char) (ch + 0xE800)` | the `U+10000` test only |

The last two kill disjoint sets, so neither of those tests is redundant. On `3.x` the same fix gives `mvn -B -ff -ntp verify` `Tests run: 1929, Failures: 0`, including `ModuleInfoIT` and `FastDoubleParserShadingIT`.

Family sweep: `git grep -n "0xDC00" -- 'src/main/**/*.java'` finds 15 sites that build a low surrogate as `0xDC00 | (x & 0x3FF)`. Every other one either assigns back into the variable that is then written or writes it straight to the output buffer. `UTF32Reader` is the only one that computes it into a separate variable and then stores something else.

No existing test pinned the old behaviour.

</details>

Disclosure: written with AI assistance (Claude). Every number above is measured.


pjfanning:
@youdie006 thanks - have you sent in a CLA?
https://github.com/FasterXML/jackson/blob/main/CLA-jackson-pre-2026.md

It is required for PRs to this project.

youdie006:
Thanks for checking. I’m not comfortable signing the CLA since it asks for my real name and a scanned physical signature, so I’m going to close this. The bug is still documented in #1713 if anyone wants to pick it up.

Issue #1718, opened 2026-09-17
Source opened for `File`/`Path` parser leaks if decorated stream `close()` fails on failed construction


cowtowncoder:
(filed by Claude Code)

## Description

`createParser(File)` / `createParser(Path)` in `TextualTSFactory` and `BinaryTSFactory` reassign the
raw `InputStream` reference when decorating it:

```java
IOContext ioCtxt = _createContext(_createContentReference(f), true);
InputStream in = null;
try {
    in = _fileInputStream(f);
    in = _decorate(ioCtxt, in);   // reference to stream Jackson opened is lost here
    return _createParser(readCtxt, ioCtxt, in);
} catch (RuntimeException e) {
    _closeOnFailedConstruction(in, e);
    _releaseOnFailedConstruction(ioCtxt, e);
    throw e;
}
```

After a successful `_decorate()` the only retained reference is the decorated stream. If
`_createParser()` then fails, cleanup closes the decorated stream -- which is correct -- but if that
`close()` throws, the stream Jackson opened (`_fileInputStream()` / `_pathInputStream()`) is never
closed and leaks:

```text
open InputStream -> decorate (wrapper returned) -> parser construction fails
   -> wrapper.close() throws -> stream Jackson opened never closed
```

The secondary failure is only recorded as suppressed, so the leak is silent.

This is the read-side counterpart of the write-side ownership problem fixed in #1711, and completes
the cleanup work done in #1692 / #1693 / #1694.

Note that if `_decorate()` itself fails, `in` still refers to the raw stream and it is closed
correctly; only the post-decoration window is affected.

## Affected paths

- `TextualTSFactory.createParser(ObjectReadContext, File)`
- `TextualTSFactory.createParser(ObjectReadContext, Path)`
- `BinaryTSFactory.createParser(ObjectReadContext, File)`
- `BinaryTSFactory.createParser(ObjectReadContext, Path)`

Caller-provided sources (`InputStream`, `Reader`, `byte[]`, `String`) are not affected: Jackson does
not own those.

## Suggested fix

#1711 added a helper in `DecorableTSFactory` for exactly this ownership question:

```java
static void _closeOnFailedConstruction(Closeable toClose, Closeable rawFallback,
        RuntimeException failure)
```

It closes the outermost resource and falls back to the raw one only if that close threw (so the
stream Jackson opened is not closed twice in the normal case). Tracking the outermost resource, as
the generator paths now do, fixes all four methods:

```java
[...]

PR #1719, opened 2026-09-17
Fix #1718: Close File/Path parser sources if decorated stream close fails

kalayciburak:
Fixes #1718

## Description

Read-side counterpart of #1711: `createParser(File)` / `createParser(Path)` in
`TextualTSFactory` and `BinaryTSFactory` reassigned the raw `InputStream` after
decoration. If parser construction then failed and closing the decorated stream
threw, the stream Jackson opened leaked (the close failure was only recorded as
suppressed).

Construction now tracks the outermost successfully created `Closeable` and uses
`_closeOnFailedConstruction(Closeable, Closeable, RuntimeException)` from #1711:
the decorated stream is closed first, and the raw stream is closed only if that
close threw.

Caller-provided sources (`InputStream`, `Reader`, `byte[]`, `String`) are unchanged:
Jackson does not own those.

## Tests

- `FailedParserConstructionCloseTest` 16/0 (4 new cases: JSON and binary `File`/`Path`
  when decorated `close()` throws)
- `FailedGeneratorConstructionCloseTest` 16/0
- `./mvnw -B -ff -ntp test` 1762/0 (2 skipped)


pjfanning:
Small issue:
- FailingCloseInputStream.closeAttempted and FailingCloseInputDecorator.wrappers are populated but never asserted — either assert wrappers.get(0).closeAttempted (proves close order: decorated first, raw as fallback) or drop them.

kalayciburak:
asserted wrappers.get(0).closeAttempted, pushed

cowtowncoder (approved):

cowtowncoder:
Thank you, @kalayciburak ! Looks good, will merge.

PR #1721, opened 2026-09-19
Fix null append handling in SegmentedStringWriter

jakezwang:
`SegmentedStringWriter.append(null)` and `append(null, start, end)` currently throw `NullPointerException`. The `Writer` contract requires them to append `"null"` or the requested subsequence of it.

Handle null in both overloads and add regressions for the full string, subsequences, empty ranges, invalid bounds, and the fluent return value. Non-null `CharSequence` handling is unchanged.

Validation: `./mvnw -B -ff -ntp verify` on JDK 21 and `./mvnw -B -ff -ntp clean verify animal-sniffer:check` on JDK 8 pass (1,561 tests, 2 skipped; Android API 26 compatibility passes).


src/main/java/com/fasterxml/jackson/core/io/SegmentedStringWriter.java
```diff
@@ -49,14 +49,14 @@ public Writer append(char c) throws IOException {
...
-        String str = csq.toString();
+        String str = String.valueOf(csq);
         _buffer.append(str, 0, str.length());
         return this;
     }
 
     @Override
     public Writer append(CharSequence csq, int start, int end) throws IOException {
-        String str = csq.subSequence(start, end).toString();
+        String str = (csq == null ? "null" : csq).subSequence(start, end).toString();
```
cowtowncoder, line 59: No, `null` should be invalid and throw NPE.

src/main/java/com/fasterxml/jackson/core/io/SegmentedStringWriter.java
```diff
@@ -49,14 +49,14 @@ public Writer append(char c) throws IOException {
 
     @Override
     public Writer append(CharSequence csq) throws IOException {
-        String str = csq.toString();
+        String str = String.valueOf(csq);
```
cowtowncoder, line 52: As in other case, no `null` allowed.

cowtowncoder:
`SegmentedStringWriter` is Jackson internal helper class, not part of API.  Its null-handling is an internal concern. I don't see much point in trying to change handling here.

Is there are some actual problem being solved by this PR?

